PR #103, opened 2018-11-14
first version of spec template

SebastienGllmt:
See README in the PR


docs/specs/TEMPLATE.md
```diff
@@ -0,0 +1,15 @@
...
+
+Why you think the proposal is necessary
+
+# Background
+
+Background information related to the proposal (current implementation, rationale for the implementation you will propose)
+
+# Proposal
+
+What you propose should be done
```
vantuz-subhuman: Should we add something like this at the end 🤔?

```
#Possible negative drawbacks or complications

Optionally describe what are the complications that could go wrong while or after implementing this feature.
```

SebastienGllmt: I think that would go in the `background` section where you could contrast different implementation ideas

PR #114, opened 2018-11-20
Cleanup Store folder & associated code

SebastienGllmt:
-Remove a bunch of dead code (inherited from Daedalus)
-Added types (a large amount of stuff resolved to `any`)
-Better abstraction for picking which API to use (ex: `ada`)
-cleanup/documentation


nicarq:
Awesome!!! 🙏

PR #117, opened 2018-11-21
Sort mnemonic phrase to make them easier to click

SebastienGllmt:
Crazy this wasn't done a long time ago. Can anybody think of a reason why this wasn't the case from the start?


SebastienGllmt:
image
Here is an image showing the change

PR #122, opened 2018-11-24
Cleanup/ui

SebastienGllmt:
More work needs to be done documenting these two folders but a few different improvements are already included in this PR so I'll get this one merged first.

This PR
- Gets rid of injection of stores/actions into containers (we need this to get rid of the weird Flow workaround where `mobx-react` injection isn't compatible with Flow. See [this article](https://wietse.loves.engineering/using-flowtype-with-decorators-in-react-af4fe69e66d6))
- Fixes eslint [require-default-props](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-default-props.md)
- Fixes eslint [button-has-type](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/button-has-type.md)
- Fixes eslint [no-array-index-key](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md)
- Fixed `TopBar.js` (component) depending on the `wallet` store (components should not depend on store)
- Fixed `WalletSettings.js` and `WalletSendForm.js` (both components) depending on containers (containers should not depend on components)

With this there should be no `eslint` errors left :)

### DIFF
--- a/app/components/daedalusTransfer/DaedalusTransferErrorPage.js
+++ b/app/components/daedalusTransfer/DaedalusTransferErrorPage.js
@@ -28,6 +28,9 @@ type Props = {
 
 @observer
 export default class DaedalusTransferErrorPage extends Component<Props> {
+  static defaultProps = {
+    error: undefined
+  };
 
   static contextTypes = {
     intl: intlShape.isRequired
--- a/app/components/wallet/transactions/WalletTransactionsList.js
+++ b/app/components/wallet/transactions/WalletTransactionsList.js
@@ -121,8 +121,8 @@ export default class WalletTransactionsList extends Component<Props> {
 
     return (
       <div className={styles.component}>
-        {transactionsGroups.map((group, groupIndex) => (
-          <div className={styles.group} key={walletId + '-' + groupIndex}>
+        {transactionsGroups.map(group => (
+          <div className={styles.group} key={walletId + '-' + group.transactions.map(tx => tx.id).join('-')}>
             <div className={styles.groupDate}>{this.localizedDate(group.date)}</div>
             <div className={styles.list}>
               {group.transactions.map((transaction, transactionIndex) => (
--- a/app/components/wallet/transactions/WalletTransactionsList.js
+++ b/app/components/wallet/transactions/WalletTransactionsList.js
@@ -11,6 +11,7 @@ import Transaction from './Transaction';
 import WalletTransaction from '../../../domain/WalletTransaction';
 import LoadingSpinner from '../../widgets/LoadingSpinner';
 import type { AssuranceMode } from '../../../types/transactionAssuranceTypes';
+import { Logger } from '../../../utils/logging';
 
 const messages = defineMessages({
   today: {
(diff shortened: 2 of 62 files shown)


SebastienGllmt:
FYI I would say maybe `no-array-index-key` is the only one that requires some amount of thought in of the `eslint` issues

app/containers/wallet/WalletSendPage.js
```diff
@@ -22,14 +20,27 @@ export default class WalletSendPage extends Component<Props> {
...
     const { isValidAddress } = wallets;
     const { calculateTransactionFee, validateAmount, hasAnyPending } = transactions;
     const activeWallet = wallets.active;
 
     // Guard against potential null values
     if (!activeWallet) throw new Error('Active wallet required for WalletSendPage.');
 
+    // Callback that creates a container to avoid the component knowing about actions/stores
+    const dialogCallback = function (props: DialogProps) {
+      return (<WalletSendConfirmationDialogContainer
```
SebastienGllmt: I'm not sure if this kind of callback to create a container is really the right idiomatic approach. I tried to search around to see if there's a well-defined way to deal with these cases but couldn't really find anything. It feel like there's a good chance you're not supposed to structure your code so that this ever becomes an issue. If anybody knows more about this, I'd love to hear

app/components/daedalusTransfer/DaedalusTransferErrorPage.js
```diff
@@ -28,6 +28,9 @@ type Props = {
 
 @observer
 export default class DaedalusTransferErrorPage extends Component<Props> {
+  static defaultProps = {
```
nicarq: where is this used? I didn’t see it

app/components/wallet/transactions/WalletTransactionsList.js
```diff
@@ -121,8 +121,8 @@ export default class WalletTransactionsList extends Component<Props> {
 
     return (
       <div className={styles.component}>
-        {transactionsGroups.map((group, groupIndex) => (
-          <div className={styles.group} key={walletId + '-' + groupIndex}>
+        {transactionsGroups.map(group => (
+          <div className={styles.group} key={walletId + '-' + group.transactions.map(tx => tx.id).join('-')}>
```
nicarq: this adds some performance hit. Can we just use the hash of the first tx of every group? they should be unique

app/containers/wallet/WalletSendPage.js
nicarq: It seems that this is been constructed right away, but I guess if `WalletSendConfirmationDialogContainer` is not too heavy is completely fine.

app/components/daedalusTransfer/DaedalusTransferErrorPage.js
SebastienGllmt: it's used by React internally to give a default value for the arguments to the React component

nicarq: okay

app/components/wallet/transactions/WalletTransactionsList.js
SebastienGllmt: good idea!

PR #124, opened 2018-11-24
Enable React devtools

SebastienGllmt:
Idk if there is a better way to do this that I don't know about but this worked for me.


chrome/views/main_window.pug
```diff
@@ -2,6 +2,8 @@ doctype html
 
 html
   head
+    if src=env == 'development'
```
nicarq: pug? more like wut

PR #133, opened 2018-11-30
Add mobx dev tools

SebastienGllmt:
You can see the kind of features this enables here
https://github.com/mobxjs/mobx-react-devtools#mobx-react-devtools


SebastienGllmt:
Fun fact: turns out Daedalus uses this also but they added it slightly differently. You can see it at https://github.com/input-output-hk/daedalus/blob/develop/source/renderer/app/App.js#L28

PR #136, opened 2018-12-01
Feature adding Theme selection

SebastienGllmt:
The option to select "themes" (different color settings) is present in Daedalus but was partially removed in Yoroi. I added this feature back by copy-pasting the Daedalus code back in place.

This is still a WIP for the following reasons:
- [ ] I added a 2nd theme to Yoroi to test the feature but it is just the original theme copy/pasted
- [x] We have no icons to preview the themes in the menu yet

Additionally, I changed the code from Daedalus to use `svg` for the theme previews instead of `png`. I will explain the reason for this after posting this

**NOTE**: This feature depends on `cleanup/uipart2` to be merged.


SebastienGllmt:
See https://github.com/Emurgo/yoroi-frontend/pull/139 for a spec on how I propose we take this feature from the base feature I added in this PR.

SebastienGllmt:
Replaced by #284

PR #149, opened 2018-12-04
Reusable Transfer Screen

SebastienGllmt:
To enable transferring V2 wallets as discussed in #125
We will have two screens: One for v2 wallet transfer and one for Daedalus transfer

To make these screen easier to make, I extracted parts of the Daedalus transfer screen into customizable components that way we can reuse the UI code.

I also got rid of the dialog option on the transfer screen
image
Here is the rationale I gave for doing so
```
Is anybody opposed to removing the "no" option entirely and simplifying the dialog to JUST the button to "transfer all funds from Daedalus wallet"?
The `no` option is objectively a bad decision since it will be more tedious for them to create the transaction and it will generate a change address. Additionally, it seems like pretty obvious information
I want to make this page more generic to be reusable for paper wallets, yoroi import, etc. but it the option makes even less sense in that case (ex: "Do you have a hot wallet that already loaded your paper wallet? If so, go to the receive screen!" makes no sense)
```
**Note:** To avoid the PR getting too big, this PR only extracts out the components and doesn't actually implement the new v2 wallet transfer screen.

### DIFF
--- a/app/components/transfer/TransferInstructionsPage.js
+++ b/app/components/transfer/TransferInstructionsPage.js
@@ -65,8 +65,8 @@ export default class TransferInstructionsPage extends Component<Props> {
       styles.button,
     ]);
 
-    const answerNoButtonClasses = classnames([
-      'answerNoButton',
+    const confirmButtonClasses = classnames([
+      'confirmButton',
       'primary',
       styles.button,
     ]);
@@ -119,7 +119,7 @@ export default class TransferInstructionsPage extends Component<Props> {
               </div>
 
               <Button
-                className={answerNoButtonClasses}
+                className={confirmButtonClasses}
                 label={confirmationText}
                 onClick={onConfirm}
                 disabled={disableTransferFunds}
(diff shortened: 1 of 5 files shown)


nicarq:
I will review it tomorrow

app/api/ada/daedalusTransfer.js
```diff
@@ -75,7 +75,7 @@ export async function generateTransferTx(payload: {
     const recoveredBalance = await getBalance(senders);
     const inputs = _getInputs(senderUtxos, addressesWithFunds);
 
-    // pick which address to send migration to
+    // pick which address to send transfer to
```
vantuz-subhuman: ❤️

app/components/transfer/AnnotatedLoader.js
```diff
@@ -0,0 +1,39 @@
...
+    return (
+      <div className={styles.component}>
+
+        <div>
+          <div className={styles.body}>
+
+            <div className={styles.title}>
+              {title}
+            </div>
+
```
nicarq: this formatting is a little strange. is it manual?

nicarq: (the line breaks)

nicarq:
LGTM

PR #151, opened 2018-12-04
Fix various manage-translations problems

SebastienGllmt:
- can no longer accidentally call `manage-translations` before building rebuilding your translation cache (this used to delete all your i18n files)
- purge-translations automatically when checking out new branch (not doing this manually before caused `manage-translations` to give incorrect results)
- suppress generation of `defaultMessages.json` (used to be automatically generated after m`manage-translations` but was unnecessary)

### DIFF
--- a/.githooks/post-checkout
+++ b/.githooks/post-checkout
@@ -1,0 +1,8 @@
+#!/bin/bash                                                                      
+
+set -e
+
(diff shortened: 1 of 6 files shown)


nicarq:
sorry the delay @SebastienGllmt I’m reviewing it right meow :D

PR #160, opened 2018-12-09
Decode Rust tx into Trezor

vantuz-subhuman:
## Rationale

Trezor implements only some primitive input-selection, and does not implement fee and change calculation at all. To mitigate this problem there was implemented a hack: we first called Rust to estimate a fee for a transaction, and then called Trezor to construct a transaction with this fee/change values we got from Rust.

The problem is that after calling Rust we **only** was reading the estimated fee from the response, and then asking Trezor to use the same fee, but supplying it the whole set of available inputs, so it would select them by itself. And coin-selection strategies differ between Rust and Trezor, so Trezor sometimes selects completely different set of inputs, but tries to use the same fee/change value, and it does not match.

## Quick abstract

When calling Rust to estimate a fee - it performs full-blown coin-selection and produces fake witnesses, to calculate the accurate fee value. Instead of just reading the fee, and throwing away all the other data - we will parse the resulting transaction and use the same sets of inputs and outputs directly in Trezor.

This way the logic is most natural and straightforward:
1. Call Rust to create am unsigned transaction
2. Translate this transaction into the Trezor format
3. Call Trezor to sign it

## New dependencies

Had to add new dependency:
```
"cbor": "4.1.3"
```

This library is used to decode the Rust txs.

## Full list of changes

1. Introduced new type `adaTypes.js/AdaFeeEstimateResponse` and used as return from `adaNewTransactions.js/getAdaTransactionFee` instead of `AdaTransactionFee`. The new type includes the fee, but also contains a decoded **unsigned** Rust transactions with full inputs (instead of pointers) and outputs.

2. Implemented function `utils.js/decodeRustTx` that takes a tx-construction response from Rust and returns fully parsed `CardanoCrypto.js/CryptoTransaction` (with pointers as inputs)

3. In `CardanoCrypto.js` - extracted type `UnsignedTransaction` from `CryptoTransaction` type, and added new type `UnsignedTransactionExt`, which is completely like `UnsignedTransaction`, but with full `TxInput` instead of `TxInputPtr`. (Also exported a lot of types, to be able to import them in other modules.)

4. Implemented `adaNewTransactions.js/decodeRustTxWithInputs` that takes a Rust response and a list of used inputs and produces, calls to `utils.js/decodeRustTx` to parse the Rust response, and then uses the parsed transactions and the provided list of inputs to produce an `UnsignedTransactionExt` (with full inputs, instead of pointers).

5. Reworked `trezorNewTransactions.js/createTrezorSignTxData` so it takes only a single argument of type `UnsignedTransactionExt` and uses all the data from it to construct a Trezor tx. This way we get in Trezor fully analogous transaction as in Rust.

6. Reworked functions `_transformToTrezorInputs` and `_generateTrezorOutputs` in `trezorNewTransactions.js`, so they use only sets of `TxInput` or `TxOutput`, received from the parsed Rust transaction. Also replaced `txsBodiesForUTXOs` with `txsBodiesForInputs`, because information from provided inputs is enough to query tx bodies.

7.  Reworked `ada/index.js/createTrezorSignTxData` so it processes new response type from `getAdaTransactionFee` and also passes only the parsed tx into `createTrezorSignTxData` and processes its new return type.

8. Fixed the return type in `ada/lib/yoroi-backend-api.js/getTxsBodiesForUTXOs`


app/api/ada/lib/utils.js
```diff
@@ -53,3 +61,40 @@ const _getTxCondition = (state: string): AdaTransactionCondition => {
   if (state === 'Pending') return 'CPtxApplying';
   return 'CPtxWontApply';
 };
+
+export function decodeRustTx(resp: SpendResponse): CryptoTransaction {
```
nicarq: if we only use `cbor_encoded_tx` I would make that the input of the function

flow/declarations/CardanoCrypto.js
```diff
@@ -162,16 +162,25 @@ declare type CryptoAccount = {
...
     witnesses: Array<TxWitness>
   }
 }
 
+export type UnsignedTransaction = {
+  inputs: Array<TxInputPtr>,
+  outputs: Array<TxOutput>
+}
+
+export type UnsignedTransactionExt = {
```
nicarq: Ext stands for extended?

app/api/ada/hardware-wallet/trezorNewTransactions.js
```diff
@@ -31,55 +31,50 @@ import {
 import type { CreateTrezorSignTxDataResponse } from '../index.js';
 import type {
   TrezorInput,
-  TrezorOutput
+  TrezorOutput, TrezorSignTxPayload
```
nicarq: could u move this one to the next line plz :D

nicarq:
aweome!! thanks Rus :D :D :D

app/api/ada/lib/utils.js
vantuz-subhuman: I wanted to do it too, but then input type would be `Array<number>` which looks kinda weird 😟 I will check if I can maybe easily add a new type alias, like `RawRustTxBody`, or something like that 🤔

flow/declarations/CardanoCrypto.js
vantuz-subhuman: Yes, and I will gladly take any ыsuggestions for this one :) 

Generally we have two types of representing a Tx:
1. `UnsignedTransaction` - just like a tx sent over the network looks, with input-pointers
2. "Extended" `UnsignedTransaction` - for internal use, just like the normal one, but with full inputs

And later we might have to have the same distinction for signed txs:
1. `CryptoTransaction` - full tx with witnesses that includes `UnsignedTransaction` (with input-pointers)
2. "Extended" `CryptoTransaction` - all the same, but includes "Extended" `UnsignedTransaction` with full inputs.

app/api/ada/lib/utils.js
vantuz-subhuman: Added new type alias `RustRawTxBody`: https://github.com/Emurgo/yoroi-frontend/pull/160/commits/0767e6d8b5c73452c9715d27b4ceebc5ae8c076b

vantuz-subhuman:
I have got successful Trezor tx construction logged (from this line: https://github.com/Emurgo/yoroi-frontend/pull/160/files#diff-d497615176f2e8955fcb2f9dc2cf40e6R316):

image

But then actual sending of this transaction to Trezor for signing fails, at least on my local mainnet build with Trezor emulator).

vantuz-subhuman:
Yeah, there was a lil error in Trezor tx formatting, now it's working fine:

image

Successfully sent a translated from Rust tx on mainnet.

vantuz-subhuman:
Finished all fixes and tested on local mainnet build, ready for review and merge @nicarq, @sarveshwar-singh

flow/declarations/CardanoCrypto.js
nicarq: The extended naming makes sense! the part that I’m not completely sold is using an abbreviation for it. Even if it’s long I would use the full naming :-/

app/api/ada/lib/utils.js
nicarq: oh good idea

flow/declarations/CardanoCrypto.js
vantuz-subhuman: Makes sense, but it's already merged now ☹️ Will try to refactor it in some later PR )

Issue #164, opened 2018-12-11
TrezorConnect.cardanoSignTransaction() change outputs payload to use derived path


sarveshwar-singh:
use `derived path` (ike 44/ 1815/ etc) instead of address in outputs payload of  `TrezorConnect.cardanoSignTransaction()`

PR #166, opened 2018-12-11
TrezorConnect.cardanoSignTransaction() change outputs payload to use derived path

vantuz-subhuman:
Closing #164

1. Introduced new optional field `TxOutput.fullAddress? : AdaAddress`
2. Added full-address resolving in `adaNewTransactions.js/decodeRustTxWithInputs`
3. Added case-check for full address in `trezorNewTransactions.js/_generateTrezorOutputs`

### DIFF
--- a/app/api/ada/adaTransactions/adaNewTransactions.js
+++ b/app/api/ada/adaTransactions/adaNewTransactions.js
@@ -203,6 +203,7 @@ function decodeRustTxWithInputs(
     outputs: tx.tx.tx.outputs.map(x => {
       if (x.address === changeAddress.cadId) {
         x.fullAddress = changeAddress;
+        x.isChange = true;
       }
       return x;
     })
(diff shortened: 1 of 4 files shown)


vantuz-subhuman:
image

app/api/ada/adaTransactions/adaNewTransactions.js
```diff
@@ -199,7 +200,12 @@ function decodeRustTxWithInputs(
   const selectedInputs = availableInputs.filter((inp: TxInput) => set.has(pointerToStr(inp.ptr)));
   return {
     inputs: selectedInputs,
-    outputs: tx.tx.tx.outputs
+    outputs: tx.tx.tx.outputs.map(x => {
+      if (x.address === changeAddress.cadId) {
+        x.fullAddress = changeAddress;
```
nicarq: maybe we could use `isChangeAddress` instead of `fullAddress`? the code looks good! but I was thinking that if someone sees `fullAdress` is difficult to infer that this is used as to determinate a changeAddress.

vantuz-subhuman: Makes sense, will change in a bit

PR #168, opened 2018-12-12
Spec/refactor

JizKadasha:
A refactor proposal which describes the basic ideas.

### DIFF
--- a/docs/specs/code/REFACTOR_PROPOSAL.md
+++ b/docs/specs/code/REFACTOR_PROPOSAL.md
@@ -24,3 +24,7 @@ We don't have a possibility to access store or actions from any component withou
 - fix the structure
 
 For now, the MainLayout component is used in most containers - they return the content wrapped in the MainLayout, and it happens in most cases. We could just wrap the routing in the MainLayout so we will achieve the same result but without breaking DRY principle. Again, for now it's a lot of really unnecessary code, with all it's cons. There can be some more issues with the structure, but this one is probably the main one.
+
+- additional: switch from Mobx to Redux
+
+Need to check if Redux has something to work with Intl, and if it does, we can use it - probably with Redux-sagas, which have a very neat syntax, very easy to test and which simplify a lot working with async actions and other side effects.
--- a/docs/specs/code/REFACTOR_PROPOSAL.md
+++ b/docs/specs/code/REFACTOR_PROPOSAL.md
@@ -1,0 +1,26 @@
+# Idea
+
+Refactor the app, use newer dependencies, improve code quality
+
+# Motivation
+
+Currently there are some things in the project which make working with it a bit too hard. The project structure could be better, the routing could be better, we could use the latest react-router and react and that could help a lot.
+
+
+# Proposal
+
+- update dependencies: react-router 4 instead of 3, React 16 instead of 15
+
+Basically the routing in react-router 4 is more handy and allows to work with it with less effort. As for React 16, there are some useful things - for example, <Fragment /> component, it can be used to avoid wrapping multiple JSX elements in the unnecessary <div />. Also there are new lifecycle methods in React 16, and the React team works on the optimisation, so this can be useful.
+
+- update routing
+
+Currently the routing breaks DRY principle. There are just a lot of repeated things in each route. This appearance makes code hard to read and understand. Also if we need to add a new route, we have to repeat all these previously repeated things for a new route too.
+
+- use Mobx Provider in a proper way
+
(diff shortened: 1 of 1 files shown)


docs/specs/code/REFACTOR_PROPOSAL.md
```diff
@@ -0,0 +1,30 @@
...
+
+- use Mobx Provider in a proper way
+
+We don't have a possibility to access store or actions from any component without it's own route. That's because there are no Provider wrapping the routes, only the IntlProvider, so we can see the current locale etc, but nothing more. Unfortunately, we can have some cases when we want to see store from the deeply nested component. As for now, this primitive thing requires a lot of repeated code, and more unnecessary code can accidentally bring more bugs. Also it's harder to maintain the project, to fix bugs or implement more features, and it would be so much heavier to new team members (if some) to understand the code.
+
+- fix the structure
+
+For now, the MainLayout component is used in most containers - they return the content wrapped in the MainLayout, and it happens in most cases. We could just wrap the routing in the MainLayout so we will achieve the same result but without breaking DRY principle. Again, for now it's a lot of really unnecessary code, with all it's cons. There can be some more issues with the structure, but this one is probably the main one.
```
SebastienGllmt: The top-level layouts are a little messy since Daedalus has a more complex Layout system that we didn't need.

docs/specs/code/REFACTOR_PROPOSAL.md
```diff
@@ -0,0 +1,30 @@
...
+
+- update dependencies: react-router 4 instead of 3, React 16 instead of 15
+
+Basically the routing in react-router 4 is more handy and allows to work with it with less effort. As for React 16, there are some useful things - for example, `<Fragment />` component, it can be used to avoid wrapping multiple JSX elements in the unnecessary `<div />`. Also there are new lifecycle methods in React 16, and the React team works on the optimisation, so this can be useful.
+
+- update routing
+
+Currently the routing breaks DRY principle. There are just a lot of repeated things in each route. This appearance makes code hard to read and understand. Also if we need to add a new route, we have to repeat all these previously repeated things for a new route too.
+
+- use Mobx Provider in a proper way
```
SebastienGllmt: The reason we got rid of the Provider is actually because it doesn't work with `Flow` and so using a Provider causes a bunch of types to fail to resolve which has lead to sneaky bugs in the past

docs/specs/code/REFACTOR_PROPOSAL.md
```diff
@@ -0,0 +1,30 @@
...
+
+- fix the structure
+
+For now, the MainLayout component is used in most containers - they return the content wrapped in the MainLayout, and it happens in most cases. We could just wrap the routing in the MainLayout so we will achieve the same result but without breaking DRY principle. Again, for now it's a lot of really unnecessary code, with all it's cons. There can be some more issues with the structure, but this one is probably the main one.
+
+- additional: switch from Mobx to Redux
+
+Need to check if Redux has something to work with Intl, and if it does, we can use it - probably with Redux-sagas, which have a very neat syntax, very easy to test and which simplify a lot working with async actions and other side effects.
```
SebastienGllmt: This one would be a lot more work than the others so probably requires more thought and analysis if we really want to go for it. If we really want to look into this I think we should get least @sarveshwar-singh or somebody else to look into it together and then make the case during one of our Tuesday meetings.

docs/specs/code/REFACTOR_PROPOSAL.md
```diff
@@ -0,0 +1,30 @@
...
+
+# Proposal
+
+- update dependencies: react-router 4 instead of 3, React 16 instead of 15
+
+Basically the routing in react-router 4 is more handy and allows to work with it with less effort. As for React 16, there are some useful things - for example, `<Fragment />` component, it can be used to avoid wrapping multiple JSX elements in the unnecessary `<div />`. Also there are new lifecycle methods in React 16, and the React team works on the optimisation, so this can be useful.
+
+- update routing
+
+Currently the routing breaks DRY principle. There are just a lot of repeated things in each route. This appearance makes code hard to read and understand. Also if we need to add a new route, we have to repeat all these previously repeated things for a new route too.
```
SebastienGllmt: The reason that they are all the same is because every route takes all `stores` and all `actions`. This is something we inherited from Daedalus that I don't think is good design. We should instead give some though into which `stores`/`actions` components need and only pass those in.

docs/specs/code/REFACTOR_PROPOSAL.md
JizKadasha: I'll check out what we can do with Flow and Provider. By getting rid of the Provider we also got rid of access to the stores for the deeply nested components.

docs/specs/code/REFACTOR_PROPOSAL.md
JizKadasha: Inject would help us; while we don't use Provider, we could at least iterate over an array or object with routes to get rid of writing manually every new route and passing to it all these things again and again.

docs/specs/code/REFACTOR_PROPOSAL.md
JizKadasha: Actually there will be much more work with fixing structure than with switching to Redux

PR #175, opened 2018-12-17
trezor topbar icon

sarveshwar-singh:
Functionality:
Display Yoroi with Trezor Icon in case of active wallet is Trezor Integrated Wallet

Implementation:
Till now we have 3 dynamic Categories ( one more which is responsible for displaying wallet name and current balance ).
Created new icon which is combination of Plus Logo and Trezor Logo and appended as new category when active wallet is Trezor Integrated wallet.


app/stores/toplevel/LoadingStore.js
```diff
@@ -52,9 +53,14 @@ export default class LoadingStore extends Store {
...
       if (wallets.first) {
+        const firstWallet: Wallet = wallets.first;
+
+        // Dynamic Initialization of Topbar Categories
+        this.stores.topbar.initCategories();
+
         this.actions.router.goToRoute.trigger({
           route: ROUTES.WALLETS.TRANSACTIONS,
-          params: { id: wallets.first.id }
+          params: { id: firstWallet.id }
```
sarveshwar-singh: this is required to fix flow error

PR #176, opened 2018-12-17
Transfer from Daedalus :: client-side tx witness validation

vantuz-subhuman:
## Abstract
We have localised potential source of the "invalid witness" problem to the couple of places in Rust involved in "transfer from Daedalus" functionality. To find more detailed information we need an access to a wallet with this problem. All attempts to reproduce a wallet like this on our side did not find success.

The only option from here - is to use for debugging failing user wallets directly, and we can do it, without exposing the wallet secrets (to ourselves or to anyone else).

For this we add a new small piece of functionality that takes the transfer transaction produced by Rust and uses alternative crypto library (vacuumlabs) to validate the generated inputs and witnesses.

After validation is performed:
1. If no errors is detected - nothing changes and functionality continues with executing the transaction the usual way.
2. If an error is detected - a tiny anonymised report is produced (for each validated input) and attached to the sent transaction.

Using these reports on the backend we will be able to determine the specific place in code where something faulty is happening. Reports contain no information that could in any way compromise the client wallet, and wallet secrets are used strictly in the same scope as previously.

Possible report results:
1. `{reason: "Derivation path did not match"}` - means the procedure that selects related addresses from the list of all available UTxOs is doing something wrong.
2. `{reason: "Pub key did not match"}` - means the derivation scheme (for address key pair) in Rust is faulty sometimes for some reason
3. `{reason: "Root does not match"}` - means both derivation path and derived key are valid, but the produced root key is still does not match the address. This error could indicate that address itself might be wrongly constructed, which would mean a much more deep problem.
4. No errors - means alternative crypto library also detected no errors, but transaction still fails, which is more like a control case that would mean that the problem is on the backend, and the python debugging script that we previously used is might be wrong too.
5. Other - means some error happened in the procedure itself, and signals a wrong code or some issue in the cryptography in this PR.

## Detailed changes
1. Added dependency to `"cardano-crypto.js"` (vacuumlabs). This cause conflicts with the webpack `UglifyJsPlugin`, so had also to replace it with `babel-minify-webpack-plugin`.
2. Updated `cbor` version, cuz of https://github.com/hildjj/node-cbor/issues/92#issuecomment-447960444
3. Main changes in `daedalusTransfer.js`:
    1. Function `_validateAddressAndSignature` - takes a single address, and rust data to compare with (derivation/addressing, and public key from witness), and uses wallet secret to decode the address and try to derive the same data, to validate everything is ok with the addressing and the provided witness.
    2. Function `_validateAddressesAndSignatures` - takes the rust-produced transaction and list of used inputs (with addressing), and then decodes the tx, and derives some top level wallet secrets to be used in further validation. Then it iterates over all inputs/addressings and calls previous function for validation, collecting all results in a list.
    3. Changed `_getInputs` a bit to preserve addresses with inputs
4. In `utils.js` introduced new functions: `decodeAddress` and `createAddressRoot`
5. In `yoroi-backend-api.js` updated the `sendTx` function to send the validation data (if present) along with the transaction body.

Other than that only introduces a couple of types for readability and passed the produced reports around the workflow.


app/api/ada/daedalusTransfer.js
```diff
@@ -73,7 +86,8 @@ export async function generateTransferTx(payload: {
       throw new NoInputsError();
     }
     const recoveredBalance = await getBalance(senders);
-    const inputs = _getInputs(senderUtxos, addressesWithFunds);
+    const inputWrappers : Array<DaedalusInputWrapper> = _getInputs(senderUtxos, addressesWithFunds);
```
nicarq:
```suggestion
    const inputWrappers: Array<DaedalusInputWrapper> = _getInputs(senderUtxos, addressesWithFunds);
```

app/api/ada/daedalusTransfer.js
```diff
@@ -99,6 +121,95 @@ export async function generateTransferTx(payload: {
...
+    const errors = inputWrappers.map((inputWrapper, idx) => {
+      const address = inputWrapper.address;
+      const addressing = inputWrapper.input.addressing;
+      const pubKey = witnessPubKeys[idx];
+      try {
+        return _validateAddressAndSignature(secret, pass, address, addressing, pubKey);
+      } catch (e) {
+        return { address, reason: 'Failed to perform validation!', error: stringifyError(e) };
+      }
+    }).filter(x => x);
```
nicarq: do this does something? I’m not so knowledgeable in dark magic :D

webpack/mainnet.config.js
```diff
@@ -29,12 +30,7 @@ module.exports = {
     }),
     new webpack.optimize.OccurrenceOrderPlugin(),
     new webpack.IgnorePlugin(/[^/]+\/[\S]+.dev$/),
-    new webpack.optimize.UglifyJsPlugin({
-      comments: false,
-      compressor: {
-        warnings: false
-      }
-    }),
+    new MinifyPlugin(),
```
nicarq: do you know if this plugin deletes the comments?

app/api/ada/daedalusTransfer.js
vantuz-subhuman: Yeah, it's that the function that we are calling for each input might either return a result (indicating an error), or return nothing (`undefined`). So, like, if we map a list of three inputs and they are all ok - then we get a list like: `[undefined, undefined, undefined]`, so this filter ensures that we throw away those undefines :)

Other way to do this could be to returns something like an `Either` type from the validation function, so that there would be an object result for valid inputs too, and then do something like:
```
.filter(r => r.isError)
```

Either is fine, and I can change if this version is not readable, but I just thought, like, why create anything if we will just ignore it in the result anyways )

Other option is to rename `_validateAddressAndSignature` into something like `_getAddressSignatureError`, so it will be more logical that it might return **nothing**, when there's no error.

webpack/mainnet.config.js
vantuz-subhuman: It definitely had an option for this, but I'm not sure if it's enabled by default, tho. Will look into it!

app/api/ada/daedalusTransfer.js
nicarq: thanks for the explanation!

IMO I think is fine how it is for the `filter` and the `name` )

PR #178, opened 2018-12-18
show notification message on new wallet creation, restoration and  Trezor Integration  

shinsuny:
**Functionality:**

Show notification message on Trezor Connect success at top block of `Transaction` page.
Notification message will be displayed for 8 seconds.
(we have one more Notification message on `Receive` page which is displayed for 10 seconds)

**Problem Faced:** 

1. Notification View is hard-coded  in `Receive` page so I had to replicate those on `Transaction page` as well.

2. `Notification` object did not contain `message` information and because to that `Receive` page Notification is strangely implemented, hence added  `message` in `Notification` object.

3. Notification block was being overlapping over other block, looks OK on `Receive` page but not good on `Transaction page` hence added more 20px top padding on all 3 pages of Wallet.

( I think we need total refactoring of Notification message API because Notification view should be part top level layout and we should not require to replicate it every time when we show it on new pages)

**Debatable:**

Actual message I have use is `You have successfully integrated with Trezor Model T device` on Trezor Connect success because for Trezor there is not restore wallet option ( not needed ).

**I am open for better message text if anybody has better suggestion.** 

**Additional:**

A. Had to run `npm run manage-translations` hence in i18n files position is corrected.

B. show two more notification:
1. wallet created notification
2. wallet restored notification

C. https://github.com/Emurgo/yoroi-frontend/issues/188


app/containers/wallet/WalletSummaryPage.js
```diff
@@ -76,15 +80,40 @@ export default class WalletSummaryPage extends Component<Props> {
...
           isLoadingTransactions={recentTransactionsRequest.isExecutingFirstTime}
         />
+
         {walletTransactions}
+
       </VerticalFlexContainer>
     );
   }
+
+  _getThisPageActiveNotification = (): Notification | any => {
```
SebastienGllmt: why `any`?
Maybe use `?Notification`

app/stores/ada/AdaWalletsStore.js
```diff
@@ -160,6 +164,16 @@ export default class AdaWalletsStore extends WalletStore {
     this.refreshWalletsData();
   };
 
+  // =================== NOTIFICATION ==================== //
+  showTrezorTWalletIntegratedNotification(): void {
```
SebastienGllmt: Why put this in the `AdaWalletStore` instead of the `TrezorConnectStore`?

app/containers/wallet/WalletSummaryPage.js
```diff
@@ -76,15 +80,40 @@ export default class WalletSummaryPage extends Component<Props> {
...
+    let notification = {};
+
+    const { firstActiveNotification } = this.props.stores.uiNotifications;
+    if (firstActiveNotification
```
SebastienGllmt: Why do we only care about the first active notification?

app/containers/wallet/WalletSummaryPage.js
sarveshwar-singh: yeah, we can use `?Notification`
I was trying some other way, just forget to remove `any` 
will update it.

app/containers/wallet/WalletSummaryPage.js
sarveshwar-singh: yeah, maybe i should change it to the most recent notification.

SebastienGllmt: Why can you not just check if it contains the notification? What is special about index 0?

app/stores/ada/TrezorConnectStore.js
```diff
@@ -368,6 +368,9 @@ export default class TrezorConnectStore extends Store {
         // Load the Yoroi with Trezor Icon
         this.stores.topbar.initCategories();
```
sarveshwar-singh: `_saveTrezor` function has grown big, try to split it.

app/i18n/locales/en-US.json
```diff
@@ -210,6 +210,7 @@
   "wallet.summary.page.showMoreTransactionsButtonLabel": "Show more transactions",
   "wallet.summary.page.todayLabel": "Today",
   "wallet.summary.page.transactionsLabel": "Number of transactions",
+  "wallet.summary.page.trezorTWalletIntegratedNotificationMessage": "You have successfully integrated with Trezor Model T device",
```
sarveshwar-singh: change it to:
You have successfully integrated with your Trezor Model T device

app/containers/wallet/WalletSummaryPage.js
sarveshwar-singh: View component is designed to support only notification at a time

app/stores/ada/AdaWalletsStore.js
```diff
@@ -160,6 +167,34 @@ export default class AdaWalletsStore extends WalletStore {
...
+      duration: config.wallets.WALLET_CREATION_SUCCESS_NOTIFICATION_DURATION,
+    };
+    this.actions.notifications.open.trigger(notification);
+  }
+
+  showWalletRestoredNotification = (): void => {
+    const notification: Notification = {
+      id: globalMessages.walletRestoredNotificationMessage.id,
+      message: globalMessages.walletRestoredNotificationMessage,
+      duration: config.wallets.WALLET_CREATION_SUCCESS_NOTIFICATION_DURATION,
```
nicarq: `WALLET_CREATION_SUCCESS_NOTIFICATION_DURATION` the name seems a little off. we should create a new one or if we are going to reuse the same, we should have a more generic name.

app/stores/ada/AdaWalletsStore.js
```diff
@@ -127,6 +131,9 @@ export default class AdaWalletsStore extends WalletStore {
     this.restoreRequest.reset();
     await this._patchWalletRequestWithNewWallet(restoredWallet);
     this.refreshWalletsData();
+
+    // show success notification
+    this.showWalletRestoredNotification();
```
nicarq: there is an extra space for some weird reason 🤔

app/stores/ada/TrezorConnectStore.js
```diff
@@ -393,6 +374,33 @@ export default class TrezorConnectStore extends Store {
...
+    Logger.debug('TrezorConnectStore::_saveTrezor loading wallet data');
+    wallets.refreshWalletsData();
+
+    // Load the Yoroi with Trezor Icon
+    this.stores.topbar.initCategories();
+
+    // show success notification
+    wallets.showTrezorTWalletIntegratedNotification();
+
+    // TODO: [TREZOR] not sure if it actully distructing this Store ??
```
nicarq: small typo: destroying

app/types/notificationType.js
```diff
@@ -1,6 +1,9 @@
 // @flow
+import type { ReactIntlMessage } from './i18nTypes';
+
 export type Notification = {
   id: string,
+  message: ReactIntlMessage,
   duration: ?number,
-  secondsTimerInterval: ?IntervalID,
+  secondsTimerInterval?: ?IntervalID,
```
nicarq: Is there any way to avoid having the property as optional and so their value? I think it’s kind of messy

app/containers/wallet/WalletSummaryPage.js
```diff
@@ -76,15 +80,46 @@ export default class WalletSummaryPage extends Component<Props> {
...
+
       </VerticalFlexContainer>
     );
   }
+
+  _getThisPageActiveNotification = (): ?Notification => {
+    let notification = null;
+
+    const { mostRecentActiveNotification } = this.props.stores.uiNotifications;
+    if (mostRecentActiveNotification
```
nicarq: I would use `.include` and I think the if would be a little more clean

app/stores/ada/AdaWalletsStore.js
sarveshwar-singh: It depends upon how you think.

image


My way of thinking is we are on `ADD WALLET` page and we have 3 ways of Wallet creation.
End result will be a new wallet created, and we are showing notification for new wallet created.

I can't find more generic name, will make one constants for each way of wallet creation.
So we will have 3 constants with same value.

app/stores/ada/AdaWalletsStore.js
sarveshwar-singh: If you are talking about line number 134 then it's intentional.

I always try to group related things together and separate each group by one line.
I feel that way is more readable.

below is an example of how Trezor separate:
image
I feel it's quite readable.

app/stores/ada/TrezorConnectStore.js
sarveshwar-singh: Thanks !!

app/types/notificationType.js
sarveshwar-singh: Yes It's messy.
As I describe in the first top comment, we need total refactoring for Notification API.
It's whole messy.

I tried my best what I could correct in given time and situation.

app/containers/wallet/WalletSummaryPage.js
sarveshwar-singh: nice idea

app/stores/ada/AdaWalletsStore.js
nicarq: it was this 
screen shot 2018-12-20 at 3 10 34 pm

the spacing in 130-131, but I don’t see it in the actual code 🤔

nicarq: I guess it’s a github formatting thing. I haven never noticed it before

PR #189, opened 2018-12-21
Spec / Export transaction history to file

vantuz-subhuman:


vantuz-subhuman:
@sarveshwar-singh had a good idea that it might be a dedicated "export" icon on the transactions tab, instead of an extendable menu:

image

vantuz-subhuman:
> @sarveshwar-singh had a good idea that it might be a dedicated "export" icon on the transactions tab, instead of an extendable menu

Decided to switch to this option for now. Spec is edited.

nicarq:
good idea @sarveshwar-singh!

docs/specs/code/EXPORT_TRANSACTION_HISTORY_TO_FILE.md
```diff
@@ -0,0 +1,67 @@
...
+
+![image](https://user-images.githubusercontent.com/5585355/50343034-8e94cc80-0536-11e9-9058-1af18be67f8a.png)
+
+When user clicks "Export to file" - new dialog window opens up:
+
+![image](https://user-images.githubusercontent.com/5585355/50346357-31534800-0543-11e9-949a-873bcae9b854.png)
+
+Dialog contains:
+1. Two date selectors: "from day", "to day"
+2. Dropdown-selector "Select period"
```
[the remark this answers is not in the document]
vantuz-subhuman: There's this part in the spec:

> There also might be a list of default periods, e.g.: "this month", "last 30 days", "this quarter", "this year", "all". Default period selected might be "this month".

Idk, if we need "last 30 days", just added as an example )

I thought most popular periods would be "this month", "last month", "this quarter", "last quarter", but we can come up with any. Idk, if we should do months into the list of predefined periods. I mean, "this month" makes sense, because it might change from day to day, as well as "last month". That's what I mostly meant by "predefined periods", like when it's easier for Yoroi to quickly calculate those.

A specific month (e.g. "January 2018") might be a useful case, by it's a fixed value. I think it could be done like this: list "Period" contains **dynamic** values like:
1. "This month"
2. "Last month"
3. "This quarter"
4. "Last quarter"
5. "This year"
6. "Last year"
7. "All time"

But then also contains values:
8. "A month"
9. "Custom period"

When user selects any **dynamic** value - they can see date-selectors, but those are locked. And when they select "Custom period" - they get to set values manually. And when they select "A month" - they see only one selector that allows to choose a year and a month.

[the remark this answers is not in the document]
vantuz-subhuman: But we can get along with only the "Custom period" for now, cuz if would allow to select any month, just by selecting first and last day. And then add "A month" later, if needed.

docs/specs/code/EXPORT_TRANSACTION_HISTORY_TO_FILE.md
```diff
@@ -0,0 +1,67 @@
...
+3. Dropdown-selector: "File format"
+4. Button "Export"
+
+Comments:
+1. It might be an option for date-selectors to be locked/disabled by default, and only display the period selected in the dropdown, unless user selects option "Custom" - at which point selectors become enabled. But when user selects any other option - selectors become disabled again.
+2. If we only gonna support XSLT for now - the dropdown selector for format might be either locked/disabled on the only default option, or replaced with just a label.
```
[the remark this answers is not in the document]
vantuz-subhuman: Arbitrary, just because we initially planned to do only XSLT and then CSV was added kinda as a backup. The same logic is applied here, like: people that know they can import CSV anywhere - will be able to select CSV. But at least people that don't know about it will get the most common and popular format by default.

docs/specs/code/EXPORT_TRANSACTION_HISTORY_TO_FILE.md
```diff
@@ -0,0 +1,67 @@
...
+
+This library might be checked for example: https://libraries.io/npm/xlsx
+
+## Exported file format
+
+[Transaction_Export_Example.xlsx](https://github.com/Emurgo/yoroi-frontend/files/2700970/Transaction_Export_Example.xlsx)
+
+![image](https://user-images.githubusercontent.com/5585355/50316566-42ab3e80-04c8-11e9-8f38-e45343472545.png)
+
+We want to support this exact file format to be compatible with other services.
```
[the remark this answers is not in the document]
vantuz-subhuman: Makes sense. We prolly might have something like a "Help" question mark icon in the export dialog that opens a quick explanation.

docs/specs/code/EXPORT_TRANSACTION_HISTORY_TO_FILE.md
```diff
@@ -0,0 +1,67 @@
...
+6. **Fee Amount** - non-empty only for `Withdrawal` (because for deposit fee is not included)
+7. **Fee Cur.** - non-empty only for `Withdrawal`, and for now only always `ADA`
+8. **Exchange** - for now always empty (user might enter manually)
+9. **Trade Group** - for now always empty
+10. **Comment** - for now always empty
+11. **Date** - transaction time
+
+When file is produced - it should be named:
+```
+Yoroi-Transaction-History_YYYY-MM-DD_to_YYYY-MM-DD.{xlsx, csv}
```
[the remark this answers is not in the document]
vantuz-subhuman: Yeah, good point. We can add `HH-MM` to it, but from my own experience I can say the file name then gets quite unreadable )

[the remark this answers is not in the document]
vantuz-subhuman: I think most popular export case would be for something like "last month", when period is already closed. So of course this is a probable case, but severity is, imo, low.

docs/specs/code/EXPORT_TRANSACTION_HISTORY_TO_FILE.md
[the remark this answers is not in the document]
nicarq: Maybe we could have premade xlsx template with some documentation in another tab

vantuz-subhuman:
@nicarq, split spec in two iterations: first one - only the basic functionality, and date-selectors in the second on

docs/specs/code/EXPORT_TRANSACTION_HISTORY_TO_FILE.md
```diff
@@ -14,12 +14,14 @@ I suggest adding a new "Export to File" icon at the top-right of the transaction
...
 
-![image](https://user-images.githubusercontent.com/5585355/50386371-fb09fa00-06f5-11e9-970e-17c27da109c9.png)
+![image](https://user-images.githubusercontent.com/5585355/50401206-c6ea1400-079d-11e9-9980-2fb1c9cbc493.png)
+
 
 Dialog contains:
 1. Comment about the fact that the whole existing history will be exported
 2. Dropdown-selector: "File format"
 3. Button "Export"
+4. Icon "Help" (Question mark)
```
nicarq: good idea on adding a help button

nicarq: I didn’t see what it should contain. Maybe we could move it to the second iteration? so we have time to give it a better thought before EOY

docs/specs/code/EXPORT_TRANSACTION_HISTORY_TO_FILE.md
```diff
@@ -74,16 +76,24 @@ Now users should be able to chose the time period to export transaction for, lik
 
 Export dialog now contains additional elements to select a date-period:
 
-![image](https://user-images.githubusercontent.com/5585355/50346357-31534800-0543-11e9-949a-873bcae9b854.png)
+![image](https://user-images.githubusercontent.com/5585355/50401307-9eaee500-079e-11e9-8972-a8162a3a5982.png)
```
nicarq: I would avoid having two components that refers to the same constrain. It could be confusing if I change the dates and then I realized that I actually want a specific period - should it should clean up the other values? -. 

I think maybe we could avoid that and use something similar to what google has - I would change the periods to what you said though-.

screen shot 2018-12-24 at 11 13 07 am
screen shot 2018-12-24 at 11 13 13 am

docs/specs/code/EXPORT_TRANSACTION_HISTORY_TO_FILE.md
vantuz-subhuman: Was added just as a response to the idea that export format might not be obvious to users by default. But moving it to i2 sounds reasonable 👌

docs/specs/code/EXPORT_TRANSACTION_HISTORY_TO_FILE.md
vantuz-subhuman: Yeah, my idea was something like this (it's in the spec too):

image

So user can set custom dates only if explicitly selected "Custom". And if user selected some dates, but then selected "last month" again - dates in the fields are changed and selectors are disabled.

Will look for some libraries that do something similar.

nicarq: nice, so we are already on the same page :D

docs/specs/code/EXPORT_TRANSACTION_HISTORY_TO_FILE.md
```diff
@@ -0,0 +1,114 @@
...
+![image](https://user-images.githubusercontent.com/5585355/50343034-8e94cc80-0536-11e9-9058-1af18be67f8a.png)
+
+When user clicks "Export to file" - new dialog window opens up:
+
+![image](https://user-images.githubusercontent.com/5585355/50401206-c6ea1400-079d-11e9-9980-2fb1c9cbc493.png)
+
+
+Dialog contains:
+1. Comment about the fact that the whole existing history will be exported
+2. Dropdown-selector: "File format"
```
sarveshwar-singh: I can assume max at we will support following file-format:
1. xlsx
2. csv
3. pdf
4. json ( but can't see any use case for json)

If total max possible supported file format is only 3 or 4 then It can be easily fit into one line.
changing that to one line will have following benefits:
1. Format selection will be max 0 or 1 click operation.
2. Every option will be visible to the used at it will be faster to select.

vantuz-subhuman: I did not manage yet, but plan to do it today - we will add another iteration to the spec that introduces more various formats to the export. Different formats in this case will not mean different **file** types, but rather different vendor formats (for compatibility with different websites that allow transaction history tracking). E.g. "Site1-CSV", "Site1-XLSX", "Site2-CSV", "Site2-XLSX", etc.

So it's preferable to make design in such a way that new formats might be easily added without changing anything visually.

nicarq:
good points! some answers:

- google files are pretty big sometimes. Yoroi exporting should be fast. Block based is fine. let’s stay with what we have not because it’s better but it’s faster
- no need for zip because the files are pretty small and they are local

docs/specs/code/EXPORT_TRANSACTION_HISTORY_TO_FILE.md
```diff
@@ -0,0 +1,114 @@
...
+## Technical
+
+The difference now is when transactions are selected from the local storage - 
+start and end dates are inserted as filters.
+
+## File naming
+
+When file is produced - it should be named:
+```
+Yoroi-Transaction-History_YYYY-MM-DD_to_YYYY-MM-DD.{xlsx, csv}
```
sarveshwar-singh: For Iteration-1 I am thinking to use file name as:
`Yoroi-Transaction-History.{xlsx, csv}`

vantuz-subhuman: For iteration 1 it might be only 1 date - when document is generated (time is not necessary, imo), so if I were to export a file today - `Yoroi-Transaction-History_2018-01-11`

sarveshwar-singh: Makes sense,
will use: `Yoroi-Transaction-History_YYYY-MM-DD.{xlsx, csv}` for Iteration1

PR #190, opened 2018-12-24
Client-side tx-validation :: added explicit private key check

zuzunker:
Whenever we detect a problem in decoding the rust-tx there are two possible problematic places in code - root private key derivation and then everything else. So added the root key validation, just to exclude this possibility from analysis. Result of this operation is only a string like: "private key matched" or "private key not matched". That's it. No other data is ever exposed.


nicarq:
@vantuz-subhuman we don’t need this PR anymore?

PR #201, opened 2018-12-28
Feature/update polymorph

SebastienGllmt:
To upgrade to the latest version of `react-polymorph` (v8) we need to do two upgrades:

[x] upgrade to v7.2 (based on [this PR](https://github.com/input-output-hk/daedalus/pull/950))
[ ] upgrade to v8 (based on [this PR](https://github.com/input-output-hk/daedalus/pull/1139))

Unfortunately v7.2 has a bug with it that is fixed in v8, but v8 also introduces large refactoring changes we have to deal with so we have to do both at once.


SebastienGllmt:
Before this is merged we need to
- [ ] Fix the test `Scenario: User can send funds from one Yoroi wallet to another (IT-54)` that fails due to [this issue in `react-polymorph`](https://github.com/input-output-hk/react-polymorph/issues/98)

PR #210, opened 2019-01-02
Ruslan / Transaction export API

vantuz-subhuman:
1. Created separate `ExportApi` that can be used abstractly
2. Also created separate function to send any file as a download to user browser (as utils)
3. Created some API-boundary types to exchange data from any other part of the system into export-api

Export for all ADA transactions may be called as:
```
const rows = await yoroi.api.ada.getTransactionRowsToExport();
yoroi.api.export.exportTransactions({ rows, fileName: '<some_name>' });
```

image

Tests for formatting function will be added.


app/api/ada/index.js
```diff
@@ -557,9 +559,45 @@ export default class AdaApi {
...
+    }
+  }
+
+  static convertAdaTransactionsToExportRows(
+    transactions: Array<AdaTransaction>
+  ): Array<TransactionExportRow> {
+    return transactions
+      .filter(tx => tx.ctCondition === 'CPtxInBlocks')
+      .map(tx => {
+        const fullValue = parseInt(tx.ctAmount.getCCoin, 10);
```
nicarq: I think this overflows, so probably we should use `BigInt`

app/api/export/index.js
```diff
@@ -0,0 +1,135 @@
...
+export type TransactionExportFileType = 'csv';
+export type TransactionExportDataFormat = 'CoinTracking';
+
+export type ExportFileResponse = {
+  data: Blob,
+  fileType: TransactionExportFileType
+}
+
+/**
+ * This api layer provides access to the electron local storage
```
nicarq: electron -> browser? :D

app/api/export/index.js
```diff
@@ -0,0 +1,135 @@
...
+
+function _formatExportRowTypeForCoinTracking(type: string): string {
+  switch (type) {
+    case 'in': return 'Deposit';
+    case 'out': return 'Withdrawal';
+    default: throw new Error('Unexpected export row type: ' + type);
+  }
+}
+
+function _writeCsvDataIntoCsvFile(data: CsvData): string {
```
nicarq: I don’t know if it’s too picky, but instead of using `write` maybe we could use something different -- because it doesn’t directly write somewhere, but rather format --

app/api/export/utils.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Make browser to download the specified blob of bytes as a file with the specified name
+ */
+export function sendFileToUser(data: Blob, fileName: string) {
```
nicarq: good idea to put this in utils!

nicarq: maybe we should make it async void?

app/api/ada/index.js
vantuz-subhuman: Oh that's a really good catch, yeah. Replaced with `BigNumber` 👌

app/api/export/index.js
vantuz-subhuman: Damned copy-paste 😬 Fixed the whole comment )

app/api/export/index.js
vantuz-subhuman: Yeah, I agree. Renamed and changed it a bit.

app/api/export/index.js
```diff
@@ -0,0 +1,135 @@
...
+ */
+
+export default class ExportApi {
+
+  /**
+   * Request object MUST contains:
+   * - rows: array of export-data
+   * - fileName: only name without extension
+   *
+   * No result will be returned. File is sent to user as side-effect.
```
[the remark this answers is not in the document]
vantuz-subhuman: Yep, changed to async 👌

app/api/export/utils.js
vantuz-subhuman: Yep, changed to async 👌

nicarq:
LGTM :D

vantuz-subhuman:
@nicarq, @sarveshwar-singh, I have added some final changes:
1. Extracted some function into utils: https://github.com/Emurgo/yoroi-frontend/pull/210/files#diff-e05e93fdfc07fce3b2d6a5b92083def5
2. Created test-config to be used in tests: https://github.com/Emurgo/yoroi-frontend/pull/210/files#diff-a2e926e9dcc51b770ec51e30643a9a5b
3. Added unit-tests for these new functions in utils: https://github.com/Emurgo/yoroi-frontend/pull/210/files#diff-b45881d223bc76e38c5909a06d0cf197

Ready for review and merge

nicarq:
awesome! we have jest now :D

PR #211, opened 2019-01-03
ID calculation for newly sent transactions

vantuz-subhuman:
1. Call to `yoroi-backend-api.js/sendTx` now returns instance of `SignedResponse` with the `txId` field.
2. `adaNewTransactions.js/newAdaTransaction` and `trezorNewTransactions.js/newTrezorTransaction` are returning the same result instance
3. `api/ada/index.js/newAdaTransaction` and `api/ada/index.js/sendTrezorSignedTx` also return the same result instance

It was necessary to implement `CborIndefiniteLengthArray` in utils, because `cbor.encode(cbor.decode(data))` genuinely returns result different from initial data.

### DIFF
--- a/app/api/ada/lib/utils.js
+++ b/app/api/ada/lib/utils.js
@@ -11,6 +11,8 @@ import type {
 import {
   stringifyError
 } from '../../../utils/logging';
+// eslint-disable-next-line
+import { Blake2b } from 'rust-cardano-crypto';
 
 export const localeDateToUnixTimestamp =
   (localeDate: string) => new Date(localeDate).getTime();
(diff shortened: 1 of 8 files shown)


vantuz-subhuman:
image

app/api/ada/lib/utils.js
```diff
@@ -65,6 +67,48 @@ const _getTxCondition = (state: string): AdaTransactionCondition => {
   return 'CPtxWontApply';
 };
 
+export class CborIndefiniteLengthArray {
```
nicarq: I think we don’t need export for this one

app/api/ada/lib/utils.js
```diff
@@ -65,6 +67,48 @@ const _getTxCondition = (state: string): AdaTransactionCondition => {
...
+      attributes
+    ]);
+    // eslint-disable-next-line
+    return Buffer.from(Blake2b.blake2b_256(enc)).toString('hex');
+  } catch (e) {
+    throw new Error('Failed to convert raw transaction to ID! ' + stringifyError(e));
+  }
+}
+
+function decodedTxToAux(decodedTx) {
```
nicarq: how painful is to add cbor types for this one?

app/api/ada/lib/utils.js
vantuz-subhuman: Agree. Removed

app/api/ada/lib/utils.js
vantuz-subhuman: Added some type-descriptors, but the point is that this function still has to take `any` as its parameter, and then force-cast the result type based on the array length. There's no better way to check the cbor-type =\

PR #234, opened 2019-01-15
Adding a spec for memos

JizKadasha:
A memo spec.

### DIFF
--- a/docs/specs/code/TRANSACTION_MEMO.md
+++ b/docs/specs/code/TRANSACTION_MEMO.md
@@ -51,27 +51,16 @@ A memo could include sensitive or personal information about either the sender, 
 
 ## Encryption
 
-To safely store data in the external service we have to use encryption and digital signatures. However, before going any further we must notice an important limitation about **hardware wallets**. For security reasons, hardware wallets only provide access to the device's public keys, **NOT** private keys. To address this, we discuss two approaches:
+To safely store data in the external service we have to use encryption and digital signatures. For this purpose, we refer to the [key derivation]() specification to create a private key chain that allows us to sign and encrypt with different keys.
 
-### Public key encryption & password encryption
+Once the keys are generated, we refer to the corresponding Yoroi specifications for [encryption](https://github.com/Emurgo/yoroi-frontend/blob/737595fec5a89409aacef827d356c9a1605515c0/docs/specs/code/ENCRYPT.md) and [message signing](https://github.com/Emurgo/yoroi-frontend/blob/336f763a7b7085e2887b4965a979ccd24719787a/docs/specs/code/SIGNING.md). 
 
-This approach relies on using solely the *account* public key (available in all wallets) as input to create a new private key chain. More specifically:
-
-* Take the **Public Deriver** public key (*aka* account public key) and hash it with `Blake2b`.
-* Use the result from the previous step as a seed to a key derivation function, which will generate a new key pair. 
-* Use a public key child derivation function with the key pair from the previous step to derive 2 new child keys.
-* From the resulting key pairs, use one to encrypt and the other to sign.
-
-For both encrypting and signing we refer the reader to the corresponding Yoroi specifications for [encryption](https://github.com/Emurgo/yoroi-frontend/blob/737595fec5a89409aacef827d356c9a1605515c0/docs/specs/code/ENCRYPT.md) and [message signing](https://github.com/Emurgo/yoroi-frontend/blob/336f763a7b7085e2887b4965a979ccd24719787a/docs/specs/code/SIGNING.md). More specifically, the signing specification describes an adapted version of the [CBOR](https://tools.ietf.org/html/rfc7049) Object Signing and Encryption ([COSE](https://tools.ietf.org/html/rfc8152)) specification. In our particular case, we must consider the following assumptions:
+The signing specification describes an adapted version of the [CBOR](https://tools.ietf.org/html/rfc7049) Object Signing and Encryption ([COSE](https://tools.ietf.org/html/rfc8152)) specification. In our particular case, we must consider the following assumptions. Let `SK` and `EK` be the keys from the private key chain with the purpose of signing and encrypting, then:
 
 * A1: We sign with a single key (SIGN_TAG = 18)
-* A2: We use the `sign` and `verify` methods provided by Cardano WASM bindings.
-* A3: Based in A2 we don't need to include the optional parameteres described in point `P1` (from the signing spec) as we already know which algorithm is used.
-* A4: Encryption with the recipient's public key (COSE_Encrypt) is better suited for multiple receivers. For the sake of simplicty, we use password-based encryption (COSE_Encrypt0) with the private key as the password for symmetric encryption.
-
-A Proof-of-Concept can be found [here](https://repl.it/repls/CorruptInexperiencedLinuxpc).
-
-Observation: the encryption method described in the Yoroi encryption specification is implemented in the Cardano JS WASM bindings and it's called `password_encrypt`.
+* A2: We use `SK` along with the `sign` and `verify` methods provided by Cardano WASM bindings.
+* A3: Based in A2, we don't need to include the optional parameteres described in point `P1` from the signing spec as we already know which algorithm is used.
+* A4: Encryption with the recipient's public key (COSE_Encrypt) is better suited for multiple receivers. For the sake of simplicty, we use password-based encryption (COSE_Encrypt0) with `EK` as the password for symmetric encryption. Note that the encryption method described in the Yoroi encryption specification is implemented in the Cardano JS WASM bindings and it's called `password_encrypt`.
 
 The following are examples of signing and then encrypting a memo following the Cardano signed message format. To make the examples easier to read, they are represented using the extended CBOR diagnostic notation (CDDL).
 
@@ -91,7 +80,7 @@ The following are examples of signing and then encrypting a memo following the C
 )
 ```
 
-In this case, the `payload` is signed with the `sign` WASM binding using one of the key pairs obtained from the private key chain. Then, the output is added in the `signature` field. After the COSE structure is constructed, it is encrypted using `password_encrypt`, and the resulting output is used as a value for the ciphertext field in the COSE structure for encrypted messages.
+The `payload` is signed with the `sign` WASM binding using `SK`. The signature is added in the `signature` field and the structure is encrypted using `password_encrypt` and `EK` as a password. The result is used as a value for the ciphertext field in the COSE structure for encrypted messages. 
 
 **Cardano encrypted message**
 ```
@@ -109,56 +98,9 @@ In this case, the `payload` is signed with the `sign` WASM binding using one of 
 )
 ```
 
-The above output is saved to a file named after the `Blake2b` hash of the corresponding transaction id and uploaded to the external storage. To decrypt we do the opposite process.
+The structure is saved to a file named after the `Blake2b` hash of the transaction id for which the memo is intended and uploaded to the external storage. To decrypt we do the opposite process. 
 
-### User password encryption
-
-The second approach is also based on the Yoroi [encryption](https://github.com/Emurgo/yoroi-frontend/blob/737595fec5a89409aacef827d356c9a1605515c0/docs/specs/code/ENCRYPT.md#encryption) specification, which uses `password_encrypt` implementation from WASM bindings. The password encryption is used to encrypt at the time of sending a transaction with a memo and to decrypt at the time of showing an existing memo. 
-
-Currently, the user-flow for a standard Yoroi wallet already uses a **spending password** to decrypt the encrypted master key of the wallet. On hardware wallets, however, the spending password is not required as the wallet handles the private keys internally. That being said, we propose to use the existing spending password to derive a new **memo password** for symmetric encryption/decryption. We prefer not to use the *spending password* itself to encrypt/decrypt memos because we want to avoid compromising the security of the spending password in any way. 
-
-We define the memo password as the result of `Blake2b` applied to the concatenation of the wallet account public key and the spending password. The memo password is encrypted using the spending password and saved to local storage.
-
-In terms of user interaction in standard Yoroi wallets, the user will continue to use the spending password to send transactions with memos without noticing any change, as the same password will decrypt the master key for create the transaction and decrypt the memo password used to encrypt memos. To read the memo of an existing transaction, the user will have to introduce his/her spending password.
-
-For hardware wallets, we propose to imitate the user-flow used in standard wallets. That is, the user will have to introduce his/her spending password at the time of sending a transaction **if and only if** he/she decides to add a memo. Similarly, the user will be asked for the spending password when reading a memo associated with a transaction.
-
-Regarding the signing process, notice that `ChaCha20Poly1305` includes a MAC which can be used to verify *integrity* and *authenticity* of the data. For the sake of simplicity, we consider that signing the memo to prove *ownership* is not necessary, as having a valid memo password implies having access to the spending password and the specific wallet (with its account public key).
-
-A Proof-of-Concept can be found [here](https://repl.it/repls/DimgreyImmaterialDebuggers).
-
-We use the Cardano encrypted message format in the same way as before. Similarly, the ciphertext is the result of applying 'password_ecnrypt' to the content of the memo.
-
-**Cardano encrypted message**
-```
-16(
-  [
-    / protected / h'' / {
-      \ is_ascii \ is_ascii:false \
-    } / ,
-    / unprotected / {
-      \ enc_type \ enc_type:1 \
-      \ version \ version:1 \
-    } /, 
-    / ciphertext / h'56294e41264b6eeaf9952f7e452ef05bdbeb8b3985a5ca05a88e848ce989483edb3783d92ba0c04cbdd4911825a3c0aebab032513ca4caea553d741434582969d356cd380587361b'
-  ]
-)
-```
-
-The above output is saved to a file named after the `Blake2b` hash of the corresponding transaction id and uploaded to the external storage. To decrypt we do the opposite process.
-
-### Public key vs User password
-
-We make a brief comparison of both methods advantages/disadvantages.
-
-| # | Description           | Public key | User password |
- |-------|-----------------------|------------|-------|
- | 1  | Stores seed/password encrypted | NO | YES |
- | 2  | Signs the data to prove ownership | YES | NO |
- | 3  | Verifies authenticity & integrity | YES | YES |
- | 4  | Requires the user to enter a password to add/view memos | NO | YES |
- | 4  | Adds extra steps to hardware wallets users | NO | YES |
- | 5  | Available in Yoroi existing libraries | YES | YES |
+**Observation**: Note that if the user needs to decrypt the file on his/her own, we only need to provide him/her with the `EK` used to encrypt, along with the file and a description of the algorithms involved. There is no need to give information about the private key chain itself, such as the master key or the mnemonics used to generate it.
 
 ## External storage
 
@@ -173,6 +115,18 @@ The proposed external storage is Dropbox. Based on its [Javascript SDK](https://
 Using Dropbox will require to [create an app](https://www.dropbox.com/developers/reference/developer-guide). The user will have to connect his/her account and grant permissions, after which an [access token](https://www.dropbox.com/developers/reference/oauth-guide) will be generated. This access token must be safely stored locally in order to make authorized requests on behalf of the user (refer to the [encryption](https://github.com/Emurgo/yoroi-frontend/blob/737595fec5a89409aacef827d356c9a1605515c0/docs/specs/code/ENCRYPT.md) spec). 
 
 Note: API rate limits will apply in a [per user/app basis](https://www.dropbox.com/developers/reference/data-ingress-guide), so it shouldn't be a concern for Yoroi itself.
+
+As for the folder structure, we propose the following:
+
+```
+cardano / network_name / feature
+```
+
+For example:
+
+```
+cardano / mainnet / transaction-memos
+```
 
 ### Fault tolerance
 
@@ -190,20 +144,22 @@ Therefore, we must consider storing an error message field along with the memo w
 
 We need to save the following items into local storage:
 
-* LS1: `memoPassword` (string): encrypted memo password. This password is used to encrypt/decrypt memo files. The memo password itself is encrypted/decrypted using the spending password (password-based strategy).
-* LS2: `dropboxAccessToken` (string): access token provided by Dropbox after the user granted access to the Yoroi app. This token is used to connect to Dropbox API on behalf of the user.
+* LS1: `EXTERNAL_STORAGE` (JSON): dictionary with the supported external services, in which each element is named after the service (e.g. `dropbox`) and has a `token` field in it. In the case Dropbox, this will be the access token provided after the user granted access to the Yoroi app, used to connect to Dropbox API on behalf of the user.
 
-As for the memo, the easiest would be to save it (encrypted) in a new column in the transactions table. Every time a transaction is made, the memo is encrypted an saved along with the transaction and then uploaded to Dropbox. Every time a transaction is received, there is an API call to check for a memo for that transaction, in which case is downloaded and saved. Lastly, as mentioned in the previous sections, there could be several sources of error when trying to save the memo to the external storage. In that case, the fields would be:
+As for the memos, to avoid unnecessary coupling of data, we propose a new IndexDB table with the following set of columns:
 
-* LS3: `Memo`: string
-* LS4: `IsMemoPending`: boolean
-* LS5: `MemoErrorMessage`: string
+* `MemoId`: serial (PK)
+* `TransactionId`: int (FK)
+* `Memo`: varchar(256)
+* `IsPending`: boolean
+* `ErrorMessage`: varchar(256)
+
 
 ## User flow
 
-The proposed user flow is the following: (TODO update according to the chosen encryption strategy)
+The proposed user flow is the following: 
 
-1. The user opens the send transaction dialog.
+1. The user opens the send transaction dialog or clicks on the summary of a received transaction.
 2. The memo textarea is closed by default. The user can click *"Add memo"* label and one of the following scenarios occurs:
    - If the user hasn't synced his external storage account yet, he/she will see the text "In order to add private memos to   transactions you need to link Yoroi to a Dropbox account." and a link to the settings page.
    		- In the settings page, the user can click *"Connect Yoroi to Dropbox"*.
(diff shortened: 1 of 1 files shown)


JizKadasha:
A VERY important thing: I've noticed that app address can be changed. It was the same for every time I've tried to reinstall the app I've received this address: chrome-extension://bflmcienanhdibafopagdcaaenkmoago/main_window.html#/

Today I was investigating about Dropbox concurrency, so I got two instances of the yoroi app in different windows for different Chrome user. For the second user the address wasn't the same as for the first one.

This means we can't use Dropbox and maybe can't use another external storages with Oauth2. The only way how to fix the situation in my opinion is to create some website to which we can redirect user to make all auth things and which will redirect user back to yoroi after auth. I will do additional research, but for now the state of things is really pessimistic.

JizKadasha:
Update: according to the latest research, this happens only in dev mode, when we load an unpacked extension.

> Once uploaded to the Chrome Web Store, your extension ID is fixed and cannot be changed any more.

JizKadasha:
No diagram here, but an advanced description and the result of the latest researches about the encoding and so languages which can be used for memos.

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,87 @@
...
+   - If it isn't empty, user will see the `/components/wallet/send/WalletCloseMemoDialog.js` dialog window with the text "Remove memo?" and two buttons, "Yes" and "No".
+     - If user clicks "No", the dialog closes and the memo textarea remain opened and have as content user's memo.
+     - If user clicks "Yes", the dialog window closes and the textarea become cleared and closes too.
+5. If user have entered the memo and makes a transaction, memo will be saved to localStorage and also to user's external storage.
+6. When user goes to his summary - `/containers/wallet/WalletSummaryPage.js` and clicks on the transaction to expand it, he or she will see his memo at the very bottom of Transaction item (if memo exists).
+7. When user links his external storage account to Yoroi, we check if there are some memos in his external storage. If there are, we download a content of each memo file via Dropbox API, read the content with FileReader object, then we save these memos to localStorage. In this way we can make user be able to install Yoroi on the another device, link his or her external storage account and get user's memos so they can be displayed in WalletSummary.
```
SebastienGllmt: I don't think we can only check this when they link their account as if the user already has 2 wallets (A & B) on 2 different computers, making a memo on computer A will not make it show up on computer B.

I think we have to check either periodically or every time a new transaction is detected (not created by our wallet).

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Tech spec
+
+#### External storages
+
+We use official [Dropbox Javascript SDK](https://github.com/dropbox/dropbox-sdk-js#readme) for authenticating and making requests to Dropbox API. We can create folders, files, read them and change the content.
+
+To make us able to auth user and manipulate his or her files, we need to [create an app in Dropbox](https://www.dropbox.com/developers/apps/create?_tk=pilot_lp&_ad=topbar5&_camp=create). When we create an app, we can get a `client_id` and we must provide a `redirect_uri` because Dropbox uses Oauth2. Both `client_id` and `redirect_uri` are used to create a link for user, which user should click to authenticate.
```
SebastienGllmt: Looks like Dropbox uses an app-user pair for rate limiting so it should be okay to register our app and not worry 👍

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,87 @@
...
+Cons:
+- If user has 10,000 transactions with memos, we'll have to read through Dropbox api 10,000 files, which is bad for application performance.
+
+As for now, the file's name is a transaction id, and the content is the actual memo. In the localStorage memos are saved as an array of objects, where each object looks like this: `{ memoId: someTransactionId, memoText: actualMemoText }`.
+
+#### Concurrency
+
+The border case #1: the user has multiple instances of the app and both are linked to the same Dropbox account
+
+The result of additional researches is: if the user tries to create a new transaction from two instances at the same time, the worst case is Dropbox will create a conflicting copy of the memo file, so we'll have two files with memos, and the last saved version will become a conflicting copy. It's really good for us because we can read both files and merge them to one.
```
SebastienGllmt: That means Yoroi would also have to poll existing memos in case they became conflicting and had their contents updated, right? Sounds kind of tricky.

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,87 @@
...
+- It's really easier and requires less code when we know that each file has the only one memo as it's content;
+- There can't be any issues with concurrency when user has multiple instances of the app and both are linked to the same Dropbox account;
+Cons:
+- If user has 10,000 transactions with memos, we'll have to read through Dropbox api 10,000 files, which is bad for application performance.
+
+As for now, the file's name is a transaction id, and the content is the actual memo. In the localStorage memos are saved as an array of objects, where each object looks like this: `{ memoId: someTransactionId, memoText: actualMemoText }`.
+
+#### Concurrency
+
+The border case #1: the user has multiple instances of the app and both are linked to the same Dropbox account
```
SebastienGllmt: If the wallets are different the transactions IDs will be different so they will never conflict. We only have to worry about same wallet on multiple machines I think.

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,87 @@
...
+When we read a content of Dropbox file, we need to read `Blob` to get the actual content, so we use a `FileReader` object.
+
+#### Storage model
+
+For now we have a separate file for each transaction memo. This approach has it's own pros and cons.
+Pros:
+- It's really easier and requires less code when we know that each file has the only one memo as it's content;
+- There can't be any issues with concurrency when user has multiple instances of the app and both are linked to the same Dropbox account;
+Cons:
+- If user has 10,000 transactions with memos, we'll have to read through Dropbox api 10,000 files, which is bad for application performance.
```
SebastienGllmt: I think we can optimize by only making Dropbox queries for transaction summaries that are visible. Since we don't show the entire wallet history at once, we don't have to pull all memos at once.

We may still have to pull all memos overtime even if they aren't displayed if we want them to show up in the "export transaction history" feature.

JizKadasha: I agree about making requests for only that summaries which are visible.

Do we want them to show up in this feature? Sorry, I don't know about this one.

docs/specs/code/TRANSACTION_MEMO.md
JizKadasha: Sorry, I can't get the idea: the user can have only one wallet per instance, so why we need to have his memos for wallet A when he uses the wallet B? Each wallet has it's own summary, so the data for wallet A won't be displayed for wallet B.

docs/specs/code/TRANSACTION_MEMO.md
JizKadasha: This thing is about "if we want to have a single file with memos, not one file for one memo." If two users try to edit the same file in the same time, we can have a conflict actually.

docs/specs/code/TRANSACTION_MEMO.md
JizKadasha: Yeah, we need to check if we have a conflict copy actually, because usually we don't. If we have, we gonna check both files, the initial one and the conflict copy, and merge them to one.

Right now this case isn't possible. It will be possible when we switch to one file for all memos, or one file for one wallet.

docs/specs/code/TRANSACTION_MEMO.md
SebastienGllmt: Oh sorry my bad. I meant same wallet on 2 different computers

JizKadasha: Aha, you're right. To check periodically is easier, but more neat solution as for me would be to check every time a new transaction is detected. I think we need to use something like websockets, but all the suggestions are welcome.

vantuz-subhuman: Latest discussion with @JizKadasha:
- When our local wallet is not active - we will locally store decrypted copy of the file and some "remote file version ID" (Which is either latest known upload timestamp, or hash. Hash is better, cuz ensures content-equality.) and on every event `E` we will query remote to compare local version ID with current version ID, and if those don't match - replace local copy with remote (download and decrypt). Where `E` might be either a complex event detection (e.g. when we see new transaction where we are the sender, but we received it from outside - this means there's some background action with another copy of the same walet), or just on every transaction history update, or some time-preiod. Keeping remote file version ID makes check fairly lightweight.
- When we add new memo - we add it to local copy of the file, then encrypt the file, and then upload it to remote, at this point Dropbox API will potentially save our file as "conflicting copy", at which point we need to download current remote file, merge them and upload again. Background wallet B then assumed to eventually "background sync" with the new merged file, by checking that its local version ID does not match.

vantuz-subhuman: @JizKadasha, plz update the flow according to these latest discussions and then it will be 👌

docs/specs/code/TRANSACTION_MEMO.md
vantuz-subhuman: Latest discussion consensus: we keep **1 file per public key** (so basically per unique wallet). So automatically when user restored the same wallet on another computer/browser - it will reference the same file.
- File name: `SHA1(publicKey)`
- File format: `JSON`

File might just contain a single object dictionary from transaction IDs to memos as UTF-8 strings. Locally Yoroi stores the whole memo "file" as object in memory. Every time before upload stringified representation is encrypted using the public-key as password, and resulting HEX string is uploaded to remote.

leivaburto:
[ch1867]

SebastienGllmt:
- Use public deriver pubkey as entropy for a new private key chain to sign/encrypt your menmonics. The format for the signature is as defined in the message signing spec and you include an index inside the header. Chain has this format

root / sign_or_encrypt / pubderiver_index / header_index

SebastienGllmt:
Maybe investigate what they do in https://blog.trezor.io/featurefriday-labeling-in-trezor-wallet-a3e9c16c82ef

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,156 @@
...
+
+* A1: We sign with a single key (SIGN_TAG = 18)
+* A2: We use the `sign` and `verify` methods provided by Cardano WASM bindings.
+* A3: Based in A2 we don't need to include the optional parameteres described in point `P1` (from the signing spec) as we already know which algorithm is used.
+* A4: Encryption with the recipient's public key (COSE_Encrypt) is better suited for multiple receivers. For the sake of simplicty, we will use password-based encryption (COSE_Encrypt0) where the ciphertext is obtained as described in the Yoroi [encryption spec](https://github.com/Emurgo/yoroi-frontend/blob/737595fec5a89409aacef827d356c9a1605515c0/docs/specs/code/ENCRYPT.md) with the hash of the shared secret as the password.
```
SebastienGllmt: > with the hash of the shared secret as the password

I think this is wrong since the shared secret is meant to deal with encryption using "Public key based encryption". We can generate a pubkey for every memo and use technique if that's what you want, but you mention using the "password-based encryption" instead. Which do you want?

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,156 @@
...
+	- If the user clicks on *"Discard the memo"*, the memo will be removed from local storage.
+	- If the user clicks on *"Try again"* Yoroi will try to upload the file again. The message will remain until the message is successfully uploaded.
+5. When the user goes to his summary of transactions and clicks on a transaction to expand it, he/she will see the transaction memo at the very bottom of the transaction item (if the memo exists). The user will also see an option to remove the memo. If the user clicks it, a dialog message will show up with a message and two buttons showing *"Yes/No"*. The message is: *"Are you sure you want to remove the memo for this transaction? Doing so will erase all copies of it and you will not be able to restore it in the future"*.
+6. When the user goes to the "Connect Yoroi to Dropbox" subsection in the settings page, he/she will see a button to connect his/her account if it's not connected already. 
+7. When the user restores a wallet, he/she will be shown a message: *"If you previously connected your Dropbox account to use transactions memos, then you can restore them by going to Settings -> Connect Yoroi to Dropbox"*
+
```
SebastienGllmt: We also need a button on the settings screen to reveal the encryption password for the memos

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,156 @@
...
+   		- In the settings page, the user can click *"Connect Yoroi to Dropbox"*.
+   - If the user has already synced his external storage account, the memo textarea will be shown.
+3. If the user clicks *"Add memo"* again:
+   - If the textarea is empty, it will be closed.
+   - If it isn't empty, the user will see a dialog window with the text *"Do you want to discard this memo?"* and two buttons showing "Yes/No".
+     - If the user clicks *"No"*, the dialog closes and the memo textarea remains opened with the same content.
+     - If the user clicks *"Yes"*, the dialog window closes and the textarea become cleared.
+4. If user has entered the memo and makes a transaction, the memo will be saved to local storage and then to external storage. If there is a problem saving to external storage, then the user will see an error message: *"An error occurred while syncing with your Dropbox account. You can try to sync again or discard the memo"*. In that case, one of the following scenarios occurs:
```
SebastienGllmt: what is the format for localstorage? It isn't described anywhere. Do we save an indexdb table? As an object in `storage.local`?
Note: if we save in `storage.sync` then it will automatically sync with their browser storage  (but it sounds like it's not well supported on Brave and other Chrome forks)

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,156 @@
...
+6. When the user goes to the "Connect Yoroi to Dropbox" subsection in the settings page, he/she will see a button to connect his/her account if it's not connected already. 
+7. When the user restores a wallet, he/she will be shown a message: *"If you previously connected your Dropbox account to use transactions memos, then you can restore them by going to Settings -> Connect Yoroi to Dropbox"*
+
+# Requirements
+
+We consider the following requirements:
+
+* The user must be able to connect Yoroi to Dropbox, which will ask to grant permissions and it will return a new access token if the permission is granted. The access token must be encrypted and saved to local storage.
+* The user must be able to add or dismiss memos when sending a transaction.
+* The memo must be encrypted and saved to local storage after the transaction is sent. The stored object consists on the encrypted content of the memo along with a *boolean* value indicating if the memo has been uploaded/downloaded.
```
SebastienGllmt: I'm confused about this boolean. I thought you wanted that memos are only stored locally if they've been successfully uploaded as mentioned earlier in this spec.
But I think it's probably better to periodically iterate un-uploaded memos and try to upload them (which does have a chance to conflict if you added a memo in a different device in the meantime)

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,156 @@
...
+* After the memo is saved to local storage, it must be uploaded to the external storage. To upload a file, a [FilesCommitInfo](https://dropbox.github.io/dropbox-sdk-js/global.html#FilesCommitInfo) object must be constructed, where `contents` is the stored encrypted content of the memo and `path` is a concatenation of the wallet folder and the hash of the transaction.
+* The name of the folder in the external storage is given by the wallet's account number plate. If a user synchronizes the same wallet on more than one device, then each device must sync with the same folder.
+* Memos must be fetched after any of the following events:
+	* A new transaction is sent.
```
SebastienGllmt: Why fetch when a transaction is sent? Do you mean upload?

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,156 @@
...
+
+# Requirements
+
+We consider the following requirements:
+
+* The user must be able to connect Yoroi to Dropbox, which will ask to grant permissions and it will return a new access token if the permission is granted. The access token must be encrypted and saved to local storage.
+* The user must be able to add or dismiss memos when sending a transaction.
+* The memo must be encrypted and saved to local storage after the transaction is sent. The stored object consists on the encrypted content of the memo along with a *boolean* value indicating if the memo has been uploaded/downloaded.
+* After the memo is saved to local storage, it must be uploaded to the external storage. To upload a file, a [FilesCommitInfo](https://dropbox.github.io/dropbox-sdk-js/global.html#FilesCommitInfo) object must be constructed, where `contents` is the stored encrypted content of the memo and `path` is a concatenation of the wallet folder and the hash of the transaction.
+* The name of the folder in the external storage is given by the wallet's account number plate. If a user synchronizes the same wallet on more than one device, then each device must sync with the same folder.
```
SebastienGllmt: I like the number plate  idea 👍 
but do we want these all to be in the root dropbox folder? Maybe some `cardano-memos` root-level folder?

leivaburto: Absolutely 👍

docs/specs/code/TRANSACTION_MEMO.md
leivaburto: Oh yeah, I meant *memos must be synchronized* (uploaded/downloaded).

docs/specs/code/TRANSACTION_MEMO.md
leivaburto: The reason is because you could either have memos in local storage that you added (and hasn't been uploaded) or you downloaded (e.g. new transaction from other device), so when you check for un-uploaded memos you know which one to upload.

leivaburto: You are right about the chance of conflict :/ In that case we would have to decide which one to keep (by setting a write mode when uploading). It could be **first come, first served** (don't overwrite existing files).

docs/specs/code/TRANSACTION_MEMO.md
leivaburto: Maybe I didn't explain it well. I took the idea from the [signing spec](https://github.com/Emurgo/yoroi-frontend/pull/892#discussion_r312735229). *"single-receiver pubkey-based encryption with COSE_Encrypt0"* 🤔  I didn't elabore it further thinking this was eventually going to be updated in the signing spec.

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,206 @@
...
+* After the memo is saved to local storage, it must be uploaded to the external storage. To upload a file, a [FilesCommitInfo](https://dropbox.github.io/dropbox-sdk-js/global.html#FilesCommitInfo) object must be constructed, where `contents` is the stored encrypted content of the memo and `path` is a concatenation of the wallet folder and the hash of the transaction.
+* Memos must be uploaded with a **first come, first served** strategy. That is, if an out-of-sync wallet uploads a duplicated memo, the first copy is kept and the conflicting copy is discarded.
+* The name of the folder in the external storage is given by the wallet's account number plate. If a user synchronizes the same wallet on more than one device, then each device must sync with the same folder.
+* Memos must be uploaded/downloaded after any of the following events:
+	* A new transaction is sent (upload).
+	* A new transaction is received (download), in which the originating address is the current account. This is intended for transactions that are originated from other devices but in the same wallet.
```
leivaburto: A user may want to add a memo to a received transaction in which he is not the sender? (*labeling*)

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,239 @@
...
+)
+```
+
+The above output is saved to a file named after the `Blake2b` hash of the corresponding transaction id and uploaded to the external storage. To decrypt we do the opposite process.
+
+### Public key vs User password
+
+We make a brief comparison of both methods advantages/disadvantages.
+
+| # | Description           | Public key | User password |
```
SebastienGllmt: as Shin said, we can't use `user password` since the password could change

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,239 @@
...
+
+For hardware wallets, we propose to imitate the user-flow used in standard wallets. That is, the user will have to introduce his/her spending password at the time of sending a transaction **if and only if** he/she decides to add a memo. Similarly, the user will be asked for the spending password when reading a memo associated with a transaction.
+
+Regarding the signing process, notice that `ChaCha20Poly1305` includes a MAC which can be used to verify *integrity* and *authenticity* of the data. For the sake of simplicity, we consider that signing the memo to prove *ownership* is not necessary, as having a valid memo password implies having access to the spending password and the specific wallet (with its account public key).
+
+A Proof-of-Concept can be found [here](https://repl.it/repls/DimgreyImmaterialDebuggers).
```
SebastienGllmt: `entropyToMnemonic` generates a too long password for the user and it may be confusing since they may think the 15-words is their mnemonic.

SebastienGllmt: I would avoid using `bip44_account`. Instead let's pick a `purpose` (like Bip44 reserves 44) that's not taken yet and document that.

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,239 @@
...
+* The user exceeded the limit of API calls to the external storage
+
+Therefore, we must consider storing an error message field along with the memo when saving it to local storage. Also, notice that when a memo fails to be uploaded, another instance of the same wallet but in a different device could add a memo for the same transation. In this case, we apply the *first come, first served* principle and keep the first memo that was uploaded. 
+
+## Local storage
+
+We need to save the following items into local storage:
+
+* LS1: `memoPassword` (string): encrypted memo password. This password is used to encrypt/decrypt memo files. The memo password itself is encrypted/decrypted using the spending password (password-based strategy).
+* LS2: `dropboxAccessToken` (string): access token provided by Dropbox after the user granted access to the Yoroi app. This token is used to connect to Dropbox API on behalf of the user.
```
SebastienGllmt: I would avoid the name `dropboxAccessToken` since we may have more  services in the future. Probably it's better to have some object with a identifier representing the service name (with only Dropbox for now) and a generic `token` field.

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,239 @@
...
+
+As for the memo, the easiest would be to save it (encrypted) in a new column in the transactions table. Every time a transaction is made, the memo is encrypted an saved along with the transaction and then uploaded to Dropbox. Every time a transaction is received, there is an API call to check for a memo for that transaction, in which case is downloaded and saved. Lastly, as mentioned in the previous sections, there could be several sources of error when trying to save the memo to the external storage. In that case, the fields would be:
+
+* LS3: `Memo`: string
```
SebastienGllmt: I would put these in a separate table instead of reusing an existing table. There's no advantage to coupling these (since you can do an SQL join to merge the tables) and it makes it harder to modify these tables in the future.

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,239 @@
...
+ | 3  | Verifies authenticity & integrity | YES | YES |
+ | 4  | Requires the user to enter a password to add/view memos | NO | YES |
+ | 4  | Adds extra steps to hardware wallets users | NO | YES |
+ | 5  | Available in Yoroi existing libraries | YES | YES |
+
+## External storage
+
+The proposed external storage is Dropbox. Based on its [Javascript SDK](https://github.com/dropbox/dropbox-sdk-js#readme), the API endpoints of our interest are:
+
+* [Create folder](https://dropbox.github.io/dropbox-sdk-js/Dropbox.html#filesCreateFolder__anchor): Create a folder at a given path.
```
SebastienGllmt: Need a format for this path

Maybe

"cardano" / "mainnet"  / "transaction-memos"

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Encryption
+
+To safely store data in the external service we have to use encryption and digital signatures. However, before going any further we must notice an important limitation about **hardware wallets**. For security reasons, hardware wallets only provide access to the device's public keys, **NOT** private keys. To address this, we discuss two approaches:
+
+### Public key encryption & password encryption
+
+This approach relies on using solely the *account* public key (available in all wallets) as input to create a new private key chain. More specifically:
+
+* Take the **Public Deriver** public key (*aka* account public key) and hash it with `Blake2b`.
```
SebastienGllmt: As I mentioned in Slack, the Public Deriver is not guaranteed to actually contain a public key. 
This is useful  for users  who want a “balance tracker” where you input a set of addresses and Yoroi monitors the balance for those addresses for you

For this case we would need to ask the user to input a memo password when they create their wallet and hash that as the entropy and then store that locally

docs/specs/code/TRANSACTION_MEMO.md
leivaburto: I added it in the user-flow section and as a note in the encryption section.

docs/specs/code/TRANSACTION_MEMO.md
leivaburto: That's right. Just for the record, we can't use a **spending-password** based strategy as the user could use different spending passwords across distinct instances of the same wallet.

docs/specs/code/TRANSACTION_MEMO.md
leivaburto: As discussed on Slack, it's not necessary to reveal the mnemonic itself.

docs/specs/code/TRANSACTION_MEMO.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Encryption
+
+To safely store data in the external service we have to use encryption and digital signatures. However, before going any further we must notice an important limitation about **hardware wallets**. For security reasons, hardware wallets only provide access to the device's public keys, **NOT** private keys. Then, the proposed approach relies on using solely the *account* public key (available in all wallets) as input to create a new private key chain. More specifically:
+
+* Take the **Public Deriver** public key (*aka* account public key) and hash it with `Blake2b`.
+* Use the result from the previous step as a seed to a key derivation function, which will generate a new key pair. 
+* Use a public key child derivation function with the key pair from the previous step to derive 2 new child keys.
```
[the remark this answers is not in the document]
leivaburto: We will continue the discussion about this in the key derivation spec: https://github.com/Emurgo/yoroi-frontend/pull/899

docs/specs/code/TRANSACTION_MEMO.md
leivaburto: We will continue the discussion about this in the key derivation spec: https://github.com/Emurgo/yoroi-frontend/pull/899

PR #241, opened 2019-01-17
Show fee & Intrawallet transactions

SebastienGllmt:
1) fees are explicitly shown in the transaction details
2) Sending to yourself is now considered an "intrawallet" transaction (this logic is taken from the mobile app)

image


app/api/ada/index.js
```diff
@@ -563,6 +571,77 @@ export default class AdaApi {
...
 // ========== TRANSFORM SERVER DATA INTO FRONTEND MODELS =========
 
+async function _getTxFinancialInfo(
+  data: AdaTransaction
+): Promise<{
+  type: TransactionType,
+  amount: BigNumber,
+  fee: BigNumber
+}> {
+  // Note: logic taken from the mobile version of Yoroi
```
nicarq: maybe it could be helpful to have more info from which file

app/components/wallet/transactions/Transaction.js
```diff
@@ -224,6 +252,10 @@ export default class Transaction extends Component<Props, State> {
               </div>
             )}
             <div>
+              <h2>
+                Fee:
```
nicarq: it’s a really small detail but I would hide the `:` to be consistent with the other ones

nicarq:
LGTM!

PR #266, opened 2019-01-28
Generate screenshots during tests

SebastienGllmt:
# Overview

This PR makes that all `cucumber-js` steps that contain "I should see" take a screenshot AFTER the step is done.

The problem with screenshot-based testing is that you may get very small pixel differences between baseline.
This is caused by UI elements like a spinner indicating a loading (the difference could just be the spinner at a different rotation)

The rationale for only taking screenshots on "I should see" tests is that these tests should BLOcK on UI elements and "see" a stable view of the UI after the blocking is done
This is exactly what we want to take a screenshot of to compare with.

# Changes made to `cucumber` library

Unfortunately Cucumber doesn't let you add a hook for every step (which is what we need)

To fix this, I did the following: 

1) Upgraded `cucumber-js` to 5.1.0 (changelog [here](https://github.com/cucumber/cucumber-js/blob/master/CHANGELOG.md))
2) Forked `cucumber-js` into my own repo
3) Copied [this PR](https://github.com/cucumber/cucumber-js/pull/1121) to my personal repo
4) Made changes to my branch to match NPM releae format of `cucumber-js` (notably, `lib` folder was required (otherwise in .gitignore)

# Future work

You can see for example in my PR, one of our tests said "I should see" but in fact it had no blocking UI call at all (it only checked the value of some variable)

Therefore, to effectively do screenshot-based testing, one would have to
1) Run the E2E-test a few times and see if any of the screenshot differ by small amounts (and then figure out how we can avoid it)
2) Go over all our tests to see which steps should be renamed to use "I should see" (for maximum screenshot coverage)

This PR doesn't make screenshot testing as part of our CI but that is the ultimate goal.

Ideally, our CI would cache the screenshot tests for specific commits. Then, when you make a PR, it checks at which commit you forked from `develop` and compare your PR to that (Note: you CANNOT compare to current `develop` as too many things may have changed. You have to compare to where the branch forked)

# Save Location

- Scenarios in their own folders
- Each step is its own file (step # as first part of key)
- Some tests have multiple "examples" that mean we run the same scenario multiple times. Therefore we also add the line number to the unique key of the file

*Note*: Screenshots folder is deleted every time you run a new test

# Example

### Test definition
```
  @it-8
  Scenario Outline: Wallet renaming (IT-8)
    When I am testing "Wallet Settings Screen"
    And There is a wallet stored named Test
    And I navigate to the general settings screen
    And I click on secondary menu "wallet" item
    And I click on "name" input field
    And I enter new wallet name:
    | name         |
    | <walletName> |
    And I click outside "name" input field
    And I navigate to wallet transactions screen
    Then I should see new wallet name "<walletName>"
    Examples:
    | walletName                               |                    |
    | first Edited                             |2 words name        | 
    |ウォレットの追加                             |Japanese            |
    |지갑 추가                                 | Korean             |
    |НАСТРОЙКИ                                 | Russian            |
    | a                                        |1-characters length |
    | asdfghjklpoiuytrewqazxcvbnmlkjhgfdsaqwer |40 characters length|
```

### Run test
```
npm run test-by-tag-chrome @it-8
```

### Created files

```
yoroi-frontend/screenshots/Wallet renaming IT8/10_96-I should see new wallet name .png
yoroi-frontend/screenshots/Wallet renaming IT8/10_97-I should see new wallet name .png
yoroi-frontend/screenshots/Wallet renaming IT8/10_98-I should see new wallet name .png
yoroi-frontend/screenshots/Wallet renaming IT8/10_99-I should see new wallet name .png
yoroi-frontend/screenshots/Wallet renaming IT8/10_100-I should see new wallet name .png
yoroi-frontend/screenshots/Wallet renaming IT8/10_101-I should see new wallet name .png
```
### Screenshot generated

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ production-key.pem
 build/
 dev/
 dll/
+screenshots/
 
 *.zip
 *.xpi
--- a/features/settings-ui.feature
+++ b/features/settings-ui.feature
@@ -158,9 +158,9 @@ Feature: Wallet UI Settings
     When I navigate to the general settings screen
     And I open General Settings language selection dropdown
     And I select Japanese language
-    Then I should see Japanese language as selected
+    Then The Japanese language should be selected
     When I refresh the page
-    Then I should see Japanese language as selected
+    Then The Japanese language should be selected
 
   @it-3
   Scenario: Yoroi Settings Screen / Terms of Use in Default English(IT-3)
--- a/features/step_definitions/common-steps.js
+++ b/features/step_definitions/common-steps.js
@@ -1,11 +1,27 @@
 // @flow
 
-import { BeforeAll, Given, After, AfterAll } from 'cucumber';
+import { Before, BeforeAll, Given, After, AfterAll, setDefinitionFunctionWrapper } from 'cucumber';
 import { getMockServer, closeMockServer } from '../support/mockServer';
 import { buildFeatureData, getFeatureData, getFakeAddresses } from '../support/mockDataBuilder';
 import i18nHelper from '../support/helpers/i18n-helpers';
 
+const { promisify } = require('util');
+const fs = require('fs');
+const rimraf = require('rimraf');
+
+const screenshotsDir = './screenshots/';
+
+/** We need to keep track of our progress in testing to give unique names to screenshots */
+const testProgress = {
+  scenarioName: '',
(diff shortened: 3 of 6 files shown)


sarveshwar-singh:
LGTM, cool !!

PR #284, opened 2019-02-07
[WIP] Version 1.1 for Theme Feature (based on Sebastien Theme Code)

Mercurial:
- I am converting the html.style CSS string into JSON string before saving into localStorage so it will be directly loadable by the ThemeManager via "variables" property

- I am manipulating SVG directly from React as Component to dynamically generate theme preview. (Which seems to be causing eslint to fail)

So we can argue that maybe we can replace SVG with pure HTML Elements to generate preview

- I also edited SVG code manually to be more responsive for aesthetic purposes

image

image


app/stores/toplevel/ProfileStore.js
```diff
@@ -93,6 +128,32 @@ export default class SettingsStore extends Store {
     await this.getProfileLocaleRequest.execute(); // eagerly cache
   };
 
+  _updateTheme = async ({ theme }: { theme: string }) => {
+    await this.unsetCustomThemeRequest.execute();
```
SebastienGllmt: I'm confused why you play around with custom theme here. I think both of these lines should be removed.

app/stores/toplevel/ProfileStore.js
```diff
@@ -93,6 +128,32 @@ export default class SettingsStore extends Store {
...
+      const attributes: any = html.attributes;
+      await this.unsetCustomThemeRequest.execute();
+      await this.setCustomThemeRequest.execute(attributes.style.value);
+      await this.getCustomThemeRequest.execute(); // eagerly cache
+    }
+  };
+
+
+  // TODO: not sure where to put this method (Clark)
+  getThemeVars = ({ theme }: { theme: string }) => {
```
SebastienGllmt: I think this function is doing the wrong thing. I should take your currently selected theme and merge in any custom overrides into it.

app/components/settings/categories/DisplaySettings.js
```diff
@@ -0,0 +1,132 @@
...
+          <button
+            type="button"
+            className={themeYoroiModernClasses}
+            onClick={selectTheme.bind(this, { theme: THEMES.YOROI_MODERN })}
+          >
+            <div>
+              <ThemeThumbnail themeVars={getThemeVars({ theme: THEMES.YOROI_MODERN })} />
+            </div>
+            <span>{intl.formatMessage(messages.themeYoroiModern)}</span>
+          </button>
```
SebastienGllmt: I assume this is only here 3 times for testing purposes? Maybe remove?
That said, before merging this PR, we have to disable the UI part of this feature anyway (since we don't have an actual "modern" ui to show. It is a dummy theme created to debug this feature).
We can make that this dummy only appears when build in dev mode or test mode to allow us to test this feature without exposing it to the public

app/components/settings/categories/DisplaySettings.scss
```diff
@@ -0,0 +1,86 @@
+$break-row: 1080px;
```
SebastienGllmt: I'm not a big fan of all the hard-coded stuff in this file. Why do we need these instead of setting up the flex container so all of this is handled for us?

app/assets/images/themes/yoroi-classic.inline.svg
```diff
@@ -0,0 +1,22 @@
+<svg width="580" height="360" xmlns="http://www.w3.org/2000/svg">
```
SebastienGllmt: I don't think we need this file anymore now that we have the `ThemeThumbnail`

app/stores/toplevel/ProfileStore.js
Mercurial: Hi this is a helper function that basically takes a `theme` name (one of the pre-built themes) and extract the CSS variables of the specific theme.

This is helpful specially when trying to generate the preview of the thumbnails because I can access the variables any theme.

So if I was trying to generate a preview of `YoroiModern` theme, I will utilize this function to get the `YoroiModern` variables and then dynamically change properties of the SVG for example based on those variables.

Please let me know what you think/criticism or maybe if I have misunderstood what you meant.

Thank You!

app/i18n/locales/en-US.json
```diff
@@ -77,8 +77,13 @@
   "profile.termsOfUse.checkboxLabel": "I agree with terms of use",
   "profile.termsOfUse.submitLabel": "Continue",
   "profile.termsOfUse.title": "Terms Of Use",
+  "settings.display.themeLabel": "Theme",
```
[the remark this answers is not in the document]
Mercurial: hello I committed updated translations but I am not sure if I did it correctly as I am not too familiar with the translation manager.

Please advice
Clark

app/components/settings/categories/DisplaySettings.scss
Mercurial: My idea is here is to always display 3 theme items per row and move the next items to the next row. And at 1080px and smaller since the layout starts to not look good it will then respond to that size and become mobile / tablet layout and display 1 theme per row with a vertical scrollbar.

I will humbly accept any criticism / suggestions 
Thanks
Clark

app/stores/toplevel/ProfileStore.js
```diff
@@ -93,6 +128,32 @@ export default class SettingsStore extends Store {
...
 
+  _updateTheme = async ({ theme }: { theme: string }) => {
+    await this.unsetCustomThemeRequest.execute();
+    await this.getCustomThemeRequest.execute(); // eagerly cache
+    await this.setThemeRequest.execute(theme);
+    await this.getThemeRequest.execute(); // eagerly cache
+  };
+
+  _exportTheme = async () => {
+    // TODO: It should be ok to access DOM Style from here
```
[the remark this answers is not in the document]
Mercurial: I will do my best to find an alternative more elegant solution

Thank you

app/stores/toplevel/ProfileStore.js
SebastienGllmt: Ah my bad I didn't notice this function was used in multiple places. That means the merge of the selected theme & custom theme should happen in `currentThemeVars`, correct?

app/stores/toplevel/ProfileStore.js
Mercurial: yep already committed it earlier today

app/stores/toplevel/ProfileStore.js
Mercurial: Yep! I added a new commit to improve and make this behavior more obvious as-well as optimized

Mercurial:
Hi Guys!
I have completed all the requested change by Sebastien except the DOM direct read access to get current customized styles.

Let me know what you think
Thanks
Clark

app/components/settings/categories/display/ThemeThumbnail.js
```diff
@@ -0,0 +1,33 @@
+import React, { Component } from 'react';
+
+type Props = {
+  themeVars: Object
+};
+
+export default class ThemeThumbnail extends Component<Props> {
```
SebastienGllmt: I think you can clean this up by saving this as an `svg` file and then applying the color transformation using css `style` overrides in the Javascript. Maybe not a big deal though?

app/stores/toplevel/ProfileStore.js
```diff
@@ -95,6 +134,39 @@ export default class SettingsStore extends Store {
...
+      const attributes: any = html.attributes;
+      await this.unsetCustomThemeRequest.execute();
+      await this.setCustomThemeRequest.execute(attributes.style.value,
+        this.getThemeVars({ theme: this.currentTheme }));
+      await this.getCustomThemeRequest.execute(); // eagerly cache
+    }
+  };
+
+
+  // TODO: not sure where to put this method (Clark)
```
SebastienGllmt: remove this TODO?

app/components/settings/categories/DisplaySettings.js
```diff
@@ -0,0 +1,124 @@
...
+              && hasCustomTheme() &&
+                <div className={styles.themeWarning}>
+                  {intl.formatMessage(messages.themeWarning)}
+                </div>)
+            }
+            <ThemeThumbnail themeVars={getThemeVars({ theme: THEMES.YOROI_CLASSIC })} />
+            <span>{intl.formatMessage(messages.themeYoroiClassic)}</span>
+          </button>
+
+          <button
```
SebastienGllmt: Could you make that the "modern" theme is disabled for production builds (mainnnet)? We need this to test the feature is working but we don't actually want to show this to production users. 
See `environment.js` for how to do this.

Mercurial: this is done now

thanks
Clark

Mercurial:
by the way, there is a specific test that fails because it expect to interact with a second theme but since we disabled it does fail. 

Just a heads-up
Clark

SebastienGllmt:
@Mercurial this is why I suggested you do "if not mainnet" instead of "if dev" that way the theme selection is still available during testing (this way the test passes)

Mercurial:
firefox tests seems to fail, not sure why though

PR #285, opened 2019-02-07
Feature/design change

NikeOnly:


SebastienGllmt:
Wallet details broken
image

No way to get to settings from wallet creation

TestnetWarningBanner throws unique key error

Strange errors:
```
transactionFeeError null
amountField.error null
```

Issue #287, opened 2019-02-08
[AddWalletPage] Footer current release sync plus improvements.


sarveshwar-singh:
**[A] SYNC**
1. Footer Item horizontal alignment has been changed from `center` to `left`
![image](https://user-images.githubusercontent.com/19986226/52473085-4dd13f80-2bd8-11e9-8c78-011133442597.png)

**[B] IMPROVEMENTS**
1. Better display mode switching logic, switch to all icon mode or all text plus mode.
![image](https://user-images.githubusercontent.com/19986226/52474584-41e77c80-2bdc-11e9-829b-cb7812240eec.png)
![image](https://user-images.githubusercontent.com/19986226/52474590-490e8a80-2bdc-11e9-9f21-814647483f16.png)
![image](https://user-images.githubusercontent.com/19986226/52474605-4f9d0200-2bdc-11e9-95b2-cd7be992f146.png)

2. Add on-hover tool-tip irrespective of the display mode, so when its displaying in icon only mode then user will have better idea (all tool-tip text can be same as what its displaying in icon with text mode)

3. Delete unwanted line: https://github.com/Emurgo/yoroi-frontend/blob/b0694650ec9a34750ac3d0114f7b511104436839/app/components/footer/FooterItem.js#L24

4. Delete dead code or implement all `Wallet` page footer:
https://github.com/Emurgo/yoroi-frontend/blob/b0694650ec9a34750ac3d0114f7b511104436839/app/containers/wallet/Wallet.js#L40
https://github.com/Emurgo/yoroi-frontend/blob/b0694650ec9a34750ac3d0114f7b511104436839/app/containers/wallet/Wallet.js#L46
(Initially it was decided to show `HelpLinkFooter`(common implementation) on `Add Wallet` page plus all `Wallet pages`, but for all `Wallet` pages it was temporarily suspended. Now we have separate implementation for `Add Wallet` (i.e `AddWalletFooter`),  so we can delete that code or implement footer for all all `Wallet` page.)

REF: https://github.com/Emurgo/yoroi-frontend/issues/110 https://github.com/Emurgo/yoroi-frontend/issues/111 https://github.com/Emurgo/yoroi-frontend/issues/205

PR #282, opened 2019-02-06
Refactoring the footer

ASUKA1006:
Fix #205 

Use HorizontalFlexContainer so it is much cleaner.

I also add "What is a hardware wallet?" which was previously disabled.

# Big screen

screen shot 2019-02-06 at 1 38 51 am

# Small screen

screen shot 2019-02-06 at 1 39 30 am

Issue #205, opened 2018-12-30
[AddWalletPage] HelpLinkFooter disappear when screen under 914px


nicarq:
<img width="917" alt="screen shot 2018-12-30 at 12 47 43 pm" src="https://user-images.githubusercontent.com/1622112/50546961-414ce500-0c31-11e9-96d9-a7f45d02c369.png">

sarveshwar-singh:
plus:
![image](https://user-images.githubusercontent.com/19986226/51675490-29d9ff80-2017-11e9-9f1d-650e45d519dd.png)

SebastienGllmt:
I would guess this is because of the upgrade to `react-router`

https://github.com/Emurgo/yoroi-frontend/pull/197/files

You can see I added the `height: '100%'` at two different places to make the UI work and this probably also got rid of the scrollbar.

Issue #111, opened 2018-11-19
on every page of not hardware integrated wallet add links to promote Hardware wallet


sarveshwar-singh:

sarveshwar-singh:
How to connect Trezor: https://youtu.be/Dp0wXwtToX0
Referral Link: https://yoroi-wallet.com/get-trezor

PR #289, opened 2019-02-09
fix of #287

naomegane:
**[A] SYNC**

1. [Fixed] Footer Item horizontal alignment has been changed from `center` to `left`

**[B] IMPROVEMENTS**

1. [Fixed] Better display mode switching logic, switch to all icon mode or all text plus mode.

2. [Fixed] Add on-hover tool-tip irrespective of the display mode, so when its displaying in icon only mode then user will have better idea (all tool-tip text can be same as what its displaying in icon with text mode)

3. [Deleted] Delete unwanted line: 
 
4. [Deleted] Delete dead code or implement all `Wallet` page footer:
 
REF: https://github.com/Emurgo/yoroi-frontend/issues/110 https://github.com/Emurgo/yoroi-frontend/issues/111 https://github.com/Emurgo/yoroi-frontend/issues/205 https://github.com/Emurgo/yoroi-frontend/pull/282

### DIFF
--- a/app/components/footer/FooterItem.js
+++ b/app/components/footer/FooterItem.js
@@ -21,13 +21,13 @@ export default class FooterItem extends Component<Props> {
 
     return (
       <div className={styles.component}>
-        {/* Buy a Trezor Hardware wallet */}
         {
           <a
             href={url}
             target="_blank"
             rel="noopener noreferrer"
             className={styles.block}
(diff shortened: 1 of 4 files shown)


SebastienGllmt:
Thanks! I'm kind of concerned with the media query being a fixed size as the text length can change in different languages but all the languages we offer right now fit so probably it's okay.

PR #290, opened 2019-02-09
Fix punctuation

ASUKA1006:
For each translation file I deleted the unnecessary question mark ( #287)

### DIFF
--- a/app/components/footer/FooterItem.scss
+++ b/app/components/footer/FooterItem.scss
@@ -22,9 +22,6 @@ $blockHeight: 52px;
     display: inline-block;
     vertical-align: middle;
   }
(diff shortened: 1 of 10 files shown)


sarveshwar-singh:
@ASUKA1006 @SebastienGllmt 

Initially it was decided to use `?` by following reasoning:
```
"How to do X" is fine. "How do I do X?" needs a question mark
So, "How to" is okay with no question mark.
and What needs question mark.
```

and i forget that reasoning, so had requested as point 2 of SYNC: https://github.com/Emurgo/yoroi-frontend/issues/287
but i silently deleted when recalled back :)

**I am ok with making it consistent without using `?`**

PR #269, opened 2019-01-28
"About Yoroi" section

vantuz-subhuman:
Closes #263 

image

Issue #263, opened 2019-01-27
"About Yoroi" section in settings


vantuz-subhuman:
## Problem

Can't see current Yoroi version and various stuff about it anywhere in UI (need to go to Chrome app management page for this).

## Abstract

We need a section where we can put any kind of info stuff about the app, but that would not be interfering with the main UX much.

## Idea

Create a special section "About Yoroi" in settings, and put this idea there:

1. Current version as link to changelog (e.g. "Yoroi version: [1.2.0](https://github.com/Emurgo/yoroi-frontend/releases/tag/1.2.0)")
2. Link to website: https://yoroi-wallet.com
3. Link to Emurgo blog: https://medium.com/@emurgo_io
4. Link to twitter: https://twitter.com/YoroiWallet
5. Link to GitHub: https://github.com/Emurgo/yoroi-frontend
6. Playlist about Yoroi on Emurgo YT: https://www.youtube.com/watch?v=GLNgpr-3t2E&list=PLFLTrdAG7xRZUmi04s44T1VEF20xKquF2
7. Link to community help channel: https://t.me/CardanoCommunityTechSupport

## UI

Because the "General" section in settings is very small atm and there's only the language selector - I would propose to add a text section on top there:

![image](https://user-images.githubusercontent.com/5585355/51801396-0ea50500-224e-11e9-8579-00c8b8fb1c57.png)

Section just similar to how it looks in "Support":

![image](https://user-images.githubusercontent.com/5585355/51801405-40b66700-224e-11e9-8d9c-da3fc30a0e96.png)

But with all the links instead.

If ever we got more settings for the "General" section - we can move "About" to it's own menu-item.

vantuz-subhuman:
Suggest any additional info we need to put there, plz

adamparish:
Great idea! This will make the user feel much more confident with the Yoroi Wallet.

sarveshwar-singh:
https://github.com/Emurgo/yoroi-frontend/pull/305

PR #291, opened 2019-02-09
About Yoroi block

ASUKA1006:
1) Removed the scrollbar for the footer (size was too small to fit SVG)
2) Adds About Us to the settings page

- [ ] Need SVG logo for EMURGO
- [ ] Yoroi logo is plain white so it is hard to see
- [ ] Review these are the links we want on the About Us
- [ ] Maybe left-align instead of center?
- [ ] #289 Added a media query with a size specifically for the add wallet page (5-element rows and full width of app). This means the text for our icons disappears even when there is a lot of space left. Probably we need to replace this media query with something smarter.

screen shot 2019-02-10 at 3 13 12 am


sarveshwar-singh:
(Was busy with something else, just followed: https://github.com/Emurgo/yoroi-frontend/issues/263 https://github.com/Emurgo/yoroi-frontend/pull/269 )
1. I strongly believe we should make it as new item rather than fitting somewhere else. 
Remove from `General`, add new item as `About Yoroi`, move content to `About Yoroi`
image

2. Add Version, Commit, Network (can find logic here: https://github.com/Emurgo/yoroi-frontend/blob/06e9e353da26609435af36e0b085e80f32a4f9b6/app/utils/logging.js#L50), with standard Yoroi styling (fonts)
image

3. Change `Yoroi version` to `Yoroi changelog` (@SebastienGllmt  or any better)

4. 
>  Need SVG logo for EMURGO

try out: https://github.com/Emurgo/yoroi-mobile/blob/a3d72218b1e63f6362152aae2f03c8763c168795/src/assets/EmurgoIcon.js#L10

5. 
> * Yoroi logo is plain white so it is hard to see

Yes, we can change it to black.
Following is how to customize it:
```
.yoroiLogo {
  & > svg {
    display: block;
    width: 100px;
    margin: 0 114px;
    & > path {
      fill: var(--theme-button-primary-background-color);
    }
  }
}
```  
https://github.com/Emurgo/yoroi-frontend/blob/develop/app/components/loading/Loading.scss
There is only white color logo is there in repo but we customize it depends upon need.

6. 
> * #289 Added a media query with a size specifically for the add wallet page (5-element rows and full width of app). This means the text for our icons disappears even when there is a lot of space left. Probably we need to replace this media query with something smarter.

was not aware the you are going to reuse that here.

I would suggest:
A. 
move this: https://github.com/Emurgo/yoroi-frontend/tree/develop/app/components/footer
to: https://github.com/Emurgo/yoroi-frontend/tree/develop/app/components/widgets
and give a more generic name rather than `FooterItem`
(`FooterItem` is not appropriate for Setting page)

B.
move this: https://github.com/Emurgo/yoroi-frontend/blob/develop/app/containers/footer/AddWalletFooter.js
to: https://github.com/Emurgo/yoroi-frontend/tree/develop/app/components/footer

also move `media query` from `FooterItem`(current) to here.


can directly use: `AddWalletFooter` at https://github.com/Emurgo/yoroi-frontend/blob/fe956dc9b035232ed653a1619a2839ff64010ade/app/containers/wallet/WalletAddPage.js#L107

7.
`app/containers/footer/GeneralSettingsFooter.js` should no be in footer from start.
It should should have been inside `app/containers/settings` but if you do `#1` and `#6` as i mentioned there will be new file `app/containers/settings/AboutYoroiPage.js` and mostly we don't need `app/containers/footer/GeneralSettingsFooter.js`

ASUKA1006:
For the request 1, 
I removed  `About Yoroi` from `General` and added a new item for `About Yoroi`. Also, I moved that content from `General` to it.
Now, the pages look like this.

[About Yoroi page]
screen shot 2019-02-13 at 12 23 59 am

[General page]
screen shot 2019-02-13 at 12 22 55 am

SebastienGllmt:
Here is the list from the marketing team for the URLs to include:

1) Yoroi Facebook needs to be added.
2) Yoroi Twitter and Facebook should go first in order
3) Delete Yoroi Version (replace with suggestion from Shin)
4) Replace "EMURGO Blog" with just "EMURGO" to emurgo.io
5) Add Medium blog link
5) Delete the telegram link in the "About Us" and put in the "Support" page instead. Rename as "Telegram Tech Support."

app/components/layout/GridFlexContainer.scss
```diff
@@ -0,0 +1,7 @@
+.component {
```
SebastienGllmt: This is all unused

app/components/layout/HorizontalFlexContainer.scss
```diff
@@ -1,5 +1,6 @@
 .component {
   width: 100%;
+  height: 100%;
```
SebastienGllmt: I don't think this is necessary?

SebastienGllmt:
Hey @ASUKA1006 thanks for the fixes. Your latest commits didn't change the color of the Yoroi icon (but it was close!) so I fixed that for you. Here is what it looks like now
image

I think this is good enough for a v1 and we can fix the other issues later after we release `1.3.0`. Thoughts?

SebastienGllmt:
image
Just pushed some small fixes for @vantuz-subhuman

PR #320, opened 2019-02-28
1.3.1 Release

SebastienGllmt:
- Adds the custom theme feature (but the menu option is disabled until we get the blog written)
- Adds German support
- Update Japanese redemption text


app/api/localStorage/index.js
```diff
@@ -66,9 +71,85 @@ export default class LocalStorageApi {
     } catch (error) {} // eslint-disable-line
   });
 
+  // ========== User Theme ========== //
+
+  getUserTheme = (): Promise<string> => new Promise((resolve, reject) => {
+    try {
+      const theme = localStorage.getItem(storageKeys.THEME);
+      if (!theme) return resolve('');
```
nicarq: @SebastienGllmt shouldn‘t this resolve to default rather than empty?

PR #322, opened 2019-03-01
Fix Desync issue

SebastienGllmt:
# Premise

- Yoroi (Browsers) doesn't implement bip44 correctly. Notably, it doesn't scan 20 addresses ahead of the last used address. This causes desyncs when you use Yoroi either on multiple computers with the same wallet or you use the same wallet on Chrome + mobile

# Solution

We use the same solution Yoroi-Mobile uses:
- Keep track of the next 20 "External" and "Internal" addresses (yes, internal too but this is what mobile seems to do) inside the DB
- Explicitly keep track of how many addresses the user has generated that way we can display only those addresses on the "receive" screen
- When you need an address, simply use the next one available in the DB (don't generate a new address through Rust)
- When a tx is received, generate new addresses to maintain the bip44 gap

# Migration system

Old wallets have two problems:
1) They are missing the 20 address gap for bip44 compliance
1) They transaction history may be incorrect (due to desync issue)

To fix this, I introduced a migration system. This is generic enough to help us perform migrations to new versions to maintain compatibility in the future also

To keep track of migrations, we now save `LAST-LAUNCH-VER` in `localstorage`. That way whenever you launch the app, you know how many migration steps need to be applied to become compatible with the latest version.

Since these steps may become long in the future, I exposed all the loading steps through the LoadingStore. This means in the future we can have a more fine-grained loading message (instead of just "loading..." we can say
```
loading Rust...
Done.
Processing migrations...
Done
```

# Misc
I removed `generateSTxs.js` as it was used to stress-test the backend while Yoroi was still being developed and it's no longer used (and I didn't want to migrate the code to use the new bip44 setup)


app/api/ada/adaLocalStorage.js
```diff
@@ -56,6 +83,19 @@ export function getLastBlockNumber(): number {
   return Number(lastBlockNum);
 }
 
+/* Last block number storage */
+
+export function saveLastReceiveAddressIndex(index: number): void {
+  const stored: LocalStorageWallet = _getFromStorage(storageKeys.WALLET_KEY);
+  stored.lastReceiveAddressIndex = index;
+  updateWallet(stored);
+}
```
vantuz-subhuman: This looks a bit weird that usage of `storageKeys.WALLET_KEY` is spread across multiple functions, but then there's `updateWallet` function that incapsulates that key. It appears as if `updateWallet` does something very different from `_getFromStorage`, even tho this is basically an inverse function, but with one argument hidden.

My advice would be to define `getWalletFromStorage` and optionally rename `updateWallet` into `updateWalletInStorage`. Or remove `updateWallet` and always use `storageKeys.WALLET_KEY` explicitly 🤔

app/api/ada/adaMigration.js
```diff
@@ -0,0 +1,95 @@
...
+  // if we had more than one address, then the WALLET key must exist in localstorage
+  saveLastReceiveAddressIndex(maxIndex);
+
+  /**
+   * Once we've saved the receive address, we dump the DB entirely
+   * We need to do this since old wallets may have incorrect history
+   * Due to desync issue caused by the incorrect bip44 implementation
+   */
+  reset();
+}
```
vantuz-subhuman: ❤️

app/api/ada/adaTransactions/adaTransactionsHistory.js
```diff
@@ -55,18 +58,75 @@ export const getAdaTxsHistoryByWallet = async (): Promise<AdaTransactions> => {
...
+  * This is because transactions in these chains are not mutually exclusive
+  * in the case of interwallet transaction
+  * Therefore we need to make sure necessary addresses are generated in both chains
+  * to ensure they are correctly detected as ours
+  */
+  await syncAddresses(account, 'Internal');
+  await syncAddresses(account, 'External');
+
+  await refreshChains();
+}
```
vantuz-subhuman: ❤️

PR #325, opened 2019-03-02
Some new assets

SebastienGllmt:
I had to change the icons a little bit from the ones Marta gave me to make the colors properly render in the application.

Additionally, I made the icons bigger (it used to be the Yoroi icon was bigger than all others, but now I made them all big) hoping the added screen spaces makes the icon easier to see.

Here is the final result
image


SebastienGllmt:
After fixing the stroke-width issue and disabling the size-adjustment, it now looks like this
image

nicarq:
the design LGTM! cc @Marta-Ryzhok

PR #332, opened 2019-03-06
daedalus-migration icon fix

sarveshwar-singh:


SebastienGllmt:
Thank you for find the root cause and a repro for fixing it 🙏 🙏 🙏 
Let's try and use #333 to solve this

PR #336, opened 2019-03-07
Feature/remove word button

ASUKA1006:
I changed the 'confirm' button to a 'remove last' button.
Before this change, when `enteredPhrase` was true, the page automatically passed to the next dialog as the 'confirm' button actually wasn't used.

screen shot 2019-03-08 at 1 18 06 am


app/components/wallet/backup-recovery/WalletRecoveryPhraseEntryDialog.js
```diff
@@ -95,10 +95,9 @@ export default class WalletRecoveryPhraseEntryDialog extends Component<Props> {
     const actions = [];
 
     actions.push({
-      className: isSubmitting ? styles.isSubmitting : null,
-      label: intl.formatMessage(messages.buttonLabelConfirm),
-      onClick: onFinishBackup,
-      disabled: !canFinishBackup,
+      label: intl.formatMessage(messages.buttonLabelRemoveLast),
```
SebastienGllmt: Probably this is what is causing the test to fail. I think you have to keep `onFinishBackup` somehow

SebastienGllmt:
I'm not super convinced about having a big "Remove last" button for the UI. It feels like it should be a tiny button somewhere but idk what would be the best UI for the least amount of work.

PR #340, opened 2019-03-07
SVG Loader on all mainnets + remove stale cleanups

SebastienGllmt:
This PR does two things:

1) In my PR #333 I only applied the fix to `development.config.js`. In this PR I add the fix to the webpack config for each network
1) Our code contained a lot of ` cleanup={['title']}`. This is because a `<title>` tag on an SVG image gives it an on-hover message. This isn't desired because the on-hover message generated by programs that export SVGs are often unhelpful (see example below). All these `cleanup` props are error-prone since we forget it in a few places. Since our new way of loading SVGs allows us to remove this at load-time, this is definitely the better solution so I use that instead and remove all the `cleanup` props

Example of useless auto-generated `<title>`
image

### DIFF
--- a/webpack/development.config.js
+++ b/webpack/development.config.js
@@ -117,7 +117,7 @@ const baseDevConfig = () => ({
       {
         test: /\.inline\.svg$/,
         issuer: /\.js$/,
-        loader: 'svg-inline-loader?removeSVGTagAttrs=false&removeTags=true&removingTags=title&removingTags=desc&idPrefix=[sha512:hash:hex:5]-',
+        loader: 'svg-inline-loader?removeSVGTagAttrs=false&removeTags=true&removingTags[]=title&removingTags[]=desc&idPrefix=[sha512:hash:hex:5]-',
       },
       {
         test: /\.(eot|otf|ttf|woff|woff2|gif)$/,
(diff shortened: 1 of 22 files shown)


webpack/development.config.js
```diff
@@ -117,7 +117,7 @@ const baseDevConfig = () => ({
       {
         test: /\.inline\.svg$/,
         issuer: /\.js$/,
-        loader: 'svg-inline-loader?removeSVGTagAttrs=false&removeTags=true&removingTags=title&removingTags=desc&idPrefix=[sha512:hash:hex:5]-',
+        loader: 'svg-inline-loader?removeSVGTagAttrs=false&removeTags=true&removingTags[]=title&removingTags[]=desc&idPrefix=[sha512:hash:hex:5]-',
```
nicarq:
```suggestion
        loader: 'svg-inline-loader?removeSVGTagAttrs=false&removeTags=true&removingTags=["title", "desc"]&idPrefix=[sha512:hash:hex:5]-',
```

nicarq: Looking at their tests, I think that should work too, but up to you. I don’t think is that important.

SebastienGllmt: Nope. I tried this originally and it didn't work but this other format does :(

PR #342, opened 2019-03-09
Daedalus Master Key restore

SebastienGllmt:
This is useful because it allows Daedalus users who forgot their mnemonic but remember their password to recover their wallet

- [x] update `js-cardano-wasm` once [this merges](https://github.com/input-output-hk/js-cardano-wasm/pull/52)
- [x] write test

fromMasterKey


SebastienGllmt:
Note: I turned autocomplete off on the masterkey form so that your browser doesn't memorize it. Another way to do it would be to turn it into a password field. Do people prefer it to be a password field or keep it as-is?

PR #356, opened 2019-03-19
Fix multiple tabs of Yoroi open & remove "tabs" permission

SebastienGllmt:
- [x] rebase after new WASM bindings merged
- [x] test with Trezor device connected

**Rationale**

We need to ensure we only have one copy of Yoroi running per localstorage instance as Yoroi is non-transactional.

**Old system**

The old way we did this is we would disable the Yoroi button if a Yoroi tab was already open (we know if it's already open by scanning all tabs the user has open). Instead, we would focus the existing copy of Yoroi open

This approach is not ideal because 
1) The user can access Yoroi directly by just accessing the Yoroi URL directly (this is especially problematic once we add URI schemes for payments)
2) The copy of Yoroi may be open on a different instance of Chrome (which shares the same localstorage) meaning you can't focus it.

**New system**

We monitor tabs inside Yoroi itself instead of the background thread which allows us to handle closing duplicate copies of Yoroi even if they're opened directly from the URL.
Additionally, we no longer iterate over all user tabs and instead use localcache to communicate across copies of Yoroi.

Unlike the old system that would block opening new tabs, the new system closes the old tab and allows you to open a new one.

**Problem**

If a new copied of Yoroi is opened at the SAME TIME as an old copy of Yoroi is closing, `loadRustModule` will livelock causing Yoroi to be stuck in the loading screen forever. We can solve this with a `sleep` call or maybe upgrading our WASM binding system will solve this issue 😓

**Gif example**

newtab

### DIFF
--- a/app/utils/tabManager.js
+++ b/app/utils/tabManager.js
@@ -1,0 +1,62 @@
+/**
+ * We may run into bugs if the user has two copies of Yoroi running on the same localstorage
+ * Since Yoroi data process is not transactional.
+ *
+ * To avoid these bugs, we only allow one tab open at a time.
+ *
+ * Since all copies of Yoroi running use the same localstorage instance,
+ * we use a localstorage listener on a specific key to detect new tabs being opened.
+ *
+ * Note: you can only listen on localstorage and not session storage
+ * Note: listener only fires for localstorage changes made by OTHER Yoroi tabs
+ * https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Responding_to_storage_changes_with_the_StorageEvent
+*/
+
+/**
+ * Name of key we use in localstorage
+ */
+const OPEN_TAB_ID_KEY = 'openTabId';
+
+/**
+ * It's possible to have two copies of Yoroi loading at the same time by going to the URL directly
+ * Therefore it's important that we add this listener BEFORE we modify the localstorage
+ * This avoids the race condition where both update localstorage but neither are listening
+ *
+ * Note: this may cause two copies of Yoroi loading at the same time close each other
+ */
+export function addCloseListener(window) {
+  window.onstorage = (e: StorageEvent) => {
+    // if another Yoroi tab open, close this tab
+    if (e.key === OPEN_TAB_ID_KEY) {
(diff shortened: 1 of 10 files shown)


vantuz-subhuman:
@SebastienGllmt, this one needs a rebase

SebastienGllmt:
I tested this in Trezor and everything still works as expected despite the following piece of code

```javascript
    // triggered from 'beforeunload' event
    // find tab by popup pattern and switch to it
    chrome.tabs.query({
        url: '*://connect.trezor.io/*/popup.html',
    }, (tabs) => {
        if (tabs.length < 0) return;
        chrome.tabs.update(tabs[0].id, { active: true });
    });
```

In the [Chrome docs](https://developer.chrome.com/extensions/tabs#property-Tab-url) it says we shouldn't be able to access the URL property:

> url | Match tabs against one or more URL patterns. Fragment identifiers are not matched. **This property is ignored if the extension does not have the "tabs" permission.**

SebastienGllmt:
Figured out what's up with Trezor. The code I linked above is actually not run through our main app but the Trezor Connect popup which can [send a message](https://github.com/trezor/connect/blob/7276f91dd7f190a0aab06094f58315ae47be30a5/src/js/webusb/extensionPermissions.js#L36) to the `trezor-usb-permissions.html` file where it has permission to access its own URL

sarveshwar-singh:
Reference about Trezor permission:
https://github.com/trezor/connect/blob/749490dd9bc0edf2d24d958fc1fd56472025e8ba/src/js/webextension/README.md

app/utils/tabManager.js
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * WARNING: You should only call this AFTER the wasm bindings have loaded
+ * closing a different copy of Yoroi while loading WASM causes the load to hang
+ * This bug only exists in Chrome (verified still occurs in Chrome v75)
+ *
+ * WARNING: You should call this function BEFORE making any other changes to localstorage
+*/
+export function closeOtherInstances() {
+  localStorage.setItem(OPEN_TAB_ID_KEY, Date.now().toString());
+}
```
vantuz-subhuman: ❤️

PR #358, opened 2019-03-19
Limit max height of dropdown selection

SebastienGllmt:
# Problem

Users with small screens were unable to select their language because the dropdown menu was getting cutoff

image

# Solution

To fix this, I made that dropdown selections now have a scrollbar if the size of all items goes over 300px (300 was chosen because we have dropdowns approximately this size before and nobody ever complained until we got over 300)

Here is how it looks like

## Over 300px

image

## Under 300px

Note: scrollbar doesn't even appear when under 300px (as expected)
image

### DIFF
--- a/app/themes/overrides/OptionsOverrides.scss
+++ b/app/themes/overrides/OptionsOverrides.scss
@@ -1,1 +1,11 @@
 // Css module overrides go here …
+
+.ul {
+  overflow-x: hidden;
+  overflow-y: overlay;
+  max-height: 300px;
+  &::-webkit-scrollbar-button {
+    height: 7px;
+    display: block;
+  }
+}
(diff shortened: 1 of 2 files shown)


DominikGuzei:
@SebastienGllmt thanks for the PR, is there any chance to fix this issue properly (= taking the screen into account and ensure that the dropdown never gets cut-off?)

SebastienGllmt:
@DominikGuzei I don't think this is doable in any reasonable way purely in CSS inside the override. If you want to do that kind of logic, you'd probably have to do it inside the javascript of React-Polymorph. I'm happy with a hardcoded 300px for Yoroi but if you want to spend more time on this to have a proper fix for Daedalus, you're welcome to do so.

DominikGuzei:
@SebastienGllmt yeah that's what i meant 😉 but no worries, we can take that task over

PR #360, opened 2019-03-20
Remove height restriction

SebastienGllmt:
Translations would often be too long to fit inside a button. This is hard to detect for our translation team and hard to fix (in cases where the sentence can't be easily reduces in size)

I can't think of any reason not to remove the height restriction on buttons. It may not look great in some cases, but it's probably better than missing text.

**Note:** Not specifying a height will instead default to `auto`.

## Example

image


nicarq:
great idea

PR #362, opened 2019-03-21
Webpack 4 + Rust Upgrade

SebastienGllmt:
We need to update webpack 4 for a few reasons

1) It's just better than version 3
2) We need it to compress the app in the right way for releasing on Firefox
3) We need it to use the new Rust bindings

It turns out that we can't upgrade webpack without upgrading the Rust bindings to [cardano-wallet](https://www.npmjs.com/package/cardano-wallet) at the same time.

Tasks:
- [x] upgrade webpack
- [x] upgrade to new Rust bindings
- [x] add back in error checking on Rust calls
- [x] try to speed up Daedalus migration
- [x] chunk files small enough for Firefox ( see #386)
- [x] Rebase to use Firefox 67 on CI to fix WASM MIME loading (can't launch Firefox support publicly until > May 14th)
- [x] fix Trezor transaction sending
- [ ] decide what to do with removed `jest` tests


SebastienGllmt:
Here are the improvements thanks to the Webpack v4 migration + more efficient WASM bindings

image

SebastienGllmt:
I realized these new bindings also mean we don't need `cardano-crypto.js` anymore and getting rid of it brings us down even further.

image

app/api/ada/adaAccount.js
```diff
@@ -2,38 +2,43 @@
...
 export function createCryptoAccount(
-  masterKey: string,
+  encryptedMasterKey: string,
   walletPassword: string,
   accountIndex: number = ACCOUNT_INDEX
 ): CryptoAccount {
-  const cryptoWallet = getCryptoWalletFromMasterKey(masterKey, walletPassword);
-  const result: CryptoAccount = getResultOrFail(Wallet.newAccount(cryptoWallet, accountIndex));
-  return Object.assign({}, { account: accountIndex }, result);
+  const masterKey = getCryptoWalletFromMasterKey(encryptedMasterKey, walletPassword);
```
SebastienGllmt: renane

app/api/ada/daedalusTransfer.js
```diff
@@ -52,36 +60,71 @@ export function getAddressesWithFunds(payload: {
...
-    const output = await getReceiverAddress();
+    const outputAddr = await getReceiverAddress();
+    const feeAlgorithm = RustModule.Wallet.LinearFeeAlgorithm.default();
+
+    // firts build a transaction to see what the cost would be
+    const fakeTxBuilder = new RustModule.Wallet.TransactionBuilder();
+    addTxInputs(fakeTxBuilder, inputs);
+    const inputAmount = new BigNumber(
+      fakeTxBuilder.get_input_total().to_str()
+    ).times(LOVELACES_PER_ADA);
```
SebastienGllmt: use ada() and lovelace()

app/api/ada/adaAddress.js
```diff
@@ -85,8 +85,8 @@ export function getAdaAddressesByType(addressType: AddressType): Promise<AdaAddr
  */
 export async function popBip44Address(type: AddressType): Promise<AdaAddress> {
   return type === 'Internal'
-    ? popBip44InternalAddress()
-    : popBip44ExternalAddress();
+    ? await popBip44InternalAddress()
+    : await popBip44ExternalAddress();
```
vantuz-subhuman: Should it be `await` actually? Cuz function itself returns `Promise`

app/api/ada/adaLocalStorage.js
```diff
@@ -23,11 +37,29 @@ const storageKeys = {
...
+  const pubKey = RustModule.Wallet.PublicKey.from_hex(localAccount.root_cached_key);
+  const account = RustModule.Wallet.Bip44AccountPublic.new(
+    pubKey,
+    RustModule.Wallet.DerivationScheme.v2()
+  );
+  return {
+    account: localAccount.account,
+    root_cached_key: account,
+    derivation_scheme: localAccount.derivation_scheme
+  };
```
vantuz-subhuman: Is `account` type here `string` or `Bip44AccountPublic`?

app/api/ada/adaTransactions/adaTransactionsHistory.js
```diff
@@ -42,6 +43,9 @@ import {
   getLastBlockNumber,
   getSingleCryptoAccount
 } from '../adaLocalStorage';
+import type {
+  CryptoAccount,
+} from '../adaLocalStorage';
```
vantuz-subhuman: Why two new imports are added and they are the only changes in the file?

app/api/ada/adaAddress.js
SebastienGllmt: In this case it actually doesn't matter if you write async or not

app/api/ada/adaLocalStorage.js
SebastienGllmt: `Bip44AccountPublic`. You can tell both from the return type and the way the `account` variable is generated.

app/api/ada/adaTransactions/adaTransactionsHistory.js
SebastienGllmt: Because `CryptoAccount` used to be a globally declared type from the old bindings. Since the old bindings don't exist anymore (but we still need this type because this is what we serialize for `localstorage` I had to add it back in and import it like you would any other type.

Old type definition: https://github.com/Emurgo/yoroi-frontend/blob/develop/flow/declarations/CardanoCrypto.js#L187

app/api/ada/lib/adaAddressProcessing.js
```diff
@@ -158,3 +165,21 @@ function _addFetchedAddressesInfo(
...
+  });
+  return indices.map(i => {
+    const pubKey = cryptoAccount.address_key(
+      type === 'Internal',
+      RustModule.Wallet.AddressKeyIndex.new(i)
+    );
+    const addr = pubKey.bootstrap_era_address(setting);
+    return addr.to_base58();
+  });
+}
```
vantuz-subhuman: Why `'Internal'` is hardcoded in here? Maybe it should be reflected in the function name

app/api/ada/lib/cardanoCrypto/cryptoWallet.js
```diff
@@ -73,10 +79,23 @@ const mnemonicToSeedHex = (mnemonic: string, password: ?string) => {
...
-  return encryptWithPassword(password, masterKey);
+export function generateWalletMasterKey(mnemonic : string, password : string): string {
+  const entropy = RustModule.Wallet.Entropy.from_english_mnemonics(mnemonic);
+
+  /**
+ * there is no wallet entropy password in yoroi
+ * the PASSWORD here is the password to add more _randomness_
+ * when deriving the wallet root key from the entropy
+ * it is NOT the spending PASSWORD
+ */
```
vantuz-subhuman: ❤️

app/api/ada/lib/utils.js
```diff
@@ -298,3 +170,21 @@ export function getLatestUsedIndex(addresses: Array<AdaAddress>): number {
       .map(address => address.index)
   );
 }
+
+export type UtxoLookupMap = { [string]: { [number]: UTXO }};
+export function utxosToLookupMap(
+  utxos: Array<UTXO>
+): UtxoLookupMap {
+  // first create 1-level map of tx_hash -> UTXO
```
vantuz-subhuman:
```suggestion
  // first create 1-level map of tx_hash -> [UTXO]
```

app/api/ada/lib/utils.js
```diff
@@ -298,3 +170,21 @@ export function getLatestUsedIndex(addresses: Array<AdaAddress>): number {
...
+  const txHashMap = _.groupBy(utxos, utxo => utxo.tx_hash);
+
+  // now create 2-level map of tx_hash -> index -> UTXO
```
vantuz-subhuman:
```suggestion
  // now create 2-level map of (tx_hash -> index -> UTXO)
```

app/api/ada/daedalusTransfer.js
```diff
@@ -92,29 +131,8 @@ export async function generateTransferTx(payload: {
...
-      value: utxo.amount,
-      addressing: addressingByAddress[utxo.receiver]
-    },
-    address: utxo.receiver
-  }));
+export function coinToBigNumber(coin: RustModule.Wallet.Coin): BigNumber {
+  const ada = new BigNumber(coin.ada());
+  const lovelace = ada.times(LOVELACES_PER_ADA).add(coin.lovelace());
+  return lovelace;
 }
```
vantuz-subhuman: ❤️

app/api/ada/daedalusTransfer.js
```diff
@@ -52,36 +60,67 @@ export function getAddressesWithFunds(payload: {
...
+    const realTxBuilder = new RustModule.Wallet.TransactionBuilder();
+    addTxInputs(realTxBuilder, inputs);
+    const sendAmount = inputAmount.minus(fee.toString());
+    addOutput(realTxBuilder, outputAddr, sendAmount.toString());
+
+    // sanity check
+    const balance = realTxBuilder.get_balance(feeAlgorithm);
+    if (!balance.is_zero()) {
+      throw new GenerateTransferTxError();
+    }
```
vantuz-subhuman: Could it be possible that reducing output size might reduce tx byte-size and therefore reduce fee and make balance positive? 🤔 I think in Rust code I have definitely seen a case when they give up few lovelaces in a case like this and just say it's ok to sacrifice them as fee.

vantuz-subhuman: Yes, I think it's possible because of CBOR. I'm looking for Rust example.

vantuz-subhuman: At the moment they only use sanity check: `if real_fee > selection_result.estimated_fees` - then it's error. But real fee might turn out to be `<` than estimated fee, then balance will be non-zero positive value.

https://github.com/input-output-hk/rust-cardano/blob/53d5e5fb8981523c124d2d1ebc2845bc6bdacfa8/cardano/src/wallet/scheme.rs#L156-L160

vantuz-subhuman: We even opened a related issue in there - https://github.com/input-output-hk/rust-cardano/issues/349

app/api/ada/lib/adaAddressProcessing.js
SebastienGllmt: Maybe you didn't notice but this is a equality check (`===`)
The first argument to `address_key` is whether or not the address you create is a internal. I check to see if `type === 'Internal'` to convert from the enum to a bool.

vantuz-subhuman: Ah yes, my mistake. Thank you!

PR #363, opened 2019-03-21
Add Ledger address verification

SebastienGllmt:
- [x] expose `showAddress` in bridge (see [here](https://github.com/Emurgo/yoroi-extension-ledger-bridge/blob/master/src/index.js#L127))
- [x] propogate derivation path in code
- [x] add icon in receive screen for address verification
- [x] create dialog for address verification
- [x] only show dialog when using Ledger
- [x] error message when device not connected
- [x] loader while loading and block button

Note: we should add this to Trezor too once we have the time

Here is what the UI looks like

## Receive screen

image

## Address details dialog (non-hardware wallet)

image

## Address details dialog (hardware wallet)

image

Screen Shot 2019-03-23 at 1 32 53 AM

Screen Shot 2019-03-23 at 1 33 16 AM


app/stores/ada/HWVerifyAddressStore.js
```diff
@@ -0,0 +1,177 @@
...
+      Logger.info('HWVerifyStore::ledgerVerifyAddress finalized ');
+    }
+  }
+
+  /**
+   * TODO: This needs to be moved to a single file (also from LedgerSendStore)
+   */
+
+  /** Converts error(from API or Ledger API) to LocalizableError */
+  _convertToLocalizableError = (error: any): LocalizableError => {
```
SebastienGllmt: Let's avoid having to copy paste this functio

nicarq: yeah, there is a TODO 3 lines above this so it’s wrapped up before PR review

SebastienGllmt:
Found a bug. If open the Address Details button, close it, then open it again the verify address button breaks until you refresh the page

PR #365, opened 2019-03-23
Add verify function on Trezor

SebastienGllmt:
#363 added the ability to verify addresses using your Ledger device. This PR adds the equivalent functionality for Trezor.

You can see the flow for the verification [here](https://wiki.trezor.io/User_manual:Receiving_payments#Confirm_on_Trezor) and the API for this functionality [here](https://github.com/trezor/connect/blob/develop/docs/methods/cardanoGetAddress.md)

At the same time I upgraded the version of TrezorConnect to fix a Flow issue in the `Cardano Input` type (see changelogs [here](https://github.com/trezor/connect/blob/develop/CHANGELOG.md#702))

# Problem

Although during testing the address and the derivation path match, the QR codes do not.

image

 I scanned both QR codes and both contain the same data: `Ae2tdPwUPEZ3hNNwfupiDwymWB54yoSHhm1qqY8LCVFpze4dhCRScbHLoxC`

It turns out this is due to a few differences. Trezor & Ledger use the same library for QR code generation and you can read about its features [here](https://github.com/nayuki/QR-Code-generator#features)

## Error correction level

Here is a table of the levels
image

Trezor uses [level M](https://github.com/trezor/trezor-core/blob/4f28093d2e3c1658f742b5eff678b9aa2662ca46/embed/extmod/modtrezorui/display.c#L643) but we use [level L](https://github.com/zpao/qrcode.react#available-props)

## QR Version

Trezor uses version range [1,9]. Our QR code library doesn't support this option

## Masking

Trezor uses "auto" masking. Our QR code library doesn't support this option.

### DIFF
--- a/app/stores/ada/HWVerifyAddressStore.js
+++ b/app/stores/ada/HWVerifyAddressStore.js
@@ -10,6 +10,7 @@ import {
 import {
   LedgerBridge,
 } from 'yoroi-extension-ledger-bridge';
+import TrezorConnect from 'trezor-connect';
 
 import LocalizableError from '../../i18n/LocalizableError';
 import environment from '../../environment';
@@ -24,8 +25,11 @@ import {
 } from '../../utils/logging';
 
 import {
-  convertToLocalizableError
+  convertToLocalizableError as ledgerErrorToLocalized
 } from '../../domain/LedgerLocalizedError';
+import {
+  convertToLocalizableError as trezorErrorToLocalized
+} from '../../domain/TrezorLocalizedError';
 
 import Wallet from '../../domain/Wallet';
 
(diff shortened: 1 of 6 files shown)


app/stores/ada/HWVerifyAddressStore.js
```diff
@@ -59,33 +66,32 @@ export default class AddressesStore extends Store {
...
-      Logger.error('AddressStore::_verifyAddress::error: ' + stringifyError(error));
-      this._setError(convertToLocalizableError(error));
-    } finally {
-      this._setActionProcessing(false);
+    if (params.wallet.isLedgerNanoSWallet) {
+      await this.ledgerVerifyAddress(path);
+    } else if (params.wallet.isTrezorTWallet) {
+      await this.trezorVerifyAddress(path, address);
+    } else {
+      throw new Error('AddressStore::_verifyAddress called with unrecognized hardware wallet');
```
nicarq: @SebastienGllmt I just realized after merging this PR that this throw is not catch anywhere and it shouldn’t because it’s an action connected directly to a listener. It should probably be changed to a Logger

PR #376, opened 2019-04-01
Feature/add brave tests to ci pipeline

rcmorano:


rcmorano:
I think that last change should do it :)

PR #380, opened 2019-04-03
added ledger help link for KO, RU, ZH-HANT, ZH-HANS, DE, FR, ID

sarveshwar-singh:
- [X] CrowdIn is also been updated.


nicarq:
I’m confused. I checked the links and I still see the content in english e.g. https://emurgo.io/#/en/blog/how-to-use-ledger-nano-s-with-yoroi-cardano-de (it should be in german). If we don’t have the content ready yet, we should probably wait.

PR #381, opened 2019-04-03
deleted Indonesian duplicate locale

sarveshwar-singh:


nicarq:
@SebastienGllmt @sarveshwar-singh I just updated the `Crowdin.yml` to correctly name the Indonesian language.

nicarq:
Also the Indonesian file is deleted in the same translation PR. So maybe we don’t need this PR after all.

PR #390, opened 2019-04-08
Remove fee information for INCOME transactions

### LINKED ISSUE #383: Incorrect fee on incoming transactions
It seems that for all incoming transactions the fee of the transaction is shown as 0.000000
Yoroi

Transactions always have a non-zero fee, so this is either a bug or a misleading way to show that someone else paid the transaction fee.

If we want to have a bank-like UI, where the recipient is not presented with the information about fees paid by the other party, then we should remove the Fee line (like we do in mobile).

On the other hand, receiving transactions is always free, and the information about fees is informative, so showing the value could be helpful.

**Background:**

Ledger has no problems with showing fees:

Ledger

Daedalus, on the other hand, doesn't show fees even on sent transactions:

Daedalus

AdaLite displays all fees:

AdaLite

In my opinion, we should consider this a bug and show the actual fee.

ebeloded:
Closes #383

### DIFF
--- a/app/components/wallet/transactions/Transaction.js
+++ b/app/components/wallet/transactions/Transaction.js
@@ -236,10 +236,14 @@ export default class Transaction extends Component<Props, State> {
               </div>
             )}
             <div>
-              <h2>
-                {intl.formatMessage(messages.fee)}
-              </h2>
-              <span>{formattedWalletAmount(data.fee.abs(), false)}</span>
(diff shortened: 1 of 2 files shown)


SebastienGllmt:
The tests are failing because of poorly designed tests where it takes all the information about a transaction displayed on a page, turns it into an array and then accesses a specific index.

PR #394, opened 2019-04-09
Paper wallets

vantuz-subhuman:
UI is not finished


app/api/ada/lib/cardanoCrypto/cryptoWallet.js
```diff
@@ -37,7 +45,8 @@ export const isValidEnglishAdaPaperMnemonic = (
   phrase: string,
   numberOfWords: ?number = 27
 ) => {
-  const [unscrambled, unscrambledLen] = unscramblePaperAdaMnemonic(phrase, numberOfWords);
+  const fakePassword = numberOfWords === 21 ? 'xxx' : undefined;
```
SebastienGllmt: Any password will return some valid unscrambled mnemonic so we just pass a fake password to pass downstream validation

app/api/ada/lib/cardanoCrypto/cryptoWallet.js
```diff
@@ -47,24 +56,62 @@ export const isValidEnglishAdaPaperMnemonic = (
...
   return [undefined, 0];
 };
 
+/** Scramble provided mnemonic with the provided password */
+export const scramblePaperAdaMnemonic = (
+  phrase: string,
+  password: string,
+): string => {
+  const iv = new Uint8Array(8);
+  window.crypto.getRandomValues(iv);
```
SebastienGllmt: `window`?

app/api/ada/lib/cardanoCrypto/cryptoWallet.js
```diff
@@ -22,6 +22,14 @@ const protocolMagic = CONFIG.network.protocolMagic;
 /** Generate a random mnemonic based on 160-bits of entropy (15 words) */
 export const generateAdaMnemonic = () => bip39.generateMnemonic(160).split(' ');
 
+/** Generate a random mnemonic based on 96-bits of entropy (9 words), and also return hexed seed */
+export const generatePasswordAndMnemonic = () => {
+  const mnemonic = bip39.generateMnemonic(96);
```
SebastienGllmt: remove

app/api/ada/index.js
```diff
@@ -213,8 +217,36 @@ export type UpdateWalletPasswordResponse = boolean;
...
+  async createAdaPaperPdf({ paper, isMainnet, logback }: { paper: AdaPaper, isMainnet?: boolean, logback?: Function }): Promise<Blob> {
+    const { addresses, scrambledWords, pass } = paper;
+    return generateAdaPaperPdf({
+      words: scrambledWords,
+      addresses,
+      isMainnet: isMainnet || true,
+      isCustomPass: !!pass.password
+    }, s => {
+      Logger.info('[PaperWalletRender] ' + s);
+      return !logback || logback(s);
```
SebastienGllmt: remove `logback`?

app/containers/wallet/dialogs/WalletRestoreDialogContainer.js
```diff
@@ -5,13 +5,23 @@ import validWords from 'bip39/wordlists/english.json';
...
+  mode: "regular" | "paper" | "paper-password"
+};
 
 @observer
 export default class WalletRestoreDialogContainer extends Component<Props> {
 
-  onSubmit = (values: { recoveryPhrase: string, walletName: string, walletPassword: string }) => {
+  onSubmit = (values: { recoveryPhrase: string, walletName: string, walletPassword: string, paperPassword: string }) => {
+    if (isPaperMode(this.props.mode)) {
+      console.log('>', values.paperPassword);
```
SebastienGllmt: remove

SebastienGllmt:
Here are some notes I took during our call when we went over the code

- [ ] fix drop-down cutoff
- [ ] make cancel button cancel execution
- [ ] make the "restore paper" button on the create wallet screen one button instead of two
- [ ] consider making a "verification" step? For example, make sure address printed matches address on screen. May not be necessary
- [ ] see if we can remove the "logback". These don't match the way we use mobx
- [ ] remove 30 word option and enforce password for wallet creation
- [ ] compress or vectorize images
- [ ] delete xcf files (they are gimp projects)
- [ ] implements step 3 & 4 of paper wallet store

app/api/ada/lib/cardanoCrypto/cryptoWallet.js
vantuz-subhuman: 😬

vantuz-subhuman:
> remove 30 word option and enforce password for wallet creation

Note there are two images called:
1. `paper-wallet-certificate-page-2.png`
2. `paper-wallet-certificate-page-2_pass.png`

First one is for non-password 30-word mode, so if we are removing it - then the image might also be removed.

vantuz-subhuman:
> make the "restore paper" button on the create wallet screen one button instead of two

This one is the part of the "remove 30 word option and enforce password for wallet creation"

app/api/ada/lib/cardanoCrypto/cryptoWallet.js
vantuz-subhuman: Removed

app/api/ada/lib/cardanoCrypto/cryptoWallet.js
vantuz-subhuman: done

app/containers/wallet/dialogs/WalletRestoreDialogContainer.js
vantuz-subhuman: done

vantuz-subhuman:
So number of images is reduced to only 3 for: background, page1, and page2. Network label (e.g. "TESTNET") is printed dynamically on top of regular images. Managed to compress background image from 3.3 Mb to ~0.85 Mb by using indexed colors mode. Still looks great in the whole color-scheme, imo. Other two images are 0.12 Mb and 0.1 Mb.

Generated PDF document size is ~2.3 Mb

vantuz-subhuman:
> make cancel button cancel execution

This one is not critical anymore. Current simple cancelling implementation only caused problem when separate steps of PDF generation took significant time. Now that we figured out compression the whole process is really fast, so cancelling does not cause any serious lags.

SebastienGllmt:
- [ ] add label in restore dialog warning users about wrong password behavior

app/api/ada/paperWallet/paperWalletPdf.js
```diff
@@ -90,11 +92,17 @@ export const generateAdaPaperPdf = async (
...
+  network: string,
+  y: number,
+  r?: number,
+  xShift?: number
+) {
   doc.setFontSize(60);
   doc.setFontType('bold');
   doc.setTextColor(255, 180, 164);
-  textCenter(doc, y, 'TESTNET', null, r, (r || 0) > 90, xShift);
+  textCenter(doc, y, network.toUpperCase(), null, r, (r || 0) > 90, xShift);
```
vantuz-subhuman: ❤️

app/components/settings/categories/PaperWalletSettings.js
```diff
@@ -82,6 +82,7 @@ export default class PaperWalletSettings extends Component<Props> {
           options={numAddressOptions}
           {...numAddresses.bind()}
           skin={SelectSkin}
+          isOpeningUpward // need this to make sure all options still show on small screens
```
vantuz-subhuman: ❤️

vantuz-subhuman:
I would really like to fix this text wrapping ☹️

vantuz-subhuman:
@SebastienGllmt, this button looks really good, but mouse pointer stays looking like a standard cursor hovering it, and it should be that finger pointer, so it's clear it's clickable. Can we do it?

image

image

SebastienGllmt:
I pushed changes that include Ruslan's feedback. Here is what the flow looks like now:

# Create

image

image

image

image

image

image

image

image

# Restore

image

image

vantuz-subhuman:
All is good! Only one thing: discussed with Marta - we need to remove this image. Serves no purpose and only steals vertical space in already narrow enough dialog window.

image

PR #396, opened 2019-04-10
cleanup webpacks

SebastienGllmt:
Had to learn the webpack configs to do #386 so I decided to do some cleanup at the same time too. Basically we had a bunch of webpack files which were all copy/pastes of each other with one string changes so I simplified it to just two: `prod` and `dev`

Note: this is unrelated to the `dev:mainnet` issue since to fix that we need to fixup the way we handle manifest files.

**Note:** This PR is on a long chain of dependencies and not on `develop` so be careful about merging it. I made this a separate PR from #386 to make the diff easier to read.


scripts/build.js
```diff
@@ -11,4 +11,4 @@ tasks.copyAssets('build', argv.env);
 console.log('[Webpack Build]');
 console.log('-'.repeat(80));
 
-exec(`./node_modules/.bin/webpack --config webpack/${argv.env}.config.js --progress --profile --colors`);
+exec(`./node_modules/.bin/webpack --config webpack/prodConfig.js --progress --profile --colors --env=${argv.env}`);
```
vantuz-subhuman: So I don't get it, why we always use `prodConfig`? Even when env might be `dev`

SebastienGllmt: the `build.js` is purely for production builds. 

The `dev` environment has absolutely nothing to do with `development` builds. Having a network type `dev` was all a hack used in Icarus to get it to load up the right manifest file that allows connection to localhost.

PR #403, opened 2019-04-12
Fix some of the single-account assumptions in Yoroi

SebastienGllmt:
Across Yoroi we assume in a few places we are only dealing with a single account. In most places, this isn't necessary because actually store the current account inside localstorage.

We need to do this for the following:
1) Multi-account in Yoroi
2) Migration to Redux

Note that this doesn't remove this assumption in all places. Notably, it excludes:
1) Changes to the DB format and DB queries to handle multi-accounts. Fixing this part requires a migration for all existing users (not a big deal) so I left it out of this PR
2) Stores still assume single account. This is fine because we will throw away all stores in the Redux migration anyway.


SebastienGllmt:
This code has to be rebased after the test PR is merged to pass

PR #408, opened 2019-04-15
Ugrade react and react-dom

### LINKED ISSUE #392: Upgrade React
While working on #383, I tried using _React.Fragment_ to conditionally elements without adding another element to the hierarchy. This resulted in build or runtime errors, depending on which syntax I tried to use. Flow didn't provide any assistance in pointing out the problem (TypeScript would be more helpful here).

It turned out that the version of React used in the project is 16.0.0 (from September 2017).

There was a lot of awesome features introduced since then. Just to name a few:

- Fragments,
- New lifecycle methods,
- Performance Profiler 
- New Context API (!)
- createRef API
- Code splitting with lazy/Suspense (!) for dynamic imports
- React.memo for functional components
- StrictMode (!)
- Hooks (!!!)

Modern React development doesn't happen without these features. I think we need to upgrade React and dependencies that rely on it and resolve whatever problems pop up.

Any reason not to do it?

ebeloded:
This will bring us the following new features:

- Fragments,
- New lifecycle methods,
- New Context API 
- createRef API
- Code splitting with lazy/Suspense for dynamic imports
- React.memo for functional components
- StrictMode 
- Hooks 

Closes #392


SebastienGllmt:
Yeah we need this for the plugin system we want to eventually add and maybe this will help with Redux migration

Issue #379, opened 2019-04-03
Large memory footprint while idle


ebeloded:
Yoroi extension consumes a lot of system resources even without the app being opened. Here are two Chrome Task Manager screenshots of freshely launched browser:

![image](https://user-images.githubusercontent.com/5859177/55450956-2f495c80-55fb-11e9-88ac-ae9d14516486.png)

![image](https://user-images.githubusercontent.com/5859177/55451169-04abd380-55fc-11e9-9043-50aac2f802c0.png)

On the windows machine the app wasn't even opened yet - the extension was installed automatically for my google account and immediately went to the top of the memory footprint chart.

We need to investigate this and make sure that the memory footprint is tamed. The extension should consume a nominal amount of resources while the app is inactive.

SebastienGllmt:
Interesting. I assume this is because `background.pug` loads `vendor.js` which is ~18MiB (see [here](https://github.com/Emurgo/yoroi-frontend/blob/develop/chrome/views/background.pug#L5)).

Shrinking the size of this file in any reasonable way requires upgrading to `webpack` v4 which is a task I'm actively working on. 

However, `background` definitely doesn't need all of `vendor.js`. It only depends on `bluebird` and a single function from `lodash`. We'll see how much that improves the situation.

ebeloded:
Glad you're working on Webpack 4. Switching made a big difference in the project I worked on before.

Yeah, I think that we need to implement code splitting and only include the essentials on the background page and load the rest dynamically. Even the use of bluebird and lodash, given that we work in the modern Chrome environment, sounds like an overkill. 

But more importantly, why is vendor.js 18MB? What's in there? 

Combined with 3.6MB of yoroi.bundle.js we load more than 21MB on the page, which is quite a lot.

I checked AdaLite - the app seems to have the same functionality with less than 400KB of js. Are we doing something on the client they do on the server? What makes such a difference?

SebastienGllmt:
1) Webpack v3 blocks us from optimizing some of these files
2) We have dependencies such as PDF importing and PDF generation which take a non-trivial amount of space

We'll see how much space we take after webpack v4 migration is done.

PR #411, opened 2019-04-16
Remove WebpackDllReference plugin

ebeloded:
This PR removes the use of DllPlugin, which gives us some immediate pefromance wins, and opens doors to futher optimizations.

⚠`dll` folder is no longer used, nor ignored. Feel free to remove it `rm -rf dll`

**Why**
Until now, we have used DllPlugin to bundle all dependencies listed in package.json into a single file, and include this 'vendor' bundle on all pages. As a result of this, we were importing a big bundle, which contained some dead code. Additionally, it didn't allow webpack to perfrom [tree shaking](https://webpack.js.org/guides/tree-shaking/), and made it impossible to meaningfully use [dynamic imports](https://webpack.js.org/guides/code-splitting/#dynamic-imports).

**Immediate benefit**
`background.html` doesn't load 2.2MB of vendor.js, and instead only requires 190KB. This further addresses the problem of Yoroi using a lot of memory without being opened (#379).

**Further benefits:**

**1. [Tree Shaking](https://webpack.js.org/guides/tree-shaking/)**
Because we don't include the dependencies in full, we can [write our code](https://www.blazemeter.com/blog/the-correct-way-to-import-lodash-libraries-a-benchmark) in a way that makes use of this feature:

❌`import _ from 'lodash';`
✅`import { debounce } from 'lodash';`

The latter will only import one function (and its dependencies), instead of importing the whole module.

**2. [Dynamic Imports](https://webpack.js.org/guides/code-splitting/#dynamic-imports)**

While tree shaking is great, more benefit we will get from dynamic imports.

Some heavy dependencies we have are very rarely used, yet we currently include all of them in our bundles. For example, we should only load `pdfParser` when the user goes through the process of ada redemption. In some cases this may require minor refactoring, in many cases the process should be trivial change of import syntax:

❌`import pdfjsLib from 'pdfjs-dist';`
✅`const pdfjsLibPromise = import('pdfjs-dist')`

What's more, because we updated React, we now have access to [React.lazy and Suspense](https://reactjs.org/docs/code-splitting.html). This means that dynamically loading components is extremely easy:

❌`import OtherComponent from './OtherComponent';`
✅`const OtherComponent = React.lazy(() => import('./OtherComponent'));`

This PR didn't change the combined size of our bundles (which is ~5.1 MB), but now we have a way of chipping away the weight. Tiny changes, remarkable results.


SebastienGllmt:
Good stuff. The DllPlugin was confusing people, blocking optimizations and blocking other tools from working properly. Time to finally give it the 🪓

PR #412, opened 2019-04-17
Fix tests broken by wasm migration

SebastienGllmt:
We could solve this either by moving to `karma-webpack` and keeping the `cardano-wallet-browser` or we can keep Jest and depend both on the `nodejs` and the `browser` module and swap it dynamically when we do testing.

This PR does the swap approach.


SebastienGllmt:
FYI the tests fail because of tricky behavior. The `dynamic-import-node` needs to only be enabled for the `jest` tests which is why it's conditionally included in the `.babelrc`. Since `jest` automatically sets the ENV to `test` that's what we use but that collides with `test-prepare` which also uses `test.

Issue #339, opened 2019-03-07
Fix About Yoroi icon text


SebastienGllmt:
Yoroi has some icons in its footer when you are on the wallet creation screen.

![image](https://user-images.githubusercontent.com/2608559/53979321-d4822980-4150-11e9-8bf9-48823814aad2.png)

When the screen size is too small, we remove all the text and show just the icons. The definition of "too small" was hardcoded for this specific page.

We reused the same FooterItem class for the About Yoroi icon. This means it inherits the same hard-coded size in `FooterItem.scss`. We should rename this to something other than `FooterItem` (as it's no longer being used just as a footer) and make this hard-coded size only apply to the wallet creation page and a different size apply for the About Yoroi page

SebastienGllmt:
Closing since this has been fixed

PR #415, opened 2019-04-21
[ISSUE 339] Refactor 'About Yoroi' items

ScaredIbis:
For issue https://github.com/Emurgo/yoroi-frontend/issues/339

This change introduces a shared component (LinkButton), used directly in AboutYoroiSettings. The FooterItem component has also been refactored and uses LinkButton as a child. 

This has allowed styles used for hiding text on small screens to be separated from the AboutYoroiSettings component.

BEFORE:
image

AFTER:
image


SebastienGllmt:
Hey, thanks for the PR!
Three things to mention:

1) We just introduced a new theme so make sure this works on both themes (you can switch themes from the settings page). I fixed the merge conflict for you.
1) I noticed your PR adds a scrollbar on the LinkButton (as seen in your screenshot). That needs to be fixed.
1) The reason the media query was added was so that either all buttons show text or none show text. Looks like your PR breaks this (in both the wallet add screen and the About Yoroi screen)

ScaredIbis:
I have addressed the issues you brought up:

1. I made some minor style changes so that it functions the same and looks decent in both themes.
2. I have updated the styles for HorizontalFlexContainer (shared component) to remove the scrollbar that was showing up. From what I can see, the style update doesn't break other components that also use HorizontalFlexContainer.
3.  The media queries are now applying to both Wallet Creation and About Yoroi pages properly. These styles are defined in a way that make it easy to change the breakpoints for each page individually.

SebastienGllmt:
Looks great! Thanks!

PR #418, opened 2019-04-24
Send All options

SebastienGllmt:
We need our API to support a "sendAll" feature. This is required for two cases:

1) Sending your entire balance through the "Send" tab. This wasn't possible before because the old WASM bindings enforced a change address but the new WASM bindings allow us to do this.
2) Add a "transfer from v2 wallet" function

These two help alleviate one of the biggest UX problems for Yoroi: people who want to transfer their existing Yoroi accounts to a hardware wallet need can't migrate all of their wallet funds. 

- [x] make changes in the API layer to support sendAll
- [x] Rewrite Daedalus transfer 
- [x] Add some unit tests
- [ ] ~~add a "sendAll" button in the "send" screen~~ (cc @Marta-Ryzhok )
- [ ] ~~add a "transfer from v2" option option beside "transfer from Daedalus"~~ (cc @Marta-Ryzhok )

I tried to write this feature with some foresight to allow the same API to be used for a transaction builder screen. In case you're curious, the rust API doesn't support a "fees included" feature.

**Note:** For the tests, I moved all the state&network queries outside of the API layer. This is because we have to eventually make the API a standalone package so it needs to have these passed in.

**Note:** We have some other tasks we I need to work on so I suggest we merge this PR as-is and add the UI components later.

### DIFF
--- a/app/api/ada/adaTransactions/adaNewTransactions.js
+++ b/app/api/ada/adaTransactions/adaNewTransactions.js
@@ -2,41 +2,23 @@
 
 // Handles interfacing w/ lovefieldDB and rust-cardano to create transaction
 
-import _ from 'lodash';
-import {
-  getUTXOsForAddresses,
-  sendTx
-} from '../lib/yoroi-backend-api';
-import {
-  Logger,
-  stringifyError
-} from '../../../utils/logging';
+import BigNumber from 'bignumber.js';
 import {
   addressesToAddressMap,
-  getAdaAddressesList,
-  popBip44Address
 } from '../adaAddress';
-import {
-  getCryptoWalletFromMasterKey,
-} from '../lib/cardanoCrypto/cryptoWallet';
 import type {
   AdaAddress,
   AdaFeeEstimateResponse,
   UTXO,
+  UnsignedTxFromUtxoResponse,
   UnsignedTxResponse,
 } from '../adaTypes';
 import {
   NotEnoughMoneyToSendError,
-  SendTransactionError,
-  GetAllUTXOsForAddressesError,
-  InvalidWitnessError
 } from '../errors';
-import { getWalletMasterKey, getCurrentAccountIndex } from '../adaLocalStorage';
 import type { ConfigType } from '../../../../config/config-types';
-import { HARD_DERIVATION_START } from '../../../config/numbersConfig';
 import type { AdaAddressMap } from '../adaAddress';
-import { utxosToLookupMap } from '../lib/utils';
-import type { SignedResponse } from '../lib/yoroi-backend-api';
+import { utxosToLookupMap, coinToBigNumber } from '../lib/utils';
 
 import { RustModule } from '../lib/cardanoCrypto/rustLoader';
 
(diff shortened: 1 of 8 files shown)


app/api/ada/adaTransactions/adaNewTransactions.js
```diff
@@ -46,107 +28,149 @@ const protocolMagic = CONFIG.network.protocolMagic;
...
+  };
+}
+
+export async function newAdaUnsignedTxFromUtxo(
+  receiver: string,
+  amount: string,
+  changeAdaAddr: ?AdaAddress,
+  allUtxos: Array<UTXO>,
+): Promise<UnsignedTxFromUtxoResponse> {
+  const feeAlgorithm = RustModule.Wallet.LinearFeeAlgorithm.default();
```
nicarq: it’s a really small detail but I would move this right before it’s used. We may even not actually use it and we still invoke it.

nicarq: nvm

app/api/ada/adaTransactions/adaNewTransactions.js
```diff
@@ -46,107 +28,149 @@ const protocolMagic = CONFIG.network.protocolMagic;
...
-  const fee = txBuilder.estimate_fee(RustModule.Wallet.LinearFeeAlgorithm.default());
-  return { fee: fee.to_str() };
+  const { txBuilder } = await newAdaUnsignedTx(
+    receiver,
+    amount,
+    changeAdaAddr,
+    possibleInputAddresses,
+    addressesToUtxos
+  );
+  const fee = txBuilder.get_balance_without_fees().value();
```
nicarq: this reads pretty odd. Why is using `get_balance_without_fees()` rather than `estimate_fee`?

app/api/ada/adaTransactions/adaNewTransactions.js
```diff
@@ -46,107 +28,149 @@ const protocolMagic = CONFIG.network.protocolMagic;
 /** Calculate the transaction fee without actually sending the transaction */
 export async function getAdaTransactionFee(
   receiver: string,
-  amount: string
+  amount: string,
+  changeAdaAddr: ?AdaAddress,
+  possibleInputAddresses: Array<AdaAddress>,
```
nicarq: the possible naming is pretty good, maybe we should add a comment to explain why it’s like that

app/api/ada/adaTransactions/adaNewTransactions.js
```diff
@@ -46,107 +28,149 @@ const protocolMagic = CONFIG.network.protocolMagic;
...
+): Promise<UnsignedTxFromUtxoResponse> {
+  const totalBalance = allUtxos
+    .map(utxo => new BigNumber(utxo.amount))
+    .reduce(
+      (acc, amount) => acc.plus(amount),
+      new BigNumber(0)
+    );
+
   const feeAlgorithm = RustModule.Wallet.LinearFeeAlgorithm.default();
+  // firts build a transaction to see what the cost would be
```
nicarq: just an idea, maybe this could be grouped like this:

```
let fee;
{ 
  // firsts build a transaction to see what the cost would be
  const fakeTxBuilder = new RustModule.Wallet.TransactionBuilder();
  const inputs = utxoToTxInput(allUtxos);
  addTxInputs(fakeTxBuilder, inputs);
  const inputAmount = coinToBigNumber(fakeTxBuilder.get_input_total());
  addOutput(fakeTxBuilder, receiver, inputAmount.toString());
  fee = coinToBigNumber(fakeTxBuilder.estimate_fee(feeAlgorithm));
}
```

app/api/ada/adaTransactions/adaNewTransactions.js
```diff
@@ -46,107 +28,149 @@ const protocolMagic = CONFIG.network.protocolMagic;
...
-  } catch (err) {
+  // create a new transaction subtracing the fee from your total UTXO
+  const newAmount = totalBalance.minus(fee);
+  const unsignedTxResponse = await newAdaUnsignedTxFromUtxo(receiver, newAmount, null, allUtxos);
+
+  // sanity check
+  const balance = unsignedTxResponse.txBuilder.get_balance(feeAlgorithm);
+  if (balance.is_negative()) {
+    /**
+     * The balance may be slightly positive. This is because lowering the "amount" to send
```
nicarq: small typo 
`The balance may be slightly positive` ->  `The balance may be slightly negative`

app/api/ada/adaTransactions/adaNewTransactions.js
nicarq: you already added it :D

app/api/ada/adaTransactions/adaNewTransactions.js
SebastienGllmt: I added more comments to explain what this means

app/api/ada/adaTransactions/adaNewTransactions.js
SebastienGllmt: I added more comments to explain what this means

app/api/ada/adaTransactions/adaNewTransactions.js
SebastienGllmt: I added more comments to explain what this means

app/api/ada/adaTransactions/adaNewTransactions.js
```diff
@@ -216,17 +257,12 @@ export async function addTxIO(
   }
 }
 
-export function signTransaction(
+function addWitnesses(
   txFinalizer: RustModule.Wallet.TransactionFinalized,
   senderUtxos: Array<UTXO>,
   addressesMap: AdaAddressMap,
-  cryptoWallet: RustModule.Wallet.Bip44RootPrivateKey,
+  accountPrivateKey: RustModule.Wallet.Bip44AccountPrivate,
```
nicarq: 🙌

app/api/ada/adaTransactions/adaNewTransactions.test.js
```diff
@@ -0,0 +1,204 @@
...
+    const unsignedTxResponse = await newAdaUnsignedTxFromUtxo(
+      'Ae2tdPwUPEZKX8N2TjzBXLy5qrecnQUniTd2yxE8mWyrh2djNpUkbAtXtP4',
+      '5000', // smaller than input
+      null,
+      utxos
+    );
+    expect(unsignedTxResponse.senderUtxos).toEqual(utxos);
+    expect(unsignedTxResponse.txBuilder.get_input_total().to_str()).toEqual('1.000000');
+    expect(unsignedTxResponse.txBuilder.get_output_total().to_str()).toEqual('0.005000');
+    expect(unsignedTxResponse.txBuilder.estimate_fee(
```
nicarq: I’m confused with his one. Should we be testing `estimate_fee` if we are not using it and we are actually using `get_balance_without_fees`?

SebastienGllmt: I didn't see a reason not to add it. Didn't give it much thought.
It could change (without affecting `get_balance_without_fees`) if we change the selection policy

PR #420, opened 2019-04-26
Fix Trezor account creation

SebastienGllmt:
The refactoring to get rid of some of the single-account assumptions accidentally broke Trezor support so I fixed it.

At the same time, I move a few more stateful calls out of the API layer. We need to continue to do this to eventually be able to make the API a standalone package.


nicarq:
@SebastienGllmt good call on removing those calls to `adaLocalStorage`

PR #421, opened 2019-04-26
Tentatively add Spanish

SebastienGllmt:
Just like Indonesian, it's not visible and won't be until translation is finished

- [x] Add Spanish to code
- [ ] Add Indonesian ToS 
- [ ] Add Spanish ToS
- [ ] Add Spanish flag


nicarq:
this PR should also update `https://github.com/Emurgo/yoroi-frontend/blob/develop/crowdin.yml` to add the pairing with the new spanish format.

PR #434, opened 2019-05-04
New SCSS architecture

SebastienGllmt:
# CSS architecture before this PR

We currently support two themes, each with different CSS.
We pass a boolean flag across the codebase to distinguish the two themes.
The CSS for each theme is entirely copy pasted (no code-reuse) with relevant portions modified.

# Problem

The above architecture means
1) Our code only supports 2 themes. Adding a third theme would require modifying the whole frontend code.
2) The code is really fragile and hard to read. If you need to make a change, you need to remember to change it in both places or things will go wrong.
3) We can't test the new theme because all the CSS selectors are now different.

# Proposed fix

### For conditional rendering
We need to know which theme is active to decide whether or not to render a specific component at all (ex: the modern theme has additional pictures that aren't rendered on the classic theme). However, instead of passing around a boolean, we should instead pass around the whole theme name 

### For conditional styling
Instead of passing around a boolean flag to decide how to render the css, we instead define the theme on the top-level div of our `App`. In individual `scss` files, we check the global parent div to detect which style to use and apply it on-top of a common style.

Now
1) CSS common between themes isn't duplicated.
2) Adding a new custom theme is easier because you will start off with the common settings and can expand from that
3) Tests can use CSS selectors from the base theme so the tests can run on any theme with little or no modification

# How to avoid large PRs

We can't implement the above fix in a single PR because it would be too big for anybody to reasonably review. Instead, this PR sets up the architecture so we can split this improvement into many small PRs.

Here is how I did this:
1) `containers` now all receive the theme name from the `ProfileStore` instead of a boolean
1) `containers` convert this into a `boolean` and passes it down to components

This mean whenever you want to implement the new improved architecture on a component, you can change the `boolean` to a string on the specific component (or delete if no longer needed) and make a PR just for that component. 

# Proof of concept of for conditional styling

I implemented the architecture on `SupportSettings.js` to show how it works. Here are the relevant parts:

1) We apply the theme to the top-level div in `App.js`
```
<div className={currentTheme} style={{ height: '100%' }}>
  // yoroi app here
</div>
```

2) We modify the common functionality in the CSS
```
.component {
  // common to all themes
}

:global(.YoroiClassic) .component {
  // only for classic theme
}
:global(.YoroiModern) .component {
  // only for modern theme
}
```


app/containers/wallet/WalletReceivePage.js
```diff
@@ -127,7 +128,7 @@ export default class WalletReceivePage extends Component<Props, State> {
             isHardware={wallet.isHardwareWallet}
             verify={() => actions.ada.hwVerifyAddress.verifyAddress.trigger({ wallet })}
             cancel={() => actions.ada.hwVerifyAddress.closeAddressDetailDialog.trigger()}
-            classicTheme={theme.classic}
+            classicTheme={profile.currentTheme === THEMES.YOROI_CLASSIC}
```
nicarq: What do you think of adding a method isClassic or isModern inside profile? it works fine, but it seems to verbose

app/stores/toplevel/ProfileStore.js
```diff
@@ -125,7 +126,7 @@ export default class ProfileStore extends Store {
 
   // ========== Current/Custom Theme ========== //
 
-  @computed get currentTheme(): string {
+  @computed get currentTheme(): Themes {
```
nicarq: Maybe this one should be Theme in singular

app/containers/wallet/WalletReceivePage.js
SebastienGllmt: sure why not

app/stores/toplevel/ProfileStore.js
SebastienGllmt: sure why not

nicarq:
cc @YankoFelipe @sarveshwar-singh @vantuz-subhuman

PR #435, opened 2019-05-06
Add types to Stores

SebastienGllmt:
This is the workflow for Yoroi

UI -> Actions -> Stores -> API

People may not remember, but actually we have no types for any of the 3 arrows above connecting the 4 components (this is historically because Daedalus has no types at all and I did some work to add types but I didn't cover the whole codebase)

Since we will do a lot of refactoring of the API structure, I want types to avoid accidentally breaking everything. Unfortunately, since Flow is really bad at generics among other things, this whole thing is kind of a mess.

### DIFF
--- a/app/api/ada/adaWallet.js
+++ b/app/api/ada/adaWallet.js
@@ -156,7 +156,7 @@ export function createAdaHardwareWallet({
 }
 
 /** Wrapper function to create new mnemonic according to bip39 */
-export const generateAdaAccountRecoveryPhrase = () => (
+export const generateAdaAccountRecoveryPhrase: void => Array<string> = () => (
   generateAdaMnemonic()
 );
 
--- a/app/api/ada/adaWallet.js
+++ b/app/api/ada/adaWallet.js
@@ -30,10 +30,6 @@ import type {
   AdaWalletMetaParams,
   AdaHardwareWalletParams,
 } from './adaTypes';
-import type {
-  ChangeAdaWalletSpendingPasswordParams,
-  AdaWalletRecoveryPhraseResponse,
-} from './index';
 import {
   getUTXOsSumsForAddresses
 } from './lib/yoroi-backend-api';
@@ -160,7 +156,7 @@ export function createAdaHardwareWallet({
 }
 
 /** Wrapper function to create new mnemonic according to bip39 */
-export const generateAdaAccountRecoveryPhrase = (): AdaWalletRecoveryPhraseResponse => (
+export const generateAdaAccountRecoveryPhrase = () => (
   generateAdaMnemonic()
 );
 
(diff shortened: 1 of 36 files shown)


app/api/ada/adaWallet.js
```diff
@@ -160,7 +156,7 @@ export function createAdaHardwareWallet({
 }
 
 /** Wrapper function to create new mnemonic according to bip39 */
-export const generateAdaAccountRecoveryPhrase = (): AdaWalletRecoveryPhraseResponse => (
+export const generateAdaAccountRecoveryPhrase = () => (
```
nicarq:
```suggestion
export const generateAdaAccountRecoveryPhrase:  Array<string> = () => (
```

app/api/ada/index.js
```diff
@@ -134,16 +113,130 @@ import {
...
+
+// refreshTransactions
+
+export type GetTransactionsRequestOptions = {
+  skip: number,
+  limit: number,
+};
+export type GetTransactionsRequest = {
+  walletId: string,
+  ...$Shape<GetTransactionsRequestOptions>
```
nicarq: this is cool. I didn’t know `$Shape`

app/api/localStorage/index.js
```diff
@@ -58,7 +60,7 @@ export default class LocalStorageApi {
 
   setTermsOfUseAcceptance = (): Promise<void> => new Promise((resolve, reject) => {
     try {
-      localStorage.setItem(storageKeys.TERMS_OF_USE_ACCEPTANCE, true);
+      localStorage.setItem(storageKeys.TERMS_OF_USE_ACCEPTANCE, JSON.stringify(true));
```
nicarq: 😅

SebastienGllmt: yeah I don't know what to do about this one. Theoretically you can only save strings to localstorage (which is why the rest of this file uses `JSON` everywhere) but also saving non-strings seems to work perfectly fine

SebastienGllmt: image

It looks like Chrome just knows what to do with it. Even through `JSON.stringify(true)` returns "true" and not `true`, the result in `localstorage` is still just the literal `true` (like before) and not "true"

nicarq: yeah, I double checked and it works fine and we shouldn’t have any issues with compatibility. What scared me is that in the old code, it was just sent as a boolean and left to the API to decide to convert it. I think the change that you did is the *proper* way to do it.

app/stores/toplevel/ProfileStore.js
```diff
@@ -191,7 +222,7 @@ export default class ProfileStore extends Store {
 
   getThemeVars = ({ theme }: { theme: string }) => {
     if (theme) return require(`../../themes/prebuilt/${theme}.js`);
-    return require(`../../themes/prebuilt/${THEMES.YOROI_CLASSIC}.js`); // default
+    return require(`../../themes/prebuilt/${ProfileStore.getDefaultTheme()}.js`); // default
```
nicarq: I couldn’t find this method 🤔

nicarq:
awesome work, Seba!! 👏

PR #440, opened 2019-05-07
Added script that builds and zips Yoroi into an archive

vantuz-subhuman:


vantuz-subhuman:
Discussing the fact that we already have `compress` script, but it's kinda weird and hard to figure out

PR #442, opened 2019-05-08
Porting the settings folders to the new scss architecture

YankoFelipe:
Porting the settings folders to the new scss architecture.
Note: I haven't found the way to select the style with something like `style.elementName.$currentTheme` yet
[ch940]


nicarq:
@YankoFelipe could u fix the merge conflicts and also check why the tests are failing 🙏

app/components/profile/terms-of-use/TermsOfUseForm.js
```diff
@@ -27,7 +29,7 @@ type Props = {
   onSubmit: Function,
   isSubmitting: boolean,
   error?: ?LocalizableError,
-  classicTheme: boolean
+  currentTheme: Theme
```
nicarq: I think you could also pass the function `isClassic`

sarveshwar-singh: @nicarq 
yes, we are thinking to change it in one go at all the places.
As change will affect more than 50 files so thinking to make change in different PR.
cc @YankoFelipe

nicarq:
@YankoFelipe could u take a look at the merge conflicts 🙏

app/components/profile/terms-of-use/TermsOfUseForm.js
nicarq: sounds good! let’s not continue adding more of them in other PRs though 😄

YankoFelipe:
@nicarq Let me add the conditional styling and then I'll fix the merge conflicts

app/components/profile/terms-of-use/TermsOfUseForm.js
sarveshwar-singh: @nicarq 
Yes, we are following the approach to remove `classicTheme: boolean` but in some exceptional case we have to keep it, will think in deep if we can remove it completely.
cc @YankoFelipe

PR #445, opened 2019-05-08
Decouple state fetching from api layer

SebastienGllmt:
# About this PR

Now that the stores have types, I can safely move all networking calls out of the API.

This is useful if we ever want to support multiple network layers (such as allowing users to connect to their own machine instead of trusting our server) but more importantly we need to do this to get a stateless API layer so that we can move all this off to a separate package.

To review: **look at the state-fetcher folder before reading the rest of the code**. That will make it more clear why the rest of the code needs to change.

Note: I also add missing types to the `actions` folder in this PR to avoid things breaking.

# Going forward

This is definitely not the "final" configuration of how the API will look. I'm just trying to get towards a stateless API layer first then we can decide how the architecture should look for the final design. Trying to pick the architecture with a bunch of hidden stateful calls hidden within the API layer would probably lead to a lot of "oh, I didn't realize that X, Y and Z"


app/api/ada/hardwareWallet/newTransaction.js
```diff
@@ -89,8 +90,11 @@ export async function createTrezorSignTxPayload(
 
 /** List of Body hashes for a list of utxos by batching backend requests */
 export async function txsBodiesForInputs(
-  inputs: Array<TxoPointerType>
+  inputs: Array<TxoPointerType>,
+  getTxsBodiesForUTXOs: TxBodiesFunc,
 ): Promise<{[key: string]:string}> {
+  // TODO: move this to state-fetcher/helpers
```
SebastienGllmt: TODO

PR #446, opened 2019-05-08
Better messaging for invalid mnemonic phrases

leivaburto:
Adding a new message to show how many words are left for the mnemonic to be complete. Once it has 15 words it would say if it is valid or not.


nicarq:
@leivaburto looks good! but you need to update (as in fix) and probably add a new test :)

SebastienGllmt:
image
FYI this is what it looks like

PR #447, opened 2019-05-09
Sebastiengllmt/ch949/decouple storage from api layer

SebastienGllmt:
After decoupling the network component, next we need to decouple the storage layer.

This way both mobile and the extension can use the same API but with different storage layers.

**Note:** For this PR I didn't change much of the code that handles the actual storage. This is because as we know, there is a huge amount of stuff that needs to be changed (in fact we basically have to rewrite it) so instead of spending time improving it I just split it off and then we can think about the redesign of the storage layer in a separate spec.

**Note:** This PR isn't based on develop. We need to merge the decoupling of the network layer before this PR.


app/api/ada/restoreAdaWallet.js
```diff
@@ -7,69 +7,32 @@ import {
...
-import { createCryptoAccount } from './adaAccount';
 import type { ConfigType } from '../../../config/config-types';
 import type { FilterFunc } from './lib/state-fetch/types';
+import type {
+  SaveAsAdaAddressesFunc,
+} from './lib/storage/types';
+
+import { RustModule } from './lib/cardanoCrypto/rustLoader';
+
+const accountIndex = 0; // TODO: remove
```
SebastienGllmt: TODO

app/api/ada/lib/state-fetch/batchedFetcher.js
```diff
@@ -0,0 +1,268 @@
...
+  getTransactionsHistoryForAddresses = (body: HistoryRequest): Promise<HistoryResponse> => (
+    batchGetTransactionsHistoryForAddresses(
+      this.baseFetcher.getTransactionsHistoryForAddresses
+    )(body)
+  )
+
+  sendTx = (body: SignedRequest): Promise<SignedResponse> => (
+    // We don't batch transaction sending (it's just a single requeset)
+    // TODO: Should we support batching a list of transactions?
+    this.baseFetcher.sendTx(body)
```
vantuz-subhuman: There once was an idea that we can eventually create a "tx constructor" in Yoroi, that would allow to define multiple targets and amounts first and then enter spending password once and create/send multiple transactions. There still will be a good discussion on whether we need a functionality like this, over an option to specify multiple outputs for a single transaction, but that, imo, is the only good potential use for batched tx sending, atm

app/api/ada/lib/state-fetch/batchedFetcher.js
```diff
@@ -0,0 +1,268 @@
...
+
+declare var CONFIG: ConfigType;
+const addressesLimit = CONFIG.app.addressRequestSize;
+const transactionsLimit: number = config.wallets.TRANSACTION_REQUEST_SIZE;
+
+/**
+ * Makes calls to Yoroi backend service
+ * https://github.com/Emurgo/yoroi-backend-service/
+ */
+export class BatchedFetcher implements IFetcher {
```
vantuz-subhuman: ❤️

PR #449, opened 2019-05-10
Account Number Plates

vantuz-subhuman:
Implementation for: https://github.com/Emurgo/yoroi-frontend/blob/develop/docs/specs/code/NUMBER_PLATES.md

Implemented:
1. New component `WalletAccountIcon` with blockies and color-scheme logic incapsulated
2. When `Wallet` is returned from API - current account is now also included
3. TopBar UI is changed to display account icon and account plate ID at all times, separately from balance now
4. When creating a paper wallet - account icon and plate ID is printed on the certificate too (can be optionally disabled on the initial paper-wallet settings page, where number of addresses is selected)
5. On the final dialog screen of paper-wallet creation - account icon and plate ID are displayed as "Your Paper Wallet Account identification", so users see it even if they chose not to print it on paper itself.
6. When a regular or a paper wallet is restored, after standard dialog is filled with information and user clicks "Restore" - additional "Verify Restored Wallet" dialog is shown, which displays account icon and plate ID and wallet addresses (1 for a regular wallet and 5 for a paper wallet) and also displays warnings about restoring a wrong empty wallet. When this verification step is cancelled - user returns to regular restoring dialog with all the same information already in place.

The "Verify Restored Wallet" dialog can be used for two additional usecases:
1. Non-destructive paper validation, when user enters all the information from a paper-wallet and checks account identification and addresses, and then just cancels. Paper wallet does not lose it's "coldness" at this point.
2. Plausible deniability with paper-wallets, when user can enter mnemonics from a paper wallet, but intentionally different password. In this case Yoroi will display them account identification and addresses for an alternative wallet, which also does not become hot. At this point user can save this displayed alternative information somewhere and use it as actual wallet, while information printed on paper with original password is used as decoy.


vantuz-subhuman:
1. [ ] Language files are not updated yet with new text messages
2. [ ] Wallet restoration UI test fails because of the new verification step
3. [ ] No UI tests for account plates themselves

app/containers/TopBarContainer.js
```diff
@@ -22,6 +22,9 @@ export default class TopBarContainer extends Component<Props> {
       wallet={stores.substores[environment.API].wallets.active}
       currentRoute={app.currentRoute}
       formattedWalletAmount={formattedWalletAmount}
+      theme={{
```
SebastienGllmt: given the word `theme` is already used for a different concept, can we rename this?

app/components/wallet/WalletRestoreVerifyDialog.scss
```diff
@@ -0,0 +1,56 @@
+@import '../../themes/mixins/loading-spinner';
+@import '../../themes/mixins/error-message';
+
+.dialog {
+  font-family: var(--font-light);
+  .walletPassword {
```
SebastienGllmt: I assume `.walletPassword` and `.error` shouldn't be here and are just copy pasted?

app/containers/wallet/dialogs/WalletRestoreDialogContainer.js
```diff
@@ -51,6 +98,21 @@ export default class WalletRestoreDialogContainer extends Component<Props> {
...
     }
 
+    const { verifyRestore, submitValues } = this.state;
+    if (verifyRestore) {
+      const { addresses, accountPlate } = verifyRestore;
+      return (
+        <WalletRestoreVerifyDialog
+          addresses={addresses}
+          accountPlate={accountPlate}
+          onCopyAddress={() => {}}
```
SebastienGllmt: fake function should be fixed with a proper copy to clipboard

app/api/ada/adaLocalStorage.js
```diff
@@ -45,8 +45,11 @@ export function saveCryptoAccount(
   _saveInStorage(storageKeys.ACCOUNT_KEY, localAccount);
 }
 
-export function getCurrentCryptoAccount(): CryptoAccount {
+export function getCurrentCryptoAccount(): ?CryptoAccount {
```
SebastienGllmt: Although I agree that conceptually we don't always have an account, in practice in Yoroi we always do after account creation so I don't really get why this was necessary. It makes the code a messier, causes merge conflicts with the storage decouple PR and we'll have to redo all of this with the new API anyway

app/components/topbar/WalletTopbarTitle.js
```diff
@@ -22,17 +37,31 @@ export default class WalletTopbarTitle extends Component<Props> {
...
     } = this.props;
 
     // If we are looking at a wallet, show its name and balance
     const walletRoutesMatch = matchRoute(`${ROUTES.WALLETS.ROOT}/:id(*page)`, currentRoute);
     const showWalletInfo = walletRoutesMatch && wallet;
+
+    const isHardwareWallet = (wallet && wallet.type) === WalletTypeOption.HARDWARE_WALLET;
+    const iconDivClass = isHardwareWallet ? styles.divIconHardware : styles.divIcon;
+    const [accountPlateId, iconComponent] = (wallet && wallet.accounts) ?
+      constructPlate(wallet.accounts[0], theme.identiconSaturationFactor, iconDivClass)
```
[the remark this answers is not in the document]
vantuz-subhuman: 🤔 Good point

app/containers/wallet/dialogs/WalletRestoreDialogContainer.js
vantuz-subhuman: The copy functionality works as it is. This function is only used as a callback on successful copying, and used only to display a notification. I have taken an example from other part of the code for paper-wallets where there's also this comment:

image

I will put a comment for now here too, to avoid confusion. And maybe we want to create an issue to come up with a standard clipboard animation or something, and use it on all such cases.

vantuz-subhuman: Or better yet, I will make this parameter optional for cases like this. Underlying component actually checks whether this callback is specified or not:

image

vantuz-subhuman:
Wallet verification step in both themes:



image

app/components/topbar/WalletTopbarTitle.js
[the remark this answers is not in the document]
vantuz-subhuman: Fixed. Changed logic to have `currentAccount` encapsulated in `WalletStore`

app/containers/TopBarContainer.js
vantuz-subhuman: Renamed into `themeProperties`. The logic was that once we migrate to proper "themes" - they should be passed as objects with this identicon saturation level optionally defined.

app/components/wallet/WalletRestoreVerifyDialog.scss
vantuz-subhuman: Yep. My bad. Removed

app/api/ada/adaLocalStorage.js
vantuz-subhuman: Discussed with @SebastienGllmt - there's logical case when `getCurrentCryptoAccount` might be called when account is not initialised yet, and it was happening with the new code when I added new call to this function. I am ok to hacking around without changing these types for now, if it's gonna change anyways

PR #451, opened 2019-05-10
Info message when user selects a tier2 language

leivaburto:
Adding an informative message when user selects a tier2 language (community-based translation).


SebastienGllmt:
image

Maybe we should add a background for the one on the wallet creation page? It's kind of hard to read depending on your screen size

SebastienGllmt:
For thanking the contributors, can you do the following:

Create a new i18n message that says
> "Thank you to the following for their contribution: "
Then create a second i18n message that is simply blank

Now when a new contributor joins our translation system, they can add their own name without having to change the code.

In our code, check if the 2nd message is blank.
If not, display the 1st + 2nd message

app/components/settings/categories/GeneralSettings.js
```diff
@@ -10,13 +10,29 @@ import LocalizableError from '../../../i18n/LocalizableError';
...
 import FlagLabel from '../../widgets/FlagLabel';
+import { tier1Languages } from '../../../config/languagesConfig';
 
 const messages = defineMessages({
   languageSelectLabel: {
     id: 'settings.general.languageSelect.label',
     defaultMessage: '!!!Language',
   },
-
+  languageSelectLabelInfo: {
```
SebastienGllmt: Maybe better to make these global messages instead of duplicating them? (see `global-messages.js`)

You could even add a helper method in there that constructs the `contributors` string so you don't have to duplicate this logic either

app/config/languagesConfig.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+export const tier1Languages = ['en-US', 'ja-JP', 'ru-RU', 'es-ES'];
+
+// In the future we should add name of non-tier1 translation contributors in here
```
SebastienGllmt: remove

app/i18n/locales/zh-Hant.json
```diff
@@ -122,6 +122,10 @@
   "settings.general.aboutYoroi.website": "Yoroi網站",
   "settings.general.aboutYoroi.youtube": "EMURGO YouTube",
   "settings.general.languageSelect.label": "語言",
+  "settings.general.languageSelect.labelInfo": "The selected language translation is fully provided by the community",
```
SebastienGllmt: FYI you should only ever modify the English file in a PR. Otherwise it may cause merge conflicts with the translation system we use.

app/stores/toplevel/ProfileStore.js
```diff
@@ -302,7 +302,7 @@ export default class ProfileStore extends Store {
   };
 
   _redirectToRoot = () => {
-    this.actions.router.goToRoute.trigger({ route: ROUTES.WALLETS.ROOT });
+    this.actions.router.goToRoute.trigger({ route: ROUTES.WALLETS.ADD });
```
SebastienGllmt: Why was this changed? Doesn't match the function name anymore.

SebastienGllmt:
Is it possible to add a back arrow on the Terms of Use page like we have in dialogs? I'm don't mind approving this PR without it but it may be useful

image

app/stores/toplevel/ProfileStore.js
leivaburto: ProfileStore reactions were triggering some automatic redirections and I thought this would solve it, but it didn't. I changed this back to the original and I added a new action for updating locale and redirect to terms of use.

app/i18n/locales/zh-Hant.json
leivaburto: Good to know! Thanks

leivaburto:
I added a new action and changed a bit the conditions on reactions. I also had to update current tests to include the new behavior (which requires clicking on "Continue" button at the Language Selection Form). All checks are passing now.

sarveshwar-singh:
@leivaburto 

- `LanguageSelectionForm.js` + `LanguageSelectionForm.scss`
image
image

- `GeneralSettings.js` + `GeneralSettings.scss`
image
image

Seems to be similar, how about making a separate common component?

app/i18n/global-messages.js
```diff
@@ -249,3 +270,12 @@ export const environmentSpecificMessages = {
     },
   }),
 };
+
+export function listOfTranslators(contributorsList, contributorsAck): string {
```
sarveshwar-singh: @leivaburto 

- if you make a separate component them this whole function `listOfTranslators` can be moved to that component, don't feel this function should be here.

- `if (contributorsList !== globalMessages.translationContributors.defaultMessage)`
comparing to `defaultMessage` will get you in trouble if we decide to use language specific names (that still need to decided), but we never use `defaultMessage` for logical things.
I feel, the whole if condition is not needed here, `if (contributorsList !== globalMessages.translationContributors.defaultMessage)`

sarveshwar-singh:
@leivaburto 
image

Plz try to fix those, let me know if you need some help.

app/i18n/global-messages.js
leivaburto: Regarding your second comment, the idea is this: there would be an empty string, and contributors who are willing to do so will put their names in. Empty string messages fallback to defaultMessage value, so contributorsList would take that value. What do you mean by using language specific names? AFAIK strings would fallback to default values defined in global messages, regardless of language, which in this case is `!!!List of contributors`. Or is there another strategy to be considered?

leivaburto: Regarding your first comment, it makes sense. Actually, I was working on PR #494 and I had to make something similar, so I'm planning to use that same component in here (it's called InformativeMessage). The component is generic, so I'm trying to figure out where would be better to put the method you mention. Anyway, I want to make sure the component I added makes sense (I think it does). Could you take a look at it? and if everything is OK I could use it here. Btw, thanks for your feedback.

sarveshwar-singh: @leivaburto 
> What do you mean by using language specific names?
Got clear after your explanation.

PR #459, opened 2019-05-13
Fix checkbox & select dropdown

SebastienGllmt:
This PR contains three simple fixes:

# No margin under select and caret placement

The `Select` react-polymorph component on the classic theme didn't have any margin below making any element (in this case a checkbox) stick on it which doesn't look good. (see picture below). Additionally, the caret is not placed in the center so I adjusted that also.

### Before

image

### After

image

# Options select checkbox was being cutoff

### Before

image

### After

image


sarveshwar-singh:
@SebastienGllmt 

>This PR contains three simple fixes:
I see two block of changes,
three means **Options select checkbox was being cutoff** contains two changes (Language selection page + Setting page) ?

SebastienGllmt:
> @SebastienGllmt
> 
> > This PR contains three simple fixes:
> > I see two block of changes,
> > three means **Options select checkbox was being cutoff** contains two changes (Language selection page + Setting page) ?

yes

PR #470, opened 2019-05-14
Fixing overlapping pre-filled values with floating labels

leivaburto:
CSS hack for avoiding Chrome pre-filled inputs overlapping with floating input fields.

Before:
image

After: 
image

### DIFF
--- a/app/themes/skins/FormFieldOwnSkin.scss
+++ b/app/themes/skins/FormFieldOwnSkin.scss
@@ -38,7 +38,7 @@
   input:-webkit-autofill:hover,
   input:-webkit-autofill:focus,
   input:-webkit-autofill:active {
-      -webkit-transition-delay: 99999s;
+    -webkit-transition-delay: 99999s;
   }
 
   &:hover {
--- a/app/themes/skins/FormFieldOwnSkin.scss
+++ b/app/themes/skins/FormFieldOwnSkin.scss
@@ -31,7 +31,6 @@
     border-radius: 8px;
 
     margin-top: -17px;
-    background-color: transparent;
   }
 
   input:-webkit-autofill,
--- a/app/themes/skins/FormFieldOwnSkin.scss
+++ b/app/themes/skins/FormFieldOwnSkin.scss
@@ -31,6 +31,14 @@
     border-radius: 8px;
 
     margin-top: -17px;
+    background-color: transparent;
+  }
+
+  input:-webkit-autofill,
+  input:-webkit-autofill:hover,
+  input:-webkit-autofill:focus,
+  input:-webkit-autofill:active {
+      -webkit-transition-delay: 99999s;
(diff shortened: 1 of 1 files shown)


app/themes/skins/FormFieldOwnSkin.scss
```diff
@@ -31,6 +31,14 @@
     border-radius: 8px;
 
     margin-top: -17px;
+    background-color: transparent;
```
sarveshwar-singh: is this something forgotten to delete?

app/themes/skins/FormFieldOwnSkin.scss
```diff
@@ -31,6 +31,14 @@
...
+  }
+
+  input:-webkit-autofill,
+  input:-webkit-autofill:hover,
+  input:-webkit-autofill:focus,
+  input:-webkit-autofill:active {
+      -webkit-transition-delay: 99999s;
```
sarveshwar-singh: 2 spaces are extra

app/themes/skins/FormFieldOwnSkin.scss
leivaburto: No, it's part of the trick, otherwise the background keeps overlapping.

app/themes/skins/FormFieldOwnSkin.scss
leivaburto: Commited removing extra spaces ;)

app/themes/skins/FormFieldOwnSkin.scss
sarveshwar-singh: @leivaburto 
I deleted `background-color: transparent;` and it still works on all cases i can think of.
Can you explain in which case do you need this?

logically `-webkit-transition-delay: 99999s;` delays the `input:-webkit-autofill` transition for `99999s` so i am not sure why `background-color: transparent;` is needed at all.

app/themes/skins/FormFieldOwnSkin.scss
```diff
@@ -31,6 +31,14 @@
     border-radius: 8px;
 
     margin-top: -17px;
+    background-color: transparent;
+  }
+
```
sarveshwar-singh: I think it's better of add some comment about the trick.
basically what problem we are fixing by this trick, so it will help others understand in future.

app/themes/skins/FormFieldOwnSkin.scss
leivaburto: You are right! Must have been something that helped me in a partial solution, but now is no longer needed as you said.

app/themes/skins/FormFieldOwnSkin.scss
leivaburto: Agree, commited.

PR #480, opened 2019-05-15
Maintain button for showing password when the field is on error

leivaburto:
Eye button for showing passwords disappears if the input is incorrect so it doesn't let you see what's wrong. Fixing that.

Screen Shot 2019-05-14 at 8 14 18 PM


Marta-Ryzhok:
@leivaburto yes great! Thanks ))

PR #485, opened 2019-05-16
V almonacid/ch1013/cardano uri scheme spec

v-almonacid:
Adds and updates specs on a URI scheme for Cardano and its integration in Yoroi.


ebeloded:
Really nice research on this @v-almonacid!

The most powerful use of URL Schemes I've seen was done by Cultured Code for Things. Check it out here: https://support.culturedcode.com/customer/en/portal/articles/2803573. It would be absolutely awesome if we were to add this kind of power to Yoroi.

v-almonacid:
Thanks! but most of the work was done on a previous research by Seba. 

That's indeed a super powerful scheme, probably something we could build on top on the protocol.

PR #487, opened 2019-05-17
Cleanup background script

### LINKED ISSUE #379: Large memory footprint while idle

ebeloded:
It seems that we had some leftover code, which was bloating the background script and increasing the memory usage while the app was idle.
Closes #379 

There was also a debounce set on app initialization. With this PR the app will start loading immediately (you could definitely tell the half second difference).

If double-clicking is a real concern, we can inline a throttle function (debounce was the wrong function here and we don't need to import 70KB of lodash for this).


chrome/extension/background.js
```diff
@@ -1,29 +1,5 @@
...
-
-function promisifyAll(obj, list) {
-  list.forEach(api => bluebird.promisifyAll(obj[api], { promisifier }));
-}
-
-// let chrome extension api support Promise
-promisifyAll(chrome, ['tabs', 'windows', 'browserAction']);
-promisifyAll(chrome.storage, ['local']);
-
 const onIconClicked = () => {
```
sarveshwar-singh: may be we can make it more meaningful like:
`onIconClicked` -> `onYoroiIconClicked`

SebastienGllmt:
I noticed you removed all usage of `bluebird`. If this is really no longer required, you can remove it from the `package.json` also. I never removed it because I never knew what this line is referring to 
```
// let chrome extension api support Promise
```

SebastienGllmt:
I added the debounce back but setting `leading: true` since I feel like debounce is what we want in this case.

PR #488, opened 2019-05-17
Start Page CSS

ladomyriak:
Modern theme for Start Page


app/themes/prebuilt/YoroiModern.js
```diff
@@ -136,6 +136,16 @@ const rpTextArea = {
   '--rp-textarea-text-color': '#353535',
 };
 
+// OPTION CARD
+const rpItemCard = {
+  '--rp-card-bg-color': '#fff',
```
SebastienGllmt: rp is a prefix for "ReactPolymorph" meant to indicte that this color is meant to be used as an override for the prebuilt components that come with React-Polymorph.
Since this is not the case for you, can you remove the rp prefix (both here and in the classic file)

app/components/wallet/WalletAdd.js
```diff
@@ -104,56 +99,40 @@ export default class WalletAdd extends Component<Props> {
...
-              />
+      <div className={styles.hero}>
+        <div className={styles.heroInner}>
+          <div className={styles.heroLeft}>
+            <SvgInline svg={logo} className={styles.heroLogo} />
+            <h2 className={styles.title}>{title}</h2>
+            <h3 className={styles.subTitle}>{intl.formatMessage(messages.subTitle)}</h3>
+          </div>
+          <HorizontalFlexContainer>
+            {/* eslint-disable */}
```
SebastienGllmt: Why did eslint have to be disabled here?

app/themes/prebuilt/YoroiModern.js
ladomyriak: Fixed. Thanks

app/components/wallet/WalletAdd.js
ladomyriak: fixed with button wrapper for cards. there was an error about accessibility.

PR #489, opened 2019-05-17
shin/ch1071/fix-formfieldownskin-react-warning-keys

sarveshwar-singh:
REF:
https://app.clubhouse.io/emurgo/story/1071/fix-formfieldownskin-react-warning-keys

### DIFF
--- a/app/themes/skins/AutocompleteOwnSkin.js
+++ b/app/themes/skins/AutocompleteOwnSkin.js
@@ -71,7 +71,7 @@ export const AutocompleteOwnSkin = (props: Props) => {
     if (props.selectedOptions && !props.renderSelections) {
       // render default skin
       return props.selectedOptions.map((selectedOption, index) => (
-        <span className={theme.selectedWordBox} key={index.toString()}>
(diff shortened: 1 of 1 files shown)


SebastienGllmt:
Not sure I would consider the index a unique key per element and this could maybe cause a subtle bug in the future but I'm not sure if there's anything else we can do since `selectedOptions` is of type `Array<any>`. Only thing I can think of would be taking a hash of the JSON representation but maybe this is too computationally expensive to bother.

sarveshwar-singh:
@SebastienGllmt 
Changed:
`key={index.toString()}` -> `key={`${selectedOption}-${index.toString()}`}`

Followed the similar use-case:
https://github.com/Emurgo/yoroi-frontend/blob/e23f59db3007a94720d21bb7f07f564fde9c75fb/app/components/wallet/backup-recovery/WalletRecoveryPhraseEntryDialog.js#L129

https://github.com/Emurgo/yoroi-frontend/blob/e23f59db3007a94720d21bb7f07f564fde9c75fb/app/components/wallet/backup-recovery/WalletRecoveryPhraseEntryDialog.js#L166

PR #490, opened 2019-05-17
shin/ch1059/fix-app-components-widgets-copyableaddress

sarveshwar-singh:
**Note:**
- [X] https://github.com/Emurgo/yoroi-frontend/pull/490/commits/bbc637808d25871798b28897ed6a7dd8443bdd47 As Dialog is not a child of `#root` or `.Yoroi[Classic]/[Modern]` we need to change our policy for Dialogs.

image
```
/* Dialog is spacial case as it's not a child of React #root.
* For Theme change management, setting current theme is nessesary like below
* classicTheme ? THEMES.YOROI_CLASSIC : THEMES.YOROI_MODERN
* This should be updated on addition/deletion of theme */
```

- [X] Checked following dialogs in both of the Themes:
image

### DIFF
--- a/app/components/widgets/Dialog.js
+++ b/app/components/widgets/Dialog.js
@@ -3,6 +3,7 @@ import _ from 'lodash';
 import { observer } from 'mobx-react';
 import classnames from 'classnames';
 import type { Node } from 'react';
+import { THEMES } from '../../themes';
 import { Modal } from 'react-polymorph/lib/components/Modal';
 import { Button } from 'react-polymorph/lib/components/Button';
 import { ButtonSkin } from 'react-polymorph/lib/skins/simple/ButtonSkin';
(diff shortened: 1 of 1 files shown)


SebastienGllmt:
Why does it need to be outside of `#root`? Also, why can we just not put `root` inside of another div
```
<body>
    <div id="theme">
        <div id="root"></div>
    </div>
  </body>
```
and then apply the current theme to the `theme` div inside `App.js`?

sarveshwar-singh:
@SebastienGllmt 
> Why does it need to be outside of #root?

It's default behaviour and I could not find any props of `Modal` to set it's parent component (e.g #root).
And I figure out there is reason of keeping `Modal` in the sibling of `#root`
From: https://medium.freecodecamp.org/how-to-render-modals-in-react-bbe9685e947e

> why can we just not put root inside of another div
```
<body>
    <div id="theme">
        <div id="root"></div>
    </div>
  </body>
```
I don't think we can do this easily, React app starts with `#root` we need tweak it's life-cycle.
But we can change `<body>` which is easier but with another problem.
we can use like:
```
document.body.classList.remove('LastTheme');
document.body.classList.add('ThemeA');
```
But for this we need to change few things like, to know/store what was the last Theme (`LastTheme`)
REF: https://brettdewoody.com/modifying-body-attributes-in-react/

sarveshwar-singh:
As more robust fix is in: https://github.com/Emurgo/yoroi-frontend/pull/497
Hence closing this.

PR #501, opened 2019-05-22
[ClassicTheme] fix CREATE A NEW WALLET dialog

sarveshwar-singh:
REF: https://github.com/Emurgo/yoroi-frontend/pull/434

#### 2. https://github.com/Emurgo/yoroi-frontend/pull/501/commits/a0ea1d581fd0e71b06ef3d112970b839980339eb WalletCreateDialog apply new scss architecture

This story started with fixing this:
image
But after looking into code, I found out that it was not intentional to change the old design(it was bug in SCSS), after fixing SCSS bugs, it automatically restored(without changing anything in `WalletCreateDialog.js`) like before:
image

### DIFF
--- a/app/components/widgets/forms/PasswordInstructions.scss
+++ b/app/components/widgets/forms/PasswordInstructions.scss
@@ -1,42 +1,18 @@
-.passwordInstructionsClassic {
+.component {
   color: var(--theme-instructions-text-color);
   font-family: var(--font-light);
   line-height: 1.38;
   margin-top: 20px;
+}
+
+:global(.YoroiClassic) .component {
   opacity: 0.5;
 }
 
-.passwordInstructions {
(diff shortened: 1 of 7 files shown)


app/components/widgets/forms/PasswordInstructions.scss
```diff
@@ -1,42 +1,18 @@
...
-.passwordInstructions {
-  color: var(--theme-instructions-text-color);
+:global(.YoroiModern) .component {
   font-family: var(--font-regular);
   line-height: 1.71;
   margin-top: 14px;
   font-size: 14px;
   letter-spacing: 0;
-
-  ul {
```
sarveshwar-singh: These were unused, as this component only returns `<p>`
Confirmed with @vantuz-subhuman

PR #504, opened 2019-05-23
fix transactions page

sarveshwar-singh:
### Classic Theme
image

### Modern Theme
image

### DIFF
--- a/app/components/wallet/transactions/Transaction.scss
+++ b/app/components/wallet/transactions/Transaction.scss
@@ -213,7 +213,6 @@
 :global(.YoroiClassic) .component {
   // ========= TOGGLER =========
   .time {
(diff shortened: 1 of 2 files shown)


app/components/wallet/transactions/Transaction.scss
```diff
@@ -190,12 +212,12 @@
 
 :global(.YoroiClassic) .component {
   // ========= TOGGLER =========
-  .amount {
-    letter-spacing: 1px;
+  .time {
+    // margin-left: 12px;
```
SebastienGllmt: what is this comment for?

sarveshwar-singh: Fixed, line is not needed.

PR #507, opened 2019-05-24
Update wallet name in top bar section after changing it from settings

leivaburto:
Added a fixed to automatically trigger update of wallet name in the top bar after changing it in the settings (which was not happening before). 

Clipboard 2019-23-05 at 9 54 04 PM


nicarq:
[ch1119]

PR #519, opened 2019-05-26
shin/ch1165/sync-transactions-page-with-proposed-design-2

sarveshwar-singh:
Refer: https://projects.invisionapp.com/d/main#/console/17159202/355689453/preview

## 1. Syncing Transaction page

- ClassicTheme
image

- ModernTheme
image

#### 2. Fixing: https://app.clubhouse.io/emurgo/story/1040/fix-incorrect-logic-on-walletsummarypage-ancient


app/containers/wallet/WalletSummaryPage.js
```diff
@@ -20,13 +20,6 @@ const { formattedWalletAmount } = resolver('utils/formatters');
 
 type Props = InjectedProps
 
-const messages = defineMessages({
-  noTransactions: {
-    id: 'wallet.summary.no.transactions',
-    defaultMessage: '!!!No recent transactions',
```
SebastienGllmt: Are we okay with removing this given we have no alternative for the classic theme?

sarveshwar-singh: @SebastienGllmt 
I do not understand how it's related with any Theme.
It's kind of dead code, which is never used.
ATM, we only use `globalMessages.noTransactionsFound`

image

Which was used in a dead block.
image

Issue #344, opened 2019-03-11
TrezorConnect API firefox manifest setup


sarveshwar-singh:
- check trezor integration on firefox
- update TrezorConnect API firefox manifest setup accordingly (by default manifest is pointing to Chrome extension app)

sarveshwar-singh:
Fixed In: https://github.com/Emurgo/yoroi-frontend/issues/344

PR #520, opened 2019-05-26
shin/fix-trezor-API-firefox-manifest-setup-#344

sarveshwar-singh:
Fix of: https://github.com/Emurgo/yoroi-frontend/issues/344

### 1. https://github.com/Emurgo/yoroi-frontend/pull/520/commits/5b0bd05834b6fb0896bce1745246e6e543c099b4 Remove os module and platform property from environment
It's NodeJS Core Module: https://github.com/DiegoRBaquero/node-os
`platform` property is not used in the project

### 2. https://github.com/Emurgo/yoroi-frontend/pull/520/commits/d9cabe00ad9cd7541eecad3a8433df328488331c added UserAgentInfo utility
Easy to use utility for determining current browser/OS/architecture (needed for https://github.com/Emurgo/yoroi-frontend/pull/520/commits/8b37c6ba8313ef8262b1131a5218b18572a2f3d5)

### 3. https://github.com/Emurgo/yoroi-frontend/pull/520/commits/8b37c6ba8313ef8262b1131a5218b18572a2f3d5 Actual Trezor API firefox manifest setup
Also made same email-Id for both Trezor and Ledger to `rnd@emurgo.io`.
(Before for Trezor it was `systems@emurgo.io` ) cc @nicarq 

### 4. https://github.com/Emurgo/yoroi-frontend/pull/520/commits/691f102c67c7a0237f7a5856e9e18ce4fe5d30f0 Add userAgentInfo log in the header of downloadLogs functionality
This will help supporting user problems.
cc @Leo-Emurgo 

**Before:**
```
﻿[INFO] Yoroi v.1.6.0
[INFO] Commit: d06987d242694f34234982983963fbe279c985e8
[INFO] Network: staging
[2019-05-26T21:48:18+09:00] No errors logged.
```
**After:**
```
﻿[INFO] Yoroi v.1.6.0
[INFO] Commit: 8b37c6ba8313ef8262b1131a5218b18572a2f3d5
[INFO] Network: staging
[INFO] User Agent: {
  "ua": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/74.0.3729.169 Safari/537.36",
  "browser": {
    "name": "Chrome",
    "version": "74.0.3729.169",
    "major": "74"
  },
  "engine": {
    "name": "WebKit",
    "version": "537.36"
  },
  "os": {
    "name": "Linux",
    "version": "x86_64"
  },
  "device": {},
  "cpu": {
    "architecture": "amd64"
  }
}
----
[2019-05-26T22:24:53+09:00] No errors logged.
```

### DIFF
--- a/app/stores/ada/TrezorConnectStore.js
+++ b/app/stores/ada/TrezorConnectStore.js
@@ -93,15 +93,19 @@ export default class TrezorConnectStore extends Store implements HWConnectStoreT
     try {
       /** Starting from v7 Trezor Connect Manifest has been made mandatory
         * https://github.com/trezor/connect/blob/develop/docs/index.md#trezor-connect-manifest */
-      const trezorTConfig = Config.wallets.hardwareWallet.trezorT;
-      TrezorConnect.manifest({
-        email: trezorTConfig.manifest.EMAIL,
-        appUrl: trezorTConfig.manifest.APP_URL
-      });
+      const { manifest } = Config.wallets.hardwareWallet.trezorT;
+
+      const trezorManifest = {};
+      trezorManifest.email = manifest.EMAIL;
+      // set appUrl depending upon browser
+      if (environment.userAgentInfo.isChromeBrowser()) {
+        trezorManifest.appUrl = manifest.appURL.CHROME;
+      } else if (environment.userAgentInfo.isFirefoxBrowser()) {
+        trezorManifest.appUrl = manifest.appURL.FIREFOX;
+      }
+      TrezorConnect.manifest(trezorManifest);
 
       /** Preinitialization of TrezorConnect API will result in faster first response */
-      // TODO [TREZOR]: sometimes when user does fast action initialization is still not complete
-      // try to use same approach as ledger [for now moving this from _init() to setup()]
       TrezorConnect.init({});
     } catch (error) {
       Logger.error(`TrezorConnectStore::setup:error: ${stringifyError(error)}`);
--- a/app/stores/ada/TrezorConnectStore.js
+++ b/app/stores/ada/TrezorConnectStore.js
@@ -97,11 +97,13 @@ export default class TrezorConnectStore extends Store implements HWConnectStoreT
 
       const trezorManifest = {};
       trezorManifest.email = manifest.EMAIL;
-      // set appUrl depending upon browser
-      if (environment.userAgentInfo.isChromeBrowser()) {
-        trezorManifest.appUrl = manifest.appURL.CHROME;
-      } else if (environment.userAgentInfo.isFirefoxBrowser()) {
+      if (environment.userAgentInfo.isFirefoxExtension()) {
+        // Set appUrl for `moz-extension:` protocol using browser (like Firefox)
         trezorManifest.appUrl = manifest.appURL.FIREFOX;
+      } else {
+        // For all other browser supported that uses `chrome-extension:` protocol
+        // In future if other non chrome like browser is supported them we can consider updating
+        trezorManifest.appUrl = manifest.appURL.CHROME;
       }
       TrezorConnect.manifest(trezorManifest);
 
--- a/app/stores/ada/TrezorConnectStore.js
+++ b/app/stores/ada/TrezorConnectStore.js
@@ -97,7 +97,7 @@ export default class TrezorConnectStore extends Store implements HWConnectStoreT
 
       const trezorManifest = {};
       trezorManifest.email = manifest.EMAIL;
-      if (environment.userAgentInfo.isFirefoxExtension()) {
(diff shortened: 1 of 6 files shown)


app/utils/userAgentInfo.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+
+import UAParser from 'ua-parser-js';
```
SebastienGllmt: Probably the user agent is not the best way to do this. After some googling, it seems like `getURL('/')` (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getURL) to get whether or not your extension prefix is `moz-extension` or `chrome-extension`.
The added benefit of this method is it will work for any Chrome-like browser whereas your solution would fail to detect Chromium which doesn't use "Chrome" as the browser name.

The downside is that this solution won't work if Yoroi is ever hosted as a website (which we don't plan to do anyway) but having the browser name in the logs would be helpful so maybe we should:

1) Use the method I mentioned to find which Trezor manifest to load and in the case that it's neither Firefox nor Chrome, just use Chrome as the manifest (?)
2) Use this library to add browser / OS name to the logs

app/utils/userAgentInfo.js
```diff
@@ -0,0 +1,45 @@
...
+
+export default class UserAgentInfo {
+  /** Object structured like this:
+    * Refer: https://www.npmjs.com/package/ua-parser-js
```
SebastienGllmt: This comment seems a little excessive. Maybe just a link to the NPM package is sufficient if we can't get types for the library.

app/stores/ada/TrezorConnectStore.js
```diff
@@ -93,15 +93,19 @@ export default class TrezorConnectStore extends Store implements HWConnectStoreT
...
+      const { manifest } = Config.wallets.hardwareWallet.trezorT;
+
+      const trezorManifest = {};
+      trezorManifest.email = manifest.EMAIL;
+      // set appUrl depending upon browser
+      if (environment.userAgentInfo.isChromeBrowser()) {
+        trezorManifest.appUrl = manifest.appURL.CHROME;
+      } else if (environment.userAgentInfo.isFirefoxBrowser()) {
+        trezorManifest.appUrl = manifest.appURL.FIREFOX;
+      }
```
SebastienGllmt: needs another `else` in the case it's neither Firefox nor Chrome

app/utils/userAgentInfo.js
sarveshwar-singh: add module in `package.json`

app/utils/logging.js
```diff
@@ -47,7 +47,11 @@ export const downloadLogs = () => {
 // ========== STRINGIFY =========
 
 export const generateLogHeader = () => (
```
sarveshwar-singh: @SebastienGllmt 
I would propose to add two more items logs:
1. Yoroi user selected display language (can get from profile)
2. Browser current display language (using `navigator.language`)
How do you think ?

app/utils/userAgentInfo.js
sarveshwar-singh: @SebastienGllmt 
Yeah I agree.

Few things i would like to change here:
I think `browser.runtime.getURL()` is only Mozilla/Firefox global variable so in case of chrome it will be `undefined`

Instead use `window.location.protocol` which will give result like: 
1. Chrome like: `chrome-extension:`
2. Firefox: `moz-extension:`

app/utils/logging.js
SebastienGllmt: https://app.clubhouse.io/emurgo/story/934/add-language-and-yoroi-version-to-header-requests exists so let @leivaburto know if you do this

sarveshwar-singh: @leivaburto 
Regarding: https://app.clubhouse.io/emurgo/story/934/add-language-and-yoroi-version-to-header-requests

I am not updating, in logs.
```
1. Yoroi user selected display language (can get from profile)
2. Browser current display language (using navigator.language)
```
Thanks @SebastienGllmt

app/utils/userAgentInfo.js
```diff
@@ -3,43 +3,20 @@
...
 
   constructor() {
     this.ua = (new UAParser()).getResult();
   }
 
-  isChromeBrowser() {
-    return this.ua.browser.name === 'Chrome';
+  /** This method returns true for all browser that uses `chrome-extension:` protocol,
+    * hence it will return true for browsers like Google Chrome, Brave */
+  isChromeExtension() {
```
sarveshwar-singh: It's not used now, but can be handful later.

PR #522, opened 2019-05-27
Fix types for Trezor

SebastienGllmt:
We used fake Trezor types while the codebase for Trezor was still being written. We can fix these up a bit now to avoid errors in the future

I had to leave two types until this PR is merged: https://github.com/trezor/connect/pull/404


sarveshwar-singh:
@SebastienGllmt
https://github.com/trezor/connect/issues/282 (Fixed I think)
https://github.com/trezor/connect/issues/350 (Not sure)

There was some bug in the API, if you could please verify that is fixed.

PR #531, opened 2019-05-29
Wallet Add Classic Theme

ladomyriak:
Refer:
https://projects.invisionapp.com/d/main#/console/17217070/363708843/preview
https://projects.invisionapp.com/d/main#/console/17217070/366664616/preview


app/i18n/locales/zh-Hant.json
```diff
@@ -200,18 +200,19 @@
   "transfer.summary.recoveredBalance.label": "恢復的餘額",
   "transfer.summary.transactionFee.label": "交易費用",
   "transfer.summary.transferButton.label": "轉移資金",
-  "wallet.add.dialog.create.description": "Create wallet",
```
SebastienGllmt: please don't change non-EN files 🙏

sarveshwar-singh: Yeah, going to undo all those files, and from now on will never change non-EN.

SebastienGllmt:
Is there any reason why the size of the assets are smaller on the main page compared to the design? They are the proper size on the modern theme but not classic.

### Design
https://projects.invisionapp.com/d/main#/console/17217070/363708843/preview

### Implementation
image

PR #532, opened 2019-05-29
 shin/ch1194/increase-width-of-dialog-having-address-2

sarveshwar-singh:


sarveshwar-singh:
In this PR tried to use architecture (to override at skin level at: https://github.com/Emurgo/yoroi-frontend/blob/b240be289de18e1ca46105b1a2c57d05710e613a/app/components/widgets/Dialog.js#L41) but this architecture has problems (like: https://app.clubhouse.io/emurgo/story/1212/fix-lots-of-undefined-class-on-every-dialog) and It will work if user opens only one dialog,  if user opens some another second dialog then the first dialog's style will be applied to the second dialog. (as props are persistent)

Will continue on first approach:
https://github.com/Emurgo/yoroi-frontend/pull/525

Hence closing this.

PR #533, opened 2019-05-30
shin/fix-dialogs

sarveshwar-singh:
### Only applies to Modern Theme
As in new Theme padding and dialog `min-width`/`max-width` was set in `ModalOverrides.scss` and at Skin level so its was not easy to set custom `Dialog` width.
To fix that a hack was used: https://github.com/Emurgo/yoroi-frontend/blob/b240be289de18e1ca46105b1a2c57d05710e613a/app/components/widgets/Dialog.js#L41
but this architecture had problems in both of the Themes (like: https://app.clubhouse.io/emurgo/story/1212/fix-lots-of-undefined-class-on-every-dialog).
Initially tried this way in: https://github.com/Emurgo/yoroi-frontend/pull/532
and It will work if user opens only one dialog, if user opens some another second dialog then the first dialog's style will be applied to the second dialog. (as props are persistent) 

### New architecture
Rather then setting `padding` at top level, setting `margin`  where we have full control and can be customized easily.

Added following contents:
```
  '--theme-modal-margin-cmn'
  '--theme-modal-min-max-width-cmn'
  '--theme-modal-min-max-width-sm'
  '--theme-modal-min-max-width-lg'
```
Following two property is set at `app/components/widgets/Dialog.scss`
```
  '--theme-modal-margin-cmn'
  '--theme-modal-min-max-width-cmn'
```
Dialogs which needs custom `width` or `margin`  could use following (or add new as needed):
```
  '--theme-modal-min-max-width-sm'
  '--theme-modal-min-max-width-lg'
```

### Fixed following dialogs (which has address on it, other dialogs are OK as per design proposal)
- [X] WalletSendConfirmationDialog
- [X] HWSendConfirmationDialog
- [X] AddressDetailsDialog
- [X] WalletRestoreVerifyDialog
- [X] FinalizeDialog

**#This PR also fixes small bugs**


app/components/wallet/send/HWSendConfirmationDialog.scss
```diff
@@ -28,84 +15,4 @@
       margin-left: 18px;
     }
   }
-
```
sarveshwar-singh: Reused from `WalletSendConfirmationDialog.scss`

PR #567, opened 2019-06-12
Add flow to all files and fix related errors

SebastienGllmt:
For many Javascript files we forgot to enable Flow. I fixed this.
I also added a flow-report package as a dev-dependency so we can easily check flow coverage in the future if we suspect something has gone wrong.


app/components/widgets/ProgressSteps.js
```diff
@@ -9,24 +11,31 @@ import iconCrossSVG from '../../assets/images/widget/cross.inline.svg';
...
 import styles from './ProgressSteps.scss';
 
+// TODO: move to type folder?
+export const StepState = Object.freeze({
+  LOAD: 0,
+  PROCESS: 1,
+  ERROR: 9,
+});
+export type StepStateEnum = $Values<typeof StepState>;
+
```
leivaburto: I think this is already defined in `app/types/HWConnectStoreTypes.js`.

app/components/widgets/ProgressSteps.js
```diff
@@ -9,24 +11,31 @@ import iconCrossSVG from '../../assets/images/widget/cross.inline.svg';
...
 type Props = {
   stepsList: Array<string>,
-  progressInfo: {
-    currentStep : number, // example, 0 = pointing to stepsList[0]
-    stepState: number, // has three states, 0 = LOAD | 1 = PROCESS | 9 = ERROR
-  },
+  currentStep : number, // example, 0 = pointing to stepsList[0]
+  stepState: StepStateEnum,
```
leivaburto: I don't get the change of variables in this file. Is there a reason for removing progressInfo? I think this logic of using progressInfo was replicated from other parts of the code related to HW connection, and it's used in several parts. It doesn't seem to conflict with this file, but maybe the idea was to keep the same notation.

app/utils/routing.js
```diff
@@ -65,6 +68,7 @@ export const buildRoute = (pattern, params) => {
         // - '/path(/:param/:another_param)'
         const paramRegex = new RegExp('(/|\\(|\\)|^):' + paramName + '(/|\\)|\\(|$)');
         routePath = routePath.replace(paramRegex, (match, g1, g2) => {
+          // $FlowFixMe
```
leivaburto: This line, and the one in tabManager seem to be the only FlowFixMe comments without an explanation.

app/utils/tabManager.js
```diff
@@ -24,11 +25,12 @@ const OPEN_TAB_ID_KEY = 'openTabId';
  *
  * Note: this may cause two copies of Yoroi loading at the same time close each other
  */
-export function addCloseListener(window) {
+export function addCloseListener(window: any) {
   window.onstorage = (e: StorageEvent) => {
     // if another Yoroi tab open, close this tab
     if (e.key === OPEN_TAB_ID_KEY) {
       // note: we don't need "tabs" permission to get or remove our own tab
+      // $FlowFixMe
```
leivaburto: I think this can be [avoided by declaring](https://github.com/facebook/flow/issues/1424) chrome: `declare var chrome;`

package.json
```diff
@@ -91,6 +92,7 @@
     "eslint-plugin-react": "7.11.1",
     "file-loader": "2.0.0",
     "flow-bin": "0.86.0",
+    "flow-coverage-report": "^0.6.1",
```
nicarq: I would continue with the current philosophy of specifying the version and not leave it open.

app/api/ada/lib/decrypt.js
```diff
@@ -36,15 +40,24 @@ const hashData = (data) => {
   return hash.digest('hex');
 };
 
-export const decryptRegularVend = (key, data) => decryptWithAES(blake2b(fromMnemonic(key)), data);
-export const decryptForceVend = (key, data) => (
+export const decryptRegularVend = (
+  key: string,
+  data: any,
```
nicarq:
```suggestion
  data: string | Uint8Array,
```

app/api/ada/lib/decrypt.js
```diff
@@ -36,15 +40,24 @@ const hashData = (data) => {
...
+) => decryptWithAES(blake2b(fromMnemonic(key)), data);
+export const decryptForceVend = (
+  key: Array<string>,
+  data: any
```
nicarq:
```suggestion
  data: string | Uint8Array
```

app/api/ada/lib/decrypt.js
```diff
@@ -36,15 +40,24 @@ const hashData = (data) => {
...
   decryptWithAES(blake2b(key[0].trim().toLowerCase() +
     hashData(key[1].trim()) + key[2].trim()), data)
 );
 
 // Recovery service certificates decryption
 export const decryptRecoveryRegularVend = decryptRegularVend;
-export const decryptRecoveryForceVend = (key, data) => {
+export const decryptRecoveryForceVend = (
+  key: string,
+  data: any
```
nicarq:
```suggestion
  data: string | Uint8Array
```

nicarq: (according to comments)

app/components/settings/categories/general-setting/AboutYoroiSettingsBlock.js
```diff
@@ -94,7 +95,7 @@ const socialMediaLinks = [{
 }];
 
 @observer
-export default class AboutYoroiSettingsBlock extends Component {
+export default class AboutYoroiSettingsBlock extends Component<{}> {
```
nicarq: I thought it wasn’t necessary to specify it if it was empty

app/components/widgets/ProgressSteps.js
```diff
@@ -9,24 +11,31 @@ import iconCrossSVG from '../../assets/images/widget/cross.inline.svg';
...
   classicTheme: boolean
 };
 @observer
 export default class ProgressSteps extends Component<Props> {
 
-  createSetps = (stepsList, progressInfo): Array<Element> => {
+  createSetps = (
```
nicarq:
```suggestion
  createSteps = (
```

app/components/widgets/ProgressSteps.js
```diff
@@ -79,10 +88,15 @@ export default class ProgressSteps extends Component<Props> {
   }
 
   render() {
-    const { stepsList, progressInfo } = this.props;
+    const { stepsList, currentStep, stepState } = this.props;
+
     return (
       <div className={styles.component}>
-        {this.createSetps(stepsList, progressInfo)}
+        {this.createSetps(
```
nicarq:
```suggestion
        {this.createSteps(
```

app/components/widgets/ProgressSteps.js
```diff
@@ -48,7 +57,7 @@ export default class ProgressSteps extends Component<Props> {
       } else if (idx === currentStep) {
         // for current step, 0 = LOAD | 1 = PROCESS | 9 = ERROR
         // 0 = LOAD and 1 = PROCESS has same icon but for 9 = ERROR there is a error icon
-        displayIcon = (progressInfo.stepState === 9) ? 'error' : 'none';
+        displayIcon = (stepState === 9) ? 'error' : 'none';
```
nicarq: This is not from you, but I think is kind of ugly to have comparison by numbers. Also having to re-explain something that is declared is strange.

Is it possible to do something like this?

```suggestion
        displayIcon = (stepState === StepState.ERROR.valueOf()) ? 'error' : 'none';
```

app/domain/WalletTransaction.js
```diff
@@ -5,37 +5,29 @@ import type { AssuranceMode, AssuranceLevel } from '../types/transactionAssuranc
...
 export type TrasactionAddresses = { from: Array<string>, to: Array<string> };
-export type TransactionState = 'pending' | 'failed' | 'ok';
-export type TransactionType = 'card' | 'expend' | 'income' | 'exchange' | 'self' | 'multi';
 
-export const transactionStates: {
-  PENDING: TransactionState,
-  FAILED: TransactionState,
-  OK: TransactionState,
-} = {
+export const transactionStates = Object.freeze({
```
nicarq: maybe we should enable `'use strict';` for this 🤔 at least for dev mode

app/i18n/global-messages.js
```diff
@@ -300,7 +301,10 @@ export const environmentSpecificMessages = {
   }),
 };
 
-export function listOfTranslators(contributorsList, contributorsAck): string {
+export function listOfTranslators(
```
nicarq: this is not from this PR, but it’s kind of odd to call something a list when is a string 🤔

app/utils/routing.js
```diff
@@ -30,8 +31,10 @@ export const buildRoute = (pattern, params) => {
   const tokens = {};
 
   if (params) {
+    // assert not null
+    const paramsArgs = params;
```
nicarq: I didn’t get this part.

package-lock.json
```diff
@@ -1281,6 +1281,17 @@
         }
       }
     },
+    "@rpl/badge-up": {
```
nicarq: this is a lot of new dependencies 🤔 (634 new lines) is it just from `flow-coverage-report`?

app/utils/routing.js
SebastienGllmt: because this one is tricky to get working correctly so I didn't want to spend time on it whereas the other places it's actually Flow that is wrong and so I just `$FlowFixMe` to ignore it.

app/utils/tabManager.js
SebastienGllmt: doing `declare var chrome` just silences the error same way `$FlowFixMe` does except it does it for the whole file instead of a single line. It doesn't actually add the missing  type definitions.

app/components/widgets/ProgressSteps.js
SebastienGllmt: Because the type of `currentStep` depends on where the progress bar is used (you can imagine different processes requiring a different number of steps, all represented by their own enum type). Suppose you have the two types:
```
type Foo: {
    enum: number;
}

type MyEnum = 1 | 2 | 3;
type Bar: {
    enum: MyEnum;
}
```

You can't cast `Bar` to `Foo`. You can, however, do `Foo.enum = Bar.enum`

For this reason, we can't keep the progressInfo as previously defined without ugly code so it's easier to just flatten it.

app/domain/WalletTransaction.js
nicarq: @SebastienGllmt in practice if you try to change a object that was created using Object Freeze, nothing will happen, but if strict mode is enabled it will throw an error, which I think it will be good to let us know that we are doing something wrong

nicarq: Try the playground here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze and add the strict mode to see how is different

app/utils/routing.js
nicarq: Gotcha. Thanks!

package-lock.json
SebastienGllmt: Agreed. Removed the dependency because of this.

app/utils/routing.js
leivaburto: Oh, I meant just adding a comment like "$FlowFixMe : this one is actually tricky because (...)", given that you already know why it's tricky, so the rest of us can have a hint of what's going on. For example, the one about chrome: "$FlowFixMe : flow complains about chrome not being declared".  🙏

app/domain/WalletTransaction.js
SebastienGllmt: Javascript modules are strict by default (and in fact I don't think you can make them non-strict) so we already have this behavior

nicarq: http://www.ecma-international.org/ecma-262/6.0/#sec-strict-mode-code

PR #571, opened 2019-06-12
Add warning for slow internet connections on Daedalus transfer

SebastienGllmt:
As the blockchain gets bigger, Daedalus transfer takes much longer. Some users complain it doesn't work when in reality it is just their internet connection that isn't stable throughout the transfer process.

image

### DIFF
--- a/app/components/transfer/AnnotatedLoader.scss
+++ b/app/components/transfer/AnnotatedLoader.scss
@@ -27,5 +27,9 @@
     margin-bottom: 12px;
     word-break: break-word;
     text-align: center;
+
+    .error {
+      color: var(--rp-theme-color-error);
+    }
   }
 }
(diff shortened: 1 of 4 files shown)


app/i18n/locales/en-US.json
```diff
@@ -66,6 +66,7 @@
   "daedalusTransfer.waiting.progressInfo.checkingAddresses": "Checking addresses funds",
   "daedalusTransfer.waiting.progressInfo.restoringAddresses": "Fetching addresses",
   "daedalusTransfer.waiting.title.label": "Daedalus wallet is being restored",
+  "daedalusTransfer.waiting.warning": "⚠️ This may take a long time or fail on poor internet connections",
```
sarveshwar-singh: not sure how crowdin react on ⚠️  char :)

SebastienGllmt: I don't see why it would be a problem given it supports the rest of unicode just fine

app/components/transfer/AnnotatedLoader.scss
```diff
@@ -27,5 +27,9 @@
     margin-bottom: 12px;
     word-break: break-word;
     text-align: center;
+
+    .error {
```
sarveshwar-singh: may be we wanna name it `warning` as it shows warning :)

PR #624, opened 2019-06-17
shin/ch1453/fix-verify-restore-dialog-button-name

sarveshwar-singh:
## Before:
image

## After:
image


SebastienGllmt:
Is this what we want for the solution? Now we have 2 back buttons on the page (the arrow and the button). 
Maybe we should instead add an "X" button this dialog and then remove this button entirely?

PR #628, opened 2019-06-17
Remove default letter-spcing on modern theme

SebastienGllmt:
Marta has requested this multiple times. This changes how many screens on the modern theme look so we need to be kind of careful about merging it.

## Before

image

## After 

image

I also noticed many screens manually set a different default letter-spacing. For example, the paper wallet screen manually sets `letter-spacing: 0.5px` so this PR doesn't affect them.

### DIFF
--- a/app/components/wallet/WalletAdd.scss
+++ b/app/components/wallet/WalletAdd.scss
@@ -38,6 +38,7 @@
   .heroTitle {
     font-family: var(--font-bold);
     font-size: 54px;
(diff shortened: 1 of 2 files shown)


sarveshwar-singh:
## Left => current PR, Right => `develop` branch
image

nicarq:
@sarveshwar-singh nice catch! I just tried this branch and that was the only issue that I saw. I think if @SebastienGllmt or someone else fixes it we can merge

PR #630, opened 2019-06-17
Generic build compress script

rcmorano:
build-compress[-crx]' yarn script that uses environment var '\$CARDANO_NETWORK' as template (if empty, defaults to 'testnet')

### DIFF
--- a/docs/RELEASE.md
+++ b/docs/RELEASE.md
@@ -7,9 +7,9 @@ To create a release candidate you must compress a local build.
 Zip files can be shared
 
 ```bash
-# compress build folder to {manifest.name}.zip and crx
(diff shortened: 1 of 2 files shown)


SebastienGllmt:
Did you want to delete the other build scrips and then update the project README build steps for this?

rcmorano:
> Did you want to delete the other build scrips and then update the project README build steps for this?

Since `production` case is a bit special, we should keep it (statically pointing to `mainnet`) so we don't get confused for it. 
I just got rid of `staging` scripts, so we can use the generic approach for any other environment/network and updated the [release doc](docs/RELEASE.md).

PR #636, opened 2019-06-17
shin/ch1402/increase-tab-height-from-40px-to-50px 

sarveshwar-singh:
Before:
image

After:
image

Refer:
https://projects.invisionapp.com/d/main#/console/17159202/355689457/comments/116633915

### DIFF
--- a/app/components/wallet/layouts/WalletWithNavigation.scss
+++ b/app/components/wallet/layouts/WalletWithNavigation.scss
@@ -5,21 +5,21 @@
 }
 
 .navigation {
-  height: 40px;
+  height: var(--theme-wallet-navigation-tab-height);
   flex-shrink: 0;
 }
 
 .page {
-  height: calc(100% - 40px);
+  height: calc(100% - var(--theme-wallet-navigation-tab-height));
   overflow: auto;
 }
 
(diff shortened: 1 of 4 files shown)


app/components/wallet/layouts/WalletWithNavigation.scss
```diff
@@ -13,3 +13,13 @@
   height: calc(100% - 40px);
   overflow: auto;
 }
+
+:global(.YoroiModern) {
+  .navigation {
+    height: 50px;
```
SebastienGllmt: this should probably be a variable that you import from the other file instead of a magic number

nicarq: yeah, I agree. This number is used in multiple parts and there are some calculations using it, so it shouldn’t be hardcoded.

sarveshwar-singh: @SebastienGllmt @nicarq 
https://github.com/Emurgo/yoroi-frontend/pull/636/commits/ddfdbe509a0c4ce1afe98c710176c24766af6990

Initially i though i should make a const, but don't remember why I quit that idea :)

PR #647, opened 2019-06-19
Exactify prop types

yushih:
Summary of changes:
1. Replace `type Props = { ... }` with `type Props = {| ... |}`. Now the code base is free of such pattern except in skin components.
2. In https://github.com/Emurgo/yoroi-frontend/commit/482c8fbefb44b3a87f034391d16345eb50f5e069#diff-eb964c5702091af0927b33c347e96961 , remove unused parameter passings.
3. In https://github.com/Emurgo/yoroi-frontend/commit/5d525be70fe5b1a7deda5a5b2b22dfe86e2d944a#diff-41248b6193806412a01d9511ec0a43ee , remove the duplicate `classicTheme` property <del>and remove unused parameters passings</del>.

### DIFF
--- a/app/components/wallet/skins/AmountInputSkin.js
+++ b/app/components/wallet/skins/AmountInputSkin.js
@@ -15,7 +15,9 @@ const messages = defineMessages({
   },
 });
 
-type Props = {|
+// This type should be kept open (not "exact") because it is a react-polymorph skin
+// and should be able to pass any extra properties from react-polymorph down.
+type Props = {
   currency: string,
   fees: BigNumber,
   total: BigNumber,
@@ -27,7 +29,7 @@ type Props = {|
   themeId: string,
   value: string,
   type: string,
-|};
+};
 
 export default class AmountInputSkin extends Component<Props> {
   static defaultProps = {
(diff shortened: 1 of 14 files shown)


app/components/wallet/skins/AmountInputSkin.js
```diff
@@ -40,12 +39,33 @@ export default class AmountInputSkin extends Component<Props> {
...
+      theme,
+      themeId,
+      value,
+      type,
+    } = this.props;
     const { intl } = this.context;
 
     return (
       <div className={styles.component}>
-        {classicTheme ? <InputSkin {...this.props} /> : <InputOwnSkin {...this.props} />}
```
SebastienGllmt: This is one of those cases where I mentioned we don't use exact props on purpose in order to be able to pass arbitrary props down to child components (you can think of it like inheritance).
The only way we could make this exact is by modifying the `React-Polymorph` library to make it export its Prop types for the skins and then do a type-level union (which is maybe not a bad idea to make a PR for this to React-Polymorph)

yushih: I was aware of the situation but I thought the above reason only applied to the `<InputSkin {...this.props} />` part. In the case of `<InputOwnSkin>` (https://github.com/Emurgo/yoroi-frontend/blob/develop/app/themes/skins/InputOwnSkin.js#L65) the extra properties in AmountInputSkin.js are not actually passed down anywhere. However this change seems a little bit aggressive to me indeed... Should I revert?

yushih: Ah I was wrong. Scratch above comment.

yushih: Yeah, props types of skins should be not be exact given the way react-polymorph is.

PR #734, opened 2019-07-01
fixing columns alignment in Tx list

YankoFelipe:
Before:
Screen Shot 2019-07-01 at 16 58 14

After:
Screen Shot 2019-07-01 at 16 58 30


YankoFelipe:
I've check the other languages and the longest one was 96px, I used 100px for safety reasons.

sarveshwar-singh:
Update In: https://github.com/Emurgo/yoroi-frontend/pull/734/commits/9e15a8d2d3d7a4a9c7683c305d69e873710b396b

**Classic Theme:**
image

**Modern Theme:**
image

sarveshwar-singh:
@nicarq @YankoFelipe 
> I still see the alignment issues. Also, using min-width doesn’t look like the right solution.
The problem was `min-width` was not set only for `time` component and was set for other component.

As long as we check all language and find the maximum possible width and set it as `min-width` it will work.

**But of course it has limitation like: 'Need to check when new language added'.**

YankoFelipe:
I've just checked Indonesian and it looks good too
Screen Shot 2019-07-03 at 09 38 43

PR #742, opened 2019-07-02
Black label for input fields in classic theme

leivaburto:
Keep black labels for input fields in error state for Classic Theme. Currently, labels are red and it's an overwhelming amount of error state.

**Before**
red_labels_errors

**After**
black_labels_error

### DIFF
--- a/app/themes/skins/FormFieldOwnSkin.scss
+++ b/app/themes/skins/FormFieldOwnSkin.scss
@@ -47,11 +47,6 @@
   }
 }
 
-.error {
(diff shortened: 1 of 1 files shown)


YankoFelipe:
Does the modern theme will continue to have the titles in red when there's an error?

leivaburto:
> Does the modern theme will continue to have the titles in red when there's an error?

Yes! In Modern theme labels are part of the input field border, so having a red label combines better. So yes, modern keeps the same.

red_label_error_modern

PR #748, opened 2019-07-03
Downloads Logs button in Support section

leivaburto:
Duplicating download functionality with a more visible button. Also renaming the name of the section from **Support** to **Support / Logs**.

download_logs_button


leivaburto:
@Marta-Ryzhok this is the first approach. I'll appreciate your feedback.

Marta-Ryzhok:
yes I'm ok with it

Issue #382, opened 2019-04-04
Doubled requests to the server


ebeloded:
There is an issue regarding Yoroi communication with the backend. For some endpoints, two identical calls to the server are being made, where one should expect just one.

To see this, just open the app and take a look at XHR requests being made by the client. You will see the problems immediately with `filterUsed` and `history` endpoints. 
![image](https://user-images.githubusercontent.com/5859177/55524323-2322d500-56b7-11e9-917d-e155dfeac328.png)

Instead of this ☝️, and taking into account CORS, we should expect to see one 204, followed by a 200 for each unique POST request.

Notably, `utxoSumForAddresses` doesn't have the problem and only makes one request.

The result of this is that the client makes unnecessary requests, and the server has processes twice as many requests as needed. The problem is exaserbated by the fact that we make these requests every 20 seconds.

The problem is likely caused by a double subscription.

SebastienGllmt:
This is filtering gone wrong where you're supposed to make two requests -- one for internal and one for external. You could argue the `utxoSumForAddresses` should also be split in the same way but then way UTXOs are handled should probably be changed entirely when we move to Redux.

YankoFelipe:
Fixed on https://github.com/Emurgo/yoroi-frontend/pull/753

PR #753, opened 2019-07-05
Fixing double request to the server

YankoFelipe:
Fixing https://github.com/Emurgo/yoroi-frontend/issues/382


SebastienGllmt:
Let's not merge this now since we don't want to break something for 1.8

PR #754, opened 2019-07-05
Prep work for URI Prompt page

SebastienGllmt:
The new app start flow will be the following

Pick language => Accept ToS => Accept/decline URI scheme
Thn, based on the accept or decline, we have another page that either tells them to press OK for Chrome permission OR says sorry you aren’t interested but you can enable it in the settings page at any time

To do this, I have to make it cleaner to add new steps to app start. In the same PR, I also modify `isChrome` and `isFirefox` so that they work in browser mode also.

### DIFF
--- a/app/stores/toplevel/ProfileStore.js
+++ b/app/stores/toplevel/ProfileStore.js
@@ -1,5 +1,5 @@
 // @flow
-import { observable, computed } from 'mobx';
+import { action, observable, computed, runInAction } from 'mobx';
 import BigNumber from 'bignumber.js';
 import moment from 'moment/moment';
 import Store from '../base/Store';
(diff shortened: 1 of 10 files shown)


app/stores/toplevel/ProfileStore.js
```diff
@@ -39,6 +39,26 @@ export default class ProfileStore extends Store {
...
+  /**
+   * Need to store the selected language in-memory for when the user
+   * is at the language select screen. Only commit to storage once the user accepts.
+   */
+  @observable
+  inMemoryLanguage: string = ProfileStore.getDefaultLocale();
+
+  /** Linear list of steps that need to be completed before app start */
+  @observable
+  STEUP_STEPS = [
```
YankoFelipe: Isn't this `SETUP_STEPS`?

YankoFelipe:
> 2. Second one is more critical. If you open Yoroi for the first time, you can go directly to `chrome-extension://bflmcienanhdibafopagdcaaenkmoago/main_window.html#/wallets/add` and bypass the language and term of use screens.

This is happening in `develop` too 😮

PR #776, opened 2019-07-07
Implement UI for URI Scheme Prompt

SebastienGllmt:
This adds the UI for registering the UI scheme in Yoroi. We needed this because for Chrome, registering new protocol handler requires user permission and the pop-up asking for user permission is only triggered by a user action such as a button click.

Note the behavior of this feature is different depending on if you're using Chrome, Firefox and whether or not you're running Yoroi as a website.

- [x] signoff on text @nicarq 
- [x] add classic icons
- [x] signoff on UI @Marta-Ryzhok 
  - [x] make asset bigger on UriSkip
  - [x] merge text on UriPromptForm
  - [x] add warning banner to top of page
  - [x] line up buttons across screens

#### On app start (every single Yoroi user will see this when they launch 1.8)

image

image

#### If you press Skip

image

image

#### If you press allow

image

image

### Settings page

image

image


nicarq:
@SebastienGllmt looks great! The only feedback (small UI details):

- the Allow button in Settings could look a little bit better if we match the upper margin to the text to be the same as the margin to the separator line.
- In the three screen with texts, the buttons are like 5 pixels off so they have the same margin as the text

SebastienGllmt:
@v-almonacid yeah you're right. Fixed it

PR #862, opened 2019-07-29
Add browserlist

SebastienGllmt:
Right now we don't specify what browsers `babel` should perform transforms form. The **default behavior is to perform all transforms for es2015**. I tried restricting it to

```
last 2 chrome version
last 2 ChromeAndroid version
last 2 firefox version
last 2 FirefoxAndroid version
```

This reduces build size by 0.1MB and reduces build time by ~5 seconds (35 seconds => 30 seconds)

There's a certain danger with limiting the # of versions of Chrome & Firefox we support but it seems most people stay up-to-date. You can see the latest stats [here](https://caniuse.com/usage-table). Not sure if the benefits are worth the risk

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,4 +1,4 @@
-last 2 chrome version
-last 2 ChromeAndroid version
-last 2 firefox version
-last 2 FirefoxAndroid version
+Firefox >= 67
+Chrome >= 69
+FirefoxAndroid >= 67
+ChromeAndroid >= 69
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,4 @@
+last 2 chrome version
+last 2 ChromeAndroid version
+last 2 firefox version
+last 2 FirefoxAndroid version


nicarq:
@SebastienGllmt last 2 versions of Chrome and Firefox seems pretty risky. I’m not sure how often Brave and Yandex update their browsers. 

I saw that some people use something less restrictive like:
``` 
"browsers": [
  ">0.25%",
  "not ie 11",
  "not op_mini all"
]
```

SebastienGllmt:
I was going to close this but I realized probably it's not bad if we just specify the earliest versions that work for both browsers. AKA we know Yoroi fails (due to missing WASM functionality) on versions below the ones I specified.

Since this is not as strict, it only cuts ~1sec off build time and only around 0.05MB

SebastienGllmt:
Merging this now since it's very safe and also cuts build time by like 20% (10-15 seconds)

PR #892, opened 2019-08-10
v2 Message Signing Spec

SebastienGllmt:
Will record a video on this later.


docs/specs/code/SIGNING.md
```diff
@@ -0,0 +1,424 @@
...
+### *CBOR Object Signing and Encryption (COSE)* [[RFC 8152](https://tools.ietf.org/html/rfc8152)] 
+
+This is a standard for how to use CBOR for message signing. It is based on the JSON equivalent *JSON Object Signing and Encryption* [RFC 7520](https://tools.ietf.org/html/rfc7520). 
+
+We base our construction on COSE because all Cardano libraries already depend on CBOR due to its use in the base protocol (which means we don't need to introduce a new library). It is also more compact, which is useful in case data generated by this standard ever needs to be stored on-chain.
+
+Note that we do NOT follow the COSE specification exactly for the following reason:
+
+- We need to restrict the number of algorithms used to only ones that would already be required to provide base Cardano functionality (see E1).
+- We need to enforce the usage of a digest if the input string is not ASCII (see E3)
```
SebastienGllmt: Probably can remove this line since I don't think it needs any changes beyond just a new header element

docs/specs/code/SIGNING.md
```diff
@@ -0,0 +1,424 @@
...
+
+The `Headers` MUST have `enc_type: 1` in the `unprotected` field.
+
+The `ciphertext` is simply the result of [the Yoroi encryption spec](https://github.com/Emurgo/yoroi-frontend/blob/737595fec5a89409aacef827d356c9a1605515c0/docs/specs/code/ENCRYPT.md) on the `signed_message`.
+
+#### Public key based encryption
+
+We only allow encrypting based on `ED25519 extended` public keys (the ones used for Cardano). To encrypt based on these public keys, you must
+
+1) Compute a password consisting of 22 case-sensitive alphanumeric (a–z, A–Z, 0–9) characters (this gives you ~128 bits of entropy)
```
SebastienGllmt: Using this technique allows us to encrypt a message for multiple receivers but it's slightly more bytes than needed for just a single receiver (which is probably the more common case). 

We could have a single-receiver pubkey-based encryption mode where it uses the `COSE_Encrypt0` format with the hash of the shared secret as the password.

docs/specs/code/SIGNING.md
```diff
@@ -0,0 +1,424 @@
...
+PubKeyEncryption = 96 (COSE_Encrypt)
+```
+
+The `Headers` for the body MUST have `enc_type: 2` in the `unprotected` field.
+
+The `Headers` for the recipient MUST have a `eph_pub: bstr` label containing the public key of the ephemeral keypair.
+
+## Multisig
+
+TODO
```
SebastienGllmt: COSE has a counter signature concept which is meant for multisig but I didn't spend too much time looking into it for now

docs/specs/code/SIGNING.md
```diff
@@ -0,0 +1,424 @@
...
+
+## More headers
+
+### CWT
+
+CWT tags MAY be introduced inside `protected` headers as a CBOR map with the key `cwt`.
+
+### Payload encoding
+
+To solve `E3`, `signed_message` body header MUST `is_ascii: bool` as a `protected` header which defined whether or not the bytes the `payload` should be considered as ASCII .
```
SebastienGllmt: I only encode `is_ascii` instead of a general `encoding` header. It may be worth it to have a general `encoding` so we can do `ascii / utf8 / binary` but I figured probably supporting binary isn't really worth the trouble (could just instead sign a string of hexadecimal)

docs/specs/code/SIGNING.md
```diff
@@ -0,0 +1,424 @@
...
+- `signed_message`: `cms_`
+- `mac_message`: `cmm_` (unused in this spec)
+
+### Data
+
+Data is simply the `base64url` encoding of the message
+
+### Checksum
+
+We use `CRC32` on the data for the checksum
```
SebastienGllmt: I used `CRC32` since  we use it elsewhere but it seems CRC in general has fallen out of popularity. I'm open for a different checksum method if we want.

docs/specs/code/SIGNING.md
```diff
@@ -0,0 +1,424 @@
...
+The `Headers` for the body MUST have `enc_type: 2` in the `unprotected` field.
+
+The `Headers` for the recipient MUST have a `eph_pub: bstr` label containing the public key of the ephemeral keypair.
+
+## Multisig
+
+TODO
+
+## More headers
+
```
SebastienGllmt: Add mandatory version number

docs/specs/code/SIGNING.md
```diff
@@ -0,0 +1,424 @@
...
+### CWT
+
+CWT tags MAY be introduced inside `protected` headers as a CBOR map with the key `cwt`.
+
```
SebastienGllmt: Mention avoiding replay attacks

docs/specs/code/SIGNING.md
```diff
@@ -0,0 +1,472 @@
...
+#### Public key based encryption
+
+We only allow encrypting based on `ED25519 extended` public keys (the ones used for Cardano). To encrypt based on these public keys, you must
+
+1) Compute a password consisting of 22 case-sensitive alphanumeric (a–z, A–Z, 0–9) characters (this gives you ~128 bits of entropy)
+
+Now, for each receiver, you must
+
+2) Compute an ephemeral key pair using `ED255519 extended`
+1) Compute the shared DH secret between the private key from step (1) and the public key received (using the `exchange` functionality)
```
SebastienGllmt: Is there a case where picking a specific ephemeral key pair makes the shared secret reveal information about a private key? If so, you could for example use this to extract private keys out of your hardware wallet

SebastienGllmt:
closing this as this was moved to the EmIP repo

PR #899, opened 2019-08-26
Key derivation spec

leivaburto:
Key derivation spec, split off from the memo signing spec. We discuss how to generate a private key chain out of the account key pair to be used for purposes other than signing transactions.


SebastienGllmt:
Closing  this as we're putting this in the EmIP repository

PR #940, opened 2019-09-04
Fix Transfer from Another Wallet

SebastienGllmt:
Saving addresses to IndexDB used to be asynchronous. This is fine in the regular restoration case since every address will be saved eventually.

However, for  the "Transfer from Another Wallet"  feature, we need it to be synchronous because we need to accumulate the  addresses into an array to display to the user right away.

### DIFF
--- a/app/api/ada/index.js
+++ b/app/api/ada/index.js
@@ -1196,14 +1196,18 @@ export default class AdaApi {
       const cryptoAccount = createCryptoAccount(masterKey, walletPassword, 0);
 
       const savedAddresses = [];
+      const seenAddresses = new Set<string>();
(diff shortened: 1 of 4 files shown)


yushih:
Just to check my understanding: `saveInitialAddresses` is actually synchronous, so it's not absolutely necessary to add `await` for it.  (But it's good practice. ) `scanAndSaveAddresses` is asynchronous due to its invocation of `discoverAllAddressesFrom`.

But I don't quite understand why the addresses are not unique originally.

SebastienGllmt:
@yushih both are async

The check for uniqueness is because `saveInitialAddresses` generates addresses 0-19 and `scanAndSaveAddresses` also generates starting from 0

PR #942, opened 2019-09-12
Seiza for yoroi integration

bigamasta:


chrome/manifestEnvs.js
```diff
@@ -0,0 +1,5 @@
+// URL of full featured Seiza
+export const SEIZA_URL = process.env.SEIZA_URL || 'http://localhost:3001';
```
SebastienGllmt: Why have both `SEIZA_URL` and `SEIZA_FOR_YOROI_URL`? It seems `SEIZA_URL` isn't used anywhere

bigamasta: Right. So `SEIZA_FOR_YOROI_URL` is the URL of the yoroi version of Seiza, that's directly displayed inside `<iframe>`. 

But the Yoroi version of Seiza has an "Open in Seiza" feature that opens a new tab with original Seiza URL, which is a different URL than Seiza for Yoroi.

We need to have `SEIZA_FOR_YOROI_URL` in  `manifest.json` (`frame-src`) in order to allow displaying it inside an iframe. But we also need `SEIZA_URL` so that original Seiza URL isn't blocked when opening from within the `<iframe>`.

PR #943, opened 2019-09-13
Flow 107

SebastienGllmt:
I need some Flow fixes for storage v2 so I'm bumping this in `develop`.

There are two notable fixes required:

1) `mobx` doesn't work with Flow 103. I fixed the bug for them on `master` but they haven't released a new version in a few months now so I uploaded my own version with the fix. We can move back to `npm` whenever they release a new version.

2) Flow changes the way that spread operators get typechecked and one of the side-effects `$Shape` (which was already terribly broken before) is now even worse. If you really want to know how this new trick works, there are the two steps
    A) Read the change on Flow spread operators [here](https://medium.com/flow-type/coming-soon-changes-to-object-spreads-73204aef84e1)
    B) Read the definition of `$Rest` [here](https://flow.org/en/docs/types/utilities/#toc-rest)


nicarq:
@SebastienGllmt reviewing it today 😄

app/api/ada/index.js
```diff
@@ -232,7 +232,7 @@ export type GetTransactionsRequestOptions = {
   limit: number,
 };
 export type GetTransactionsRequest = {
-  ...$Shape<GetTransactionsRequestOptions>,
+  ...Inexact<GetTransactionsRequestOptions>,
```
nicarq: is it possible to make `GetTransactionsRequestOptions` exact rather than make this Inexact?

PR #946, opened 2019-09-19
Remove dead code

SebastienGllmt:


package.json
```diff
@@ -136,7 +136,6 @@
   "dependencies": {
     "@cardano-foundation/ledgerjs-hw-app-cardano": "1.0.7",
     "@download/blockies": "1.0.3",
-    "aes-js": "3.1.2",
```
nicarq: I think we need to delete:

`pdfjs-dist`
`blakejs`
`react-dropzone`
as well

nicarq:
@SebastienGllmt probably we also need to update the translation files

app/components/settings/menu/SettingsMenu.js
```diff
@@ -52,7 +51,7 @@ export default class SettingsMenu extends Component<Props> {
 
   render() {
     const { intl } = this.context;
-    const { onItemClick, isActiveItem, hasActiveWallet, currentLocale } = this.props;
+    const { onItemClick, isActiveItem, hasActiveWallet, } = this.props;
```
nicarq: `adaRedemption` translation should also be deleted (line ~32)

package.json
SebastienGllmt: we still use `blakejs` for the account plates so we need it. Do we want to remove `react-dropzone` considering we've talked about adding a key import feature in the future that will require re-adding all the dropzone related code?

SebastienGllmt: gave it some thought and decided to remove dropzone. If we decide to add it again later we can look at this PR as a reference.

SebastienGllmt:
@nicarq in general you should not modify non-EN files. It will cause conflicts in Crowdin. Once we remove them from English, Crowdin will automatically remove them from other languages.

PR #951, opened 2019-09-22
No floating Promise Flow implementation

SebastienGllmt:
A regular contributor to Flow helped implement `no-floating-promise` into Flow directly! I built their code myself and ran it on our codebase and fixed the bugs.

See PR: https://github.com/facebook/flow/pull/8100

Worked like a charm!

### DIFF
--- a/app/stores/ada/AdaTransactionBuilderStore.js
+++ b/app/stores/ada/AdaTransactionBuilderStore.js
@@ -37,7 +37,8 @@ export default class AdaTransactionBuilderStore extends Store {
 
   setup() {
     super.setup();
-    this._reset();
+    this._preWaitReset();
+    this._postWaitReset();
     const actions = this.actions.ada.txBuilderActions;
     actions.updateReceiver.listen(this._updateReceiver);
     actions.updateAmount.listen(this._updateAmount);
@@ -208,9 +209,22 @@ export default class AdaTransactionBuilderStore extends Store {
   }
 
   @action
-  _reset = async () => {
+  _preWaitReset = () => {
+    // resets that have to happen BEFORE waiting for createUnsignedTx to finish
     this.plannedTxInfo = [{ address: undefined, value: undefined }];
     this.shouldSendAll = false;
+  }
+  @action
+  _postWaitReset = () => {
+    // resets that have to happen AFTER waiting for createUnsignedTx to finish
+    this.createUnsignedTx.reset();
+    this.plannedTx = null;
+    this.tentativeTx = null;
+  }
+
+  @action
+  _reset = async () => {
+    this._preWaitReset();
 
     // creation of unsigned tx may still be running when we try and reset
     // we need to wait for it to finish then clear the result since we can't cancel the promise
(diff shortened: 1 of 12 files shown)


app/stores/ada/AdaTransactionBuilderStore.js
```diff
@@ -208,9 +209,22 @@ export default class AdaTransactionBuilderStore extends Store {
...
-  _reset = async () => {
+  _preWaitReset = () => {
+    // resets that have to happen BEFORE waiting for createUnsignedTx to finish
     this.plannedTxInfo = [{ address: undefined, value: undefined }];
     this.shouldSendAll = false;
+  }
+  @action
+  _postWaitReset = () => {
+    // resets that have to happen AFTER waiting for createUnsignedTx to finish
+    this.createUnsignedTx.reset();
```
nicarq: Maybe it’s more clear in setup if we take out the createUnsignedTx from here

nicarq: So instead of just having pre and post, we would have pre, create and post

app/stores/ada/TrezorConnectStore.js
```diff
@@ -114,6 +114,7 @@ export default class TrezorConnectStore
       TrezorConnect.manifest(trezorManifest);
 
       /** Preinitialization of TrezorConnect API will result in faster first response */
+      // $FlowFixMe we purposely don't want to wait for execution. Safe in practice.
```
nicarq: Pretty cool that now we have to specified that we want to ignore this on purpose :)

app/stores/ada/AdaTransactionBuilderStore.js
SebastienGllmt: where else would you put it? If we remove it from here, we then have to add it to `_reset` and `setup`

app/stores/ada/AdaTransactionBuilderStore.js
```diff
@@ -208,9 +209,22 @@ export default class AdaTransactionBuilderStore extends Store {
...
+    this.plannedTx = null;
+    this.tentativeTx = null;
+  }
+
+  @action
+  _reset = async () => {
```
nicarq: Is _reset even used after the changes?

SebastienGllmt: Yes we have a listener on it that's called by the WalletSendPage

app/stores/ada/TrezorConnectStore.js
SebastienGllmt: To be honest it's not really good design that  we call an await inside setup but as I mentioned this should never cause  a problem in practice in store setup is called waaaay before it could ever be used by any real user. @sarveshwar-singh may have refactored this  whole  process in his work on WebAuthn so he may have to merge t his

app/stores/ada/AdaTransactionBuilderStore.js
nicarq: (Heads up: I’m working on my iPad so it’s tricky to review the whole files)

app/stores/ada/AdaTransactionBuilderStore.js
nicarq: Maybe we can tune up the names, it’s mostly that preWaitReset and postWaitReset are used inside Reset so the naming is a little bit tricky

nicarq: Maybe it could be something like: preWaitCreateUnsignedTx instead of preWaitReset

PR #992, opened 2019-10-17
Offline transaction support

SebastienGllmt:
This PR does a few things

1) Storage v2 stores all the UTXO locally. However, for creating transactions we still took the UTXO from the server. This PR now makes it take the UTXO locally (which enables features like creating a transaction offline)
2) I add some missing types to the v2 WASM bindings (to make it easier to upgrade to v3 later)
3) I remove the `IGetAllAddresses` and add its functionality to `IGetAllUtxos`. This is because now we know Chimeric wallets will be implemented using a chain derivation so it no longer makes sense to just ask for all addresses. You need to instead ask for all UTXO addresses and ask for all Account addresses separately. This means that our "balance tracker" feature would probably best be implemented as a new wallet type as opposed to a subset of Bip44 (which storage v2 still supports)

### DIFF
--- a/app/api/ada/lib/storage/models/PublicDeriver/index.js
+++ b/app/api/ada/lib/storage/models/PublicDeriver/index.js
@@ -15,7 +15,7 @@ import type {
   IPublicDeriver, IPublicDeriverConstructor,
   IAddFromPublic, IAddFromPublicRequest, IAddFromPublicResponse,
   IGetPublic, IGetPublicRequest, IGetPublicResponse,
-  IGetAllAddresses, IGetAllAddressesRequest, IGetAllAddressesResponse,
+  IGetAllUtxoAddressesRequest, IGetAllUtxoAddressesResponse,
   IGetAllUtxos, IGetAllUtxosRequest, IGetAllUtxosResponse,
   IDisplayCutoff,
   IDisplayCutoffPopRequest, IDisplayCutoffPopResponse,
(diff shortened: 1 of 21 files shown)


app/api/ada/lib/storage/models/PublicDeriver/index.js
```diff
@@ -899,6 +853,48 @@ const GetAllUtxosMixin = (
...
+  rawGetAllUtxoAddresses = async (
+    tx: lf$Transaction,
+    deps: {|
+      GetPathWithSpecific: Class<GetPathWithSpecific>,
+      GetAddress: Class<GetAddress>,
+      GetBip44DerivationSpecific: Class<GetBip44DerivationSpecific>,
+    |},
+    _body: IGetAllUtxoAddressesRequest,
+  ): Promise<IGetAllUtxoAddressesResponse> => {
+    // TODO: fix implementation for chain
```
SebastienGllmt: TODO

app/api/ada/lib/storage/models/PublicDeriver/index.js
```diff
@@ -780,80 +781,14 @@ export function asGetSigningKey<T: IPublicDeriver>(
...
 // ===============
 //   GetAllUtxos
 // ===============
 
-type GetAllUtxosDependencies = IPublicDeriver & IGetAllAddresses;
+type GetAllUtxosDependencies = IPublicDeriver & IBip44Parent;
 const GetAllUtxosMixin = (
   superclass: Class<GetAllUtxosDependencies>,
-) => class GetAllUtxos extends superclass implements IGetAllUtxos {
+) => class GetAllUtxos extends superclass implements IGetAllUtxos { // TODO: probably rename
```
SebastienGllmt: TODO. Not sure what's the best name for this now. Maybe just "HasUtxo"

PR #1021, opened 2019-11-01
Fix CI for Shelley branch

SebastienGllmt:
I’ve noticed all Yoroi-Frontend CI builds fail since we’ve moved to node12. After some investigation, this is because of the `npm ci` step since it starts by doing
```
node-sass@4.11.0 install /home/circleci/.npm/tmp/git-clone-545af135/node_modules/node-sass
node scripts/install.js
```

`node-sass@4.11.0` is the wrong version — that’s for node11. We need >4.12
I’ve bumped the version Yoroi uses to 4.13 already so it’s not an issue in our package. I have no clue why CI insists on using 4.11 (`npm run ci` works just fine locally)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,7 +110,7 @@ jobs:
 
       # cancel builds from the same branch (since circle feature seems not working)
       - run: bash .circleci/stop_redundant_jobs.sh
-      - *restore_cache
+      # - *restore_cache
       - run: bash .circleci/rust_install.sh
 
       - run: npm rebuild node-sass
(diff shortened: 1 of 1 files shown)


rcmorano:
It turns out that is a devDepend from `react-polymorph-fix` depend [[0]](https://github.com/Emurgo/yoroi-frontend/blob/fix-node-ci/package-lock.json#L19949) which is causing the problem because it's requiring `node-sass@4.11.0` [[1]](https://github.com/SebastienGllmt/react-polymorph-fix/blob/master/package.json#L75), so we are updating it in a different fork (to avoid possible breakage of other depending projects) and pointing `yoroi-frontend` to this new fork on `shelley` branch.

PR #1052, opened 2019-11-03
 shin/ch2070/fix-verify-restored-wallet-copy-tool-tip 

sarveshwar-singh:
Before:
image

After:
image

This PR increases width of **Verify Restored Wallet** dialog which is not the best solution.
Other solutions could be:
1. Making tool-tip overflows over the dialog: tried but seems not working in the first attempt
2. Multi-line tool-tip: needs many changes including translation. 
3. Something else

### DIFF
--- a/app/components/wallet/WalletRestoreVerifyDialog.scss
+++ b/app/components/wallet/WalletRestoreVerifyDialog.scss
@@ -56,6 +56,7 @@
 }
 
 :global(.YoroiModern) .dialog {
+  // Refer: https://github.com/Emurgo/yoroi-frontend/pull/1052
   min-width: 680px;
   max-width: 680px; 
 }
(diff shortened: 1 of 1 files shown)


Marta-Ryzhok:
image
 is it possible to move tooltip in this way?

SebastienGllmt:
@Marta-Ryzhok no this isn't possible without making a change to the react-polymorph library.

Marta-Ryzhok:
Ok, then if you said it's temporary then make it wider. 
Btw tooltips in Yoroi are below 
image
 but here is above..

sarveshwar-singh:
@Marta-Ryzhok 
Showing tool-tip up or down is possible but that will not going to fix the problem.
But showing the content more toward left or right is not possible ATM.

Marta-Ryzhok:
@sarveshwar-singh I see that showing below doesn't fix the current problem. I wrote it in order to keep it the same everywhere if we can.

PR #1068, opened 2019-11-06
 shin/ch2081/update-routing-to-load-staking-pages

sarveshwar-singh:


sarveshwar-singh:
Will be fixed in: https://github.com/Emurgo/yoroi-frontend/pull/1071

PR #1097, opened 2019-11-11
Show plate on transfer

SebastienGllmt:
verify-flow

### DIFF
--- a/app/containers/transfer/YoroiPlatePage.js
+++ b/app/containers/transfer/YoroiPlatePage.js
@@ -1,0 +1,90 @@
+// @flow
+import React, { Component } from 'react';
+import { observer } from 'mobx-react';
+import globalMessages from '../../i18n/global-messages';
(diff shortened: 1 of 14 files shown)


app/containers/transfer/YoroiPlatePage.js
```diff
@@ -0,0 +1,90 @@
...
+} from '../../../config/config-types';
+
+declare var CONFIG : ConfigType;
+const protocolMagic = CONFIG.network.protocolMagic;
+
+type Props = {|
+  ...InjectedProps,
+  +onNext: void => void,
+  +selectedExplorer: ExplorerType,
+  +onCancel: void => void,
```
SebastienGllmt: note: this is a `cancel` and not a back button so unfortunately if you realize you made a mistake in your mnemonic you have to re-enter it from scratch. It was easier to program this way but we can change it in the future if we want.

PR #1103, opened 2019-11-11
Add shelley language page

SebastienGllmt:
image

I used media queries to avoid the background image overlapping with the form

media2


app/i18n/locales/en-US.json
```diff
@@ -188,7 +191,6 @@
   "settings.support.reportProblem.title": "Reporting a problem",
   "testnet.label.message": "WARNING: This is a {network} network. ADA has no monetary value here. For more information, check out the FAQ at {faqLink}",
   "testnet.shelley.label.message": "YOU ARE ON TESTNET NETWORK ({network}).",
-  "transfer.form.errors.invalidRecoveryPhrase": "Invalid recovery phrase",
```
SebastienGllmt: this was removed because the exact same string was duplicated in `wallet.restore.dialog.form.errors.invalidRecoveryPhrase`

PR #1115, opened 2019-11-12
Update Yoroi Shelley Testnet page title from 'Yoroi' to 'Yoroi Shelley Testnet'

sarveshwar-singh:
Fixes:
https://app.clubhouse.io/emurgo/story/2176/update-yoroi-shelley-testnet-page-title-to-yoroi-shelley-testnet

This PR changes page title to:
image

It's using [react-helmet](https://www.npmjs.com/package/react-helmet).
Right now it using only one property but later we could add other meta tags or if we want of use `i18n` texts.

**Problem:**
(This is how [react-helmet](https://www.npmjs.com/package/react-helmet) works.)
When you click Yoroi icon for 1-2 seconds (Till react loads), the title 'Yoroi' is show then it changes to 'Yoroi Shelley Testnet' because title is set to 'Yoroi' in `main_window.html`.
If we delete title `Yoroi` in `main_window.html` for 1-2 it shows extension ID as title.


package.json
```diff
@@ -171,6 +171,7 @@
     "react-copy-to-clipboard": "5.0.2",
     "react-css-themr": "2.1.2",
     "react-dom": "16.11.0",
+    "react-helmet": "5.2.1",
```
SebastienGllmt: Note sure we need a new dependency for this.

If you instead modify `main_window.html` to use `<title><%= htmlWebpackPlugin.options.title %></title>` and then modify the `HtmlWebpackPlugin` for the `main_window` to use `title: 'foo bar',` you can set the title at compile time

That being said the way you do it is more flexible and probably easier to read. It just adds in a new dependency.

PR #1202, opened 2019-12-14
Update shelley-testnet to run on real Shelley genesis info

SebastienGllmt:


SebastienGllmt:
@nicarq @rcmorano we still need
1) Value for `backendUrl` (is `iohk-mainnet.yoroiwallet.com` still correct?)
2) Value for Seiza

PR #1217, opened 2019-12-17
adds chrome id

nicarq:
adds chrome id


app/containers/wallet/staking/StakingPage.js
```diff
@@ -38,7 +38,8 @@ export default class StakingPage extends Component<Props> {
         '$$BROWSER$$',
         environment.userAgentInfo.isFirefox
           ? 'firefox'
-          : 'chrome',
+            // $FlowFixMe
+          : 'chrome&chromeId=' + chrome.runtime.id,
```
SebastienGllmt: Will not work when running Yoroi as a website FYI
I recommend using `isExtension` from `userAgentInfo.js` and throwing an error if it's not an extension (we can fix it later)

PR #1351, opened 2020-02-02
Reward History in Dashboard

SebastienGllmt:
We now show the reward history for a wallet in the dashboard. There are a lot of cases for this flow so let's go through them.

# Scenario 1 - Normal case

image

# Scenario 2 - long name

Long pool names now get ellipsized to make room for the graph if they are too long

image

# Scenario 3 - separate loading

The reward history endpoint relies on a different endpoint from the one for metadata about the pool you are delegating to. This means they can finish loading at different times:

image

# Scenario 4 - separate failures

The reward history endpoint relies on a different endpoint from the one for metadata about the pool you are delegating to. This means they can fail independently

image

image

image

# Scenario 5 - Different epoch lengths

Epoch length in Cardano can change. In this example, based on the network, the epoch length is set to 1 day in the graph instead

image

# Scenario 6 - Wallet not delegating

If the user has never delegated before, there is no point is showing the graph so instead the informational message takes the full screen

image

If you've used the "undelegate" feature, we still want to show you the graph though so we show you a smaller informational message

image

# Scenario 7 - No rewards from pool yet

If you're delegating to a pool but haven't received any rewards yet, we just show an empty chart

image


Marta-Ryzhok:
@SebastienGllmt looks good to me! 
Few details:
- The title "rewards" should remain the same font as other titles 
image

- if the epoch length may change then I would highlight the number in brackets by making it bolder or put "info" icon once it changed so to drive attention on this
image

- is there a horizontal scroll when more epochs will be passed?

SebastienGllmt:
@Marta-Ryzhok the reason the title is different is because blue used to mean active tab. However in the first release we only have one tab which is why it looks strange.
I changed it to match for this release and we can change it back in the future
image

For epoch length, it's different per network (mainnet vs testnet). It doesn't change for a given network
for now (this may change in the future but I mean like far future not any time soon) so probably not worth bolding it.

The graph automatically resizes to fit the entire content. For example, here is 1000 epochs of fake data
image

PR #1475, opened 2020-04-20
Improve debugging of migration

SebastienGllmt:
I noticed our tests for upgrading IndexedDB don't work anymore. These tests were historically not as important (and was even removed and left unused until now) because for Byron all IndexedDB state was derived from the local storage so it wasn't strictly necessary. Now that this is no longer the case, it'd be nice to bring these tests back but they no longer work because it wasn't compatible with IndexedDB's internal migration system (DB versioning).

You can tell by running test @IT-85 and checking the console error and you'll see the IndexedDB error https://google.github.io/lovefield/error_lookup/src/error_lookup.html?c=111

We can try and improve this by injecting the IndexedDB state without having to open the extension but this may be trickier to implement and may be browser-specific. Will have to investigate.

### DIFF
--- a/app/api/ada/lib/storage/database/index.js
+++ b/app/api/ada/lib/storage/database/index.js
@@ -140,18 +140,13 @@ export async function importOldDb(
   // we need to delete the database before we import
   // because indexedDB uses schema versions
   // and you can't import an old schema.
-  console.log(1);
   await oldDb.delete();
-  console.log(2);
   const schemaBuilder = schema.create(data.name, data.version);
-  console.log(3);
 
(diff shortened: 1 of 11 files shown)


app/api/ada/lib/storage/database/index.js
```diff
@@ -118,6 +118,7 @@ async function onUpgrade(
   rawDb: lf$raw$BackStore,
 ): Promise<void> {
   const version = rawDb.getVersion();
+  console.log(`Starting IndexedDB migration for version ${version}`);
```
SebastienGllmt: Prefer to use logger if possible

PR #1477, opened 2020-04-21
Move to github actions

SebastienGllmt:
Trying to move from CircleCI to Github Actions.

This will be super tedious because Github Actions is still new enough that there is still no good way to lookup how to do things. Additionally, there is basically no way to debug them except to run it through Github.

Will try not to spend too much time on this since at this point I'm basically the only one running the CI and I often just run it locally instead since it's faster.


.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,41 @@
...
+
+jobs:
+  build:
+
+    strategy:
+      fail-fast: false # don't fail Chrome if Firefox fails, etc.
+      matrix:
+        include:
+          - platform: chrome
+            container: circleci/node:12-browsers
```
SebastienGllmt, line 22: Fails with `An image label with the label circleci/node:12-browsers does not exist.`

According to the docs

```
CircleCI provides a set of pre-built images with common dependencies. These images have the USER set to circleci, which causes permissions to conflict with GitHub Actions.

We recommend that you move away from CircleCI's pre-built images when you migrate to GitHub Actions. In many cases, you can use actions to install the additional dependencies you need.
```

rcmorano:
> This will be super tedious because Github Actions is still new enough that there is still no good way to lookup how to do things. Additionally, there is basically no way to debug them except to run it through Github.

https://github.com/nektos/act#alternative-runner-images

nicarq:
#1878 replaces this one

PR #1481, opened 2020-04-26
Add AdaStat Shelley Explorer

dmitrystas:


app/domain/Explorer.js
```diff
@@ -109,6 +110,12 @@ const getExplorerInfo: void => {
     }
     : {
       jormungandr: getIohkExplorer(),
+      adastat: {
+        name: 'AdaStat',
+        address: 'https://adastat.net/address/',
```
vantuz-subhuman, line 115: If I am not mistaken Yoroi will try to use this key for accounts as well, since accounts are a type of an address in Cardano, but AdaStat treats accounts and addresses separately, e.g.:
1. https://adastat.net/en/account/6be27b3870043e5b5c6563da66c47bdc1409c756da74e35ad7d3a2a034633b32
2. https://adastat.net/en/address/6be27b3870043e5b5c6563da66c47bdc1409c756da74e35ad7d3a2a034633b32

This will make all account links broken in Yoroi when AdaStat is selected. Will be great if there might be a detect-redirect in AdaStat, like:

```
when `/address/{x}`:
  if isAccount(x):
    redirect `/account/{x}`
  else:
    render `/address/{x}`
```

vantuz-subhuman: Or maybe the search URL could be used for the `address` key, e.g. `/search/{address}` or something, which could redirect right to the exact page when there's a single result.

image

vantuz-subhuman: Although it depends on whether AdaStat uses unwrapped account keys on the account page, maybe I am mistaken here and using properly wrapped account-address would work with `/address/` 🤔 Need to find an example of a real transaction with an account, like delegation or something

vantuz-subhuman: Ah, yes, unwrapped. All perfect then 👌👌

SebastienGllmt:
LGTM. Merged it from a different PR

PR #1517, opened 2020-05-16
adaex addon

cardanians:
added type "block" (to existing address, transaction, pool) - used in future maybe too?


SebastienGllmt:
Thanks! I included this change as part of this PR: https://github.com/Emurgo/yoroi-frontend/pull/1518

PR #1529, opened 2020-05-21
Implement new tx address display

paulclindo:
Two cases for handling addresses that do not belong to you:

1) When the address book is available in Yoroi
image

2) When the address book is not available in Yoroi yet
image


app/components/wallet/transactions/Transaction.js
```diff
@@ -83,6 +83,14 @@ const messages = defineMessages({
     id: 'wallet.transaction.address.to',
     defaultMessage: '!!!To address',
   },
+  addressType: {
+    id: 'wallet.transaction.address.type',
+    defaultMessage: '!!!Address Type',
+  },
+  amount: {
```
SebastienGllmt: Try and reuse existing messages instead of creating new ones. Otherwise, it ends up doubling the translation work. (see global-messages.js for a place where we usually dump these)

PR #1567, opened 2020-06-26
Material design hover state for receive page icons

paulclindo:
Story details: https://app.clubhouse.io/emurgo/story/5248
Screen Recording 2020-06-26 at 02 03 AM


SebastienGllmt (approved):

SebastienGllmt:
It's kind of weird that the on-hover for the clipboard icon works differently than all the other (it has no animation). Maybe we should make the clipboard icon also have an animation? @Marta-Ryzhok 

Similarly, the "Export to file" button on the transactions page also uses a different style.

Marta-Ryzhok:
@SebastienGllmt yeah, it would be good to unify and apply the same hover (with animation) on the clipboard icon.

The "export to file" icon has already circle, was done to make it more noticeable, but if you can apply the same animation on hover will be also great.

Can you add the same hover on the cross icon 
image

paulclindo:
@SebastienGllmt @Marta-Ryzhok  I've just unified the icon's styles (size and hover animation) using as an example the Copy Icon of the mockups. 

Screen Shot 2020-06-29 at 1 21 32 AM

 -------------------------
Not sure about this one, the background is the same as the hover color, should we make more noticeable? I think it would be enough by showing the tooltip on hover state or should we change the hover color? 

Screen Shot 2020-06-29 at 1 23 48 AM

SebastienGllmt:
One option is to add make the address box look more like the transaction page (light gray background with box shadow)
Thoughts? @Marta-Ryzhok  image

Marta-Ryzhok:
@paulclindo @SebastienGllmt 
I'd leave as it is, I also think that tooltip is enough

PR #1568, opened 2020-06-26
Transaction page looks weird on small width monitors

paulclindo:
Story details: https://app.clubhouse.io/emurgo/story/5253


SebastienGllmt:
Hmm it's slightly better but I don't think this is enough, Maybe we should use a media query to hide the address type field when the element is too small? Probably these two changes together would be enough

paulclindo:
@SebastienGllmt What do you think about making the 2 columns (From/To) into one? So we won't hide any feature, but if the address-type is not that really important, I will hide it. 
Screen Shot 2020-06-28 at 4 15 23 PM

PR #1569, opened 2020-06-26
Receive page address types has no scrollbar

paulclindo:
Story details: https://app.clubhouse.io/emurgo/story/5251
Before: 
Screen Shot 2020-06-26 at 3 05 26 AM

After : 

Screen Shot 2020-06-26 at 2 54 32 AM

### DIFF
--- a/app/components/wallet/navigation/ReceiveNavigation.js
+++ b/app/components/wallet/navigation/ReceiveNavigation.js
@@ -1,6 +1,6 @@
 // @flow
 import React, { Component } from 'react';
-import type { Node } from 'react';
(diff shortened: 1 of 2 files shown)


SebastienGllmt:
This fix causes an issue where the info button doesn't follow the category when you scroll
image

Maybe there isn't a way to fix this other than keeping track of the scroll position through a Javascript listener and re-aligning the info buttons on the scroll event

SebastienGllmt:
Probably also best to set a min-height for the accordions section (maybe 40px -- the size of one element) to avoid it being shrunk to 0px size.

PR #1642, opened 2020-08-06
Add Adapools Support

nicarq:
Website-side pending tasks:
- [ ] No feedback about what Stakepool has already being selected
- [ ] fix links when clicking pool name
- [ ] definition for colors
- [ ] Pledge of less than 1000 ADA shows up in lovelaces. Ex: 130 ADA is `130000000`. This makes it look like 130M pledge

Yoroi-side pending tasks:

- [ ] cookie SameSite attribute (investigation result: these are set by Cloudflare -- don't think we can fix it for them)
- [ ] investigate weird wasm-eval error message (needs extra work on the website side before we can fix this)
- [x] Integration with AdaPools
- [x] Code for names updated
- [x] Logic to update names in UI
- [x] Adapools support for multilanguage
- [x] Update URL Template to Yoroi Proxy (requires yoroi proxy to be updated)
- [x] iframe sandbox
- [x] fix search box
- [x] fix popup when clicking on a pool

Screen Shot 2020-08-05 at 11 13 20 PM

Screen Shot 2020-08-05 at 11 13 10 PM


app/containers/wallet/staking/CardanoStakingPage.js
```diff
@@ -318,6 +346,7 @@ export default class CardanoStakingPage extends Component<Props> {
   }
 
   @computed get generated(): {|
+    SeizaFetcherProps: InjectedOrGenerated<SeizaFetcherData>,
```
nicarq: ah I missed this. I will delete it

app/containers/wallet/staking/CardanoStakingPage.js
```diff
@@ -459,6 +488,9 @@ export default class CardanoStakingPage extends Component<Props> {
           getTooltipActiveNotification: stores.uiNotifications.getTooltipActiveNotification,
         },
       },
+      SeizaFetcherProps: (
```
nicarq: will delete this too

SebastienGllmt (approved):

PR #1646, opened 2020-08-08
Small fixes

SebastienGllmt:
I think these are the last fixes needed to push to nightly. Unfortunately couldn't get it done before I fall asleep on my desk so I'll finish it in the morning

Still TODO:
- [x] bug in delegation page means the page needs to be refreshed if you cancel the delegation


app/api/ada/lib/storage/bridge/updateTransactions.js
```diff
@@ -317,6 +317,45 @@ export async function rawGetTransactions(
...
+        if (tx.txType === TransactionType.CardanoShelley) {
+          let implicitOutputSum = new BigNumber(0);
+          for (const cert of tx.certificates) {
+            if (
+              cert.certificate.Kind !==
+              RustModule.WalletV4.CertificateKind.MoveInstantaneousRewardsCert
+            ) {
+              continue;
+            }
+            // TODO: uncomment once this works in the WASM side
```
SebastienGllmt, line 330: This is unfortunate we can't have this, but we can live without it until Rob can look into it

nicarq: what this does?

PR #1672, opened 2020-08-31
Claim rewards for users who forgot their ITN mnemonic

SebastienGllmt:
Some users forgot their ITN recovery phrase, so I'm adding a new "claim" option for these users

It allows them to input an encrypted or unencrypted key to claim the rewards from there.

Which cases do we need any why?

## 1) Unencrypted staking key without chaincode

There are two groups that this covers:

A) Pool operators (I believe) probably earn their rewards on an unencrypted private key (64-bytes), so we might as well allow them to claim their rewards in Yoroi
B) Yoroi ITN Chrome users will be able to use an update to export their unencrypted staking key w/o chain code to them import it into Yoroi

## 2) Encrypted staking key with chaincode

Unfortunately it's impossible for us to update the Firefox extension. However, you can get your encrypted private key from IndexedDB directly. In this case, we need to take the encrypted key, decrypt it and claim the rewards that way


TheMooWalker:
This is great news. Thank you Sebastien for working on this.

TheMooWalker:
One step closer. Looking forward to this. Thanks again Seb

PR #1798, opened 2020-11-04
draft 1

ashisherc:


app/stores/toplevel/TransactionsStore.js
```diff
@@ -413,6 +414,7 @@ export default class TransactionsStore extends Store {
   @action _exportTransactionsToFile: {|
     publicDeriver: PublicDeriver<>,
     exportRequest: TransactionRowsToExportRequest,
+    delegationRequests: DelegationRequests,
```
SebastienGllmt, line 417: Instead of passing `delegationRequests` into this function, I would just put the call to `this.stores.delegation.getDelegationRequests(publicDeriver)` into this function itself

ashisherc: gotcha

PR #1803, opened 2020-11-09
Export reward draft2

ashisherc:


package.json
```diff
@@ -110,10 +110,12 @@
     "cucumber": "git+https://github.com/SebastienGllmt/cucumber-js.git",
     "eslint": "7.10.0",
     "eslint-config-airbnb": "18.2.0",
+    "eslint-config-prettier": "^6.15.0",
```
SebastienGllmt, line 113: Historically we prefer to use strict versions (remove the `^`) and manually validate any version upgrade.

app/containers/wallet/WalletSummaryPage.js
```diff
@@ -187,6 +190,35 @@ export default class WalletSummaryPage extends Component<InjectedOrGenerated<Gen
 
     const notification = this._getThisPageActiveNotification();
 
+    let exportDialog;
+
+    const delegationStore = this.generated.stores.delegation;
+    const delegationRequests = delegationStore.getDelegationRequests(publicDeriver);
+
+    if(!this.readyToExportHistory({ delegationRequests, publicDeriver })){
```
SebastienGllmt, line 198: Usually we have spaces between ifs and braces
ex
```
if (foo) {
} else {
}
```

I think it should be a lint warning

SebastienGllmt (approved):

app/containers/wallet/WalletSummaryPage.js
ashisherc: weird that eslint isn't complaining about it, will check that out

package.json
ashisherc: Sure

PR #1808, opened 2020-11-18
redesign user summary and epoch section

ashisherc:
Screenshot 2020-11-23 at 8 36 10 PM

Screenshot 2020-11-27 at 5 06 33 PM
Screenshot 2020-11-27 at 5 41 53 PM
Screenshot 2020-11-27 at 5 44 09 PM
Screenshot 2020-11-27 at 5 46 56 PM
Screenshot 2020-11-27 at 12 15 00 PM
Screenshot 2020-11-27 at 6 09 16 PM
Screenshot 2020-11-27 at 6 09 24 PM


SebastienGllmt (approved):
Looks good (pending the fix on the wording we discussed on Slack)

app/components/wallet/staking/dashboard/UserSummary.js
```diff
@@ -107,126 +117,138 @@ export default class UserSummary extends Component<Props, State> {
...
+            <div className={styles.sectionActions}>
+              {this.props.withdrawRewards != null && (
+                <Button
+                  className='secondary'
+                  label={intl.formatMessage(globalMessages.withdrawLabel)}
+                  onClick={this.props.withdrawRewards}
+                  skin={ButtonSkin}
+                />
+              )}
+              {/* <div
```
SebastienGllmt: We're talking about re-adding this somehow in Slack. We can re-add it in a separate PR if you want though

ashisherc: done

ashisherc:
> Looks good (pending the fix on the wording we discussed on Slack)

done

Marta-Ryzhok:
@ashisherc 
Hey, looks good overall! 

Few minor UI remarks
- reduce text line-height to 20 here and move text down: 
image
- please reduce the height of the button "Withdraw"  to 46px (the same will be for button "Undelegate") so the general height of cards is less
- can you please position the link "less than expected" to the right side, so again to reduce the general height of cards. 
Only when the screen width is resizing the link will jump above the button. 
image
- Increase padding between tabs and section "Your summary" starting from the title. It should be 40 px
image

Marta-Ryzhok:
Will you add "potential rewards" inside of "upcoming rewards" cards?
Those cards that are grey doesn't have rewards, but the one which is white with shadow (like all other cards) has a potential reward.
image

Marta-Ryzhok:
And the pagination below, we don't need it now, can you please hide it? we might want to add it later.

ashisherc:
> Will you add "potential rewards" inside of "upcoming rewards" cards?
> Those cards that are grey doesn't have rewards, but the one which is white with shadow (like all other cards) has a potential reward.
> image

Potential rewards will be supported in upcoming release

SebastienGllmt:
Maybe this should be vertically aligned?
image

Are we okay with the spinners being like this? Probably they should be aligned with where the text appears
image

If the name doesn't fit, we should probably only display the ticker instead of wrapping lines
image

Is the "epoch" text supposed to be right-aligned when the info button appears and left-aligned otherwise?
image

When the pool name isn't shown, we show the pool hash instead. However, this breaks the formatting
image

Are we okay with the warning icon appearing at the bottom-right of the dialog instead of the top-right like it used to?
image

Marta-Ryzhok:
@SebastienGllmt 
1. hmm, why did we remove dates when epochs end? I'd keep it.
Is it the design for epochs when the user didn't delegate to any Stake pool, right?

Can we keep the titles and do not show stake pools, rewards, so it's clear that the user didn't delegate in the specific epoch and that's why won't receive a reward. Also, there is might be a case when the user is still going to receive a reward from the current epoch, but not delegating in the upcoming epochs.
Like this:
image

2. Yes, better align spinner to where the value will appear, and I'll show all other elements, so it highlights that only ADA value is loading, like this:
image

3. Yeah can work. Also, if the name doesn't fit usually it shows like this **[2EMUR] EMURGO'...** and the more screen is resized the less name we show. The full name is to be shown in the tooltip anyway.

4. No. Why the Epoch title jumps to the right because of the info icon? Should be aligned to the left as all content in the card. The info icon should be dependent on the title.

5. Let's just shortcut the hash as we do on the Delegation list page  **7186b1...83** 
Again, we show full id and name in the tooltip anyway
e.g
image

and let's remove the word "current" next to the first epoch when the screen is resized, at least we keep titles of similar length. When it resizes even to smaller width, time will jump below, which is ok.

Marta-Ryzhok:
About the warning icon, I'm not very convinced to put it over the background icon. We may put it next to the title, similarly as we do with the "info" icons
image

app/components/widgets/CustomTooltip.scss
```diff
@@ -13,6 +13,18 @@
       margin-left: 20px; // offset = horizontal padding of icon and tooltip
     }
   }
+
+  .PoolAvatarTooltip {
```
SebastienGllmt, line 17: Usually we deal with this with `:global` overrides inside the files that need it, but this is also fine I think (since the overrides also tend to be messier to work with, I wouldn't really call them a better solution in any way)

app/components/wallet/transactions/Transaction.js
```diff
@@ -184,7 +184,6 @@ type Props = {|
   +state: TxStatusCodesType,
   +selectedExplorer: SelectedExplorer,
   +assuranceLevel: AssuranceLevel,
-  +isLastInList: boolean,
```
SebastienGllmt, line 187: Why was this removed?

SebastienGllmt (approved):

app/components/widgets/CustomTooltip.scss
ashisherc: I was thinking to have more options this way, instead of having `isAligningRight` or `SimpleCenteredTooltip` as fixed conditional class

app/components/wallet/transactions/Transaction.js
ashisherc: reverted back - looks like it was a mistake

PR #1831, opened 2020-12-01
support shelley base address for payment URLs and pick first Cardano …

ashisherc:
…wallet

### DIFF
--- a/app/containers/uri/URILandingPage.js
+++ b/app/containers/uri/URILandingPage.js
@@ -9,6 +9,7 @@ import URILandingDialogContainer from './URILandingDialogContainer';
 import type { GeneratedData as URILandingDialogContainerData } from './URILandingDialogContainer';
 import type { UriParams } from '../../utils/URIHandling';
 import { PublicDeriver } from '../../api/ada/lib/storage/models/PublicDeriver/index';
+import { isCardanoHaskell } from '../../api/ada/lib/storage/database/prepackaged/networks';
 
 export type GeneratedData = typeof URILandingPage.prototype.generated;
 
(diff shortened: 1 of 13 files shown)


app/containers/uri/URILandingPage.js
```diff
@@ -23,10 +24,17 @@ export default class URILandingPage extends Component<InjectedOrGenerated<Genera
 
   onConfirm: void => void = () => {
     const { wallets } = this.generated.stores;
+
+    const firstShelley = wallets.allWallets.find(publicDeriver => {
```
nicarq: @ashisherc I would update this to use a more descriptive name `firstShelley` is strange

app/stores/toplevel/WalletStore.js
```diff
@@ -221,10 +221,8 @@ export default class WalletStore extends Store {
     );
   }
 
-  @computed get first(): null | PublicDeriver<> {
-    return this.publicDerivers.length === 0
-      ? null
-      : this.publicDerivers[0];
+  @computed get allWallets(): Array<PublicDeriver<>> {
```
SebastienGllmt: We don't need a computed function for this. You can just use `publicDerivers` directly (it's a observable property so it will handle updates. Computed functions are useful when you need to do some calculation involving observable variables to get a final result)

app/stores/toplevel/LoadingStore.js
```diff
@@ -134,7 +134,9 @@ export default class LoadingStore extends Store {
         address => {
           // TODO: validation should be done based on wallet type
           const addressKind = tryAddressToKind(address, 'bech32', networks.CardanoMainnet);
-          const valid = addressKind != null && addressKind === CoreAddressTypes.CARDANO_LEGACY;
+          const valid = addressKind != null &&
+          ( addressKind === CoreAddressTypes.CARDANO_LEGACY ||
+            addressKind === CoreAddressTypes.CARDANO_BASE);
```
SebastienGllmt: Instead of only hard-coding base addresses, it probably makes more sense to use `isValidReceiveAddress`

app/containers/uri/URILandingPage.js
ashisherc: changed to `firstCardanoWallet`

PR #1834, opened 2020-12-04
add more explicit warning for claim and design update

ashisherc:
Screenshot 2020-12-04 at 6 33 10 PM

### DIFF
--- a/app/components/transfer/cards/TransferTypeSelect.js
+++ b/app/components/transfer/cards/TransferTypeSelect.js
@@ -47,7 +47,7 @@ export default class TransferTypeSelect extends Component<Props> {
                 messages.subInstruction,
                 { ticker: this.props.ticker }
               )}
-              <a href="https://yoroi-wallet.com/#/faq"> FAQ</a>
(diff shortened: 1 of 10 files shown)


app/components/transfer/cards/TransferTypeSelect.js
```diff
@@ -31,16 +35,25 @@ export default class TransferTypeSelect extends Component<Props> {
...
+                messages.instruction,
+                { ticker: this.props.ticker }
+              )}
+            </div>
+            <span>
+              {intl.formatMessage(
+                messages.subInstruction,
+                { ticker: this.props.ticker }
+              )}
+              <a target="_blank" rel="noreferrer" href="https://yoroi-wallet.com/#/faq/1"> FAQ</a>
```
SebastienGllmt: Does this actually work? I don't remember why, but we used to have issues with adding `<a>` tags and so we had to use `handleExternalLinkClick` instead

rahman677:
I forgot my phars recovery.
Pleas help

SebastienGllmt:
@rahman677 please contact Yoroi support instead of posting on Github
https://yoroi-wallet.com/#/support

PR #1838, opened 2020-12-13
catalyst QR code

ashisherc:


package.json
```diff
@@ -167,6 +167,7 @@
     "bs58": "4.0.1",
     "bs58check": "2.1.2",
     "cardano-wallet-browser": "1.2.2",
+    "chacha": "^2.1.0",
```
SebastienGllmt: Prefer strict versions (`2.1.0` instead of `^2.1.0`)

app/actions/ada/voting-actions.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+import { AsyncAction, Action } from '../lib/Action';
+import { PublicDeriver } from '../../api/ada/lib/storage/models/PublicDeriver/index';
+
+export default class VotingActions {
+  generateCatalystKey: AsyncAction<void> = new AsyncAction();
+  createTransaction: AsyncAction<string> = new AsyncAction();
+  signTransaction: AsyncAction<{|
```
SebastienGllmt, line 8: Do we need a special sign transaction for this?

app/api/ada/index.js
```diff
@@ -927,6 +947,10 @@ export default class AdaApi {
       };
 
       let unsignedTxResponse;
+      let trxMetadata = undefined;
```
SebastienGllmt: I think you can just use a ternary here so that you can make `trxMetadata` const instead

app/api/ada/lib/cardanoCrypto/cryptoWallet.js
```diff
@@ -100,3 +100,19 @@ export function getCryptoDaedalusWalletFromMasterKey(
   const wallet = RustModule.WalletV2.DaedalusWallet.new(privateKey);
   return wallet;
 }
+
+/**
+ * Generate catalyst private key for QR code
+ */
+
+export function generatePrivateKeyForCatalyst(): RustModule.WalletV4.Bip32PrivateKey {
```
SebastienGllmt, line 108: I think this is fine for the first version to get things working, but I would like to make a modification to CIP1852 to include a voting key path instead of generating random keys like this every time going forward.

app/api/ada/lib/storage/bridge/metadataUtils.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+import { RustModule } from '../../cardanoCrypto/rustLoader';
+
+export type TransactionMetadata = {|
+  label: string,
+  data: Object,
```
SebastienGllmt: just FYI that `Object` is aliased to `any`. If you want to force it to be a JS object, you can use `{...}` instead

app/api/common/lib/transactions/ISignRequest.js
```diff
@@ -9,6 +10,6 @@ export interface ISignRequest<T> {
   uniqueSenderAddresses(): Array<string>;
   receivers(includeChange: boolean): Array<string>;
   isEqual(tx: ?mixed): boolean;
-
+  txMetadata(): RustModule.WalletV4.TransactionMetadata | void;
```
SebastienGllmt: probably don't need to make this part of the `ISignRequest` interface and instead only make it part of the Shelley sign request. Similar to withdrawals, etc. we can just check the type using `instanceof` then case when needed.

app/stores/ada/VotingStore.js
```diff
@@ -0,0 +1,238 @@
...
+        }).slot
+      );
+
+      const catalystPubKey = Buffer.from(this.catalystPrivateKey.to_public().as_bytes())
+        .toString('hex');
+      const catalystSignature = stakingKey
+        .sign(this.catalystPrivateKey.to_public().as_bytes())
+        .to_hex();
+
+      const trxMeta = [
```
SebastienGllmt, line 149: maybe a comment to describe this (unfortunately no CIP for this yet though)

app/stores/ada/VotingStore.js
```diff
@@ -0,0 +1,238 @@
...
+      this.markStale(false);
+    } else {
+      throw new Error(
+        `${nameof(VotingStore)}::${nameof(this._createTransaction)} network not supported`
+      );
+    }
+  };
+
+  @action
+  _signTransaction: ({|
```
SebastienGllmt, line 181: Do we need to copy-paste this function?

app/stores/ada/VotingStore.js
```diff
@@ -0,0 +1,238 @@
...
+      this.encryptedKey = key;
+      this.pin = passwordArray;
+      this.catalystPrivateKey = RustModule.WalletV4.PrivateKey.from_extended_bytes(
+        rootKey.to_raw_key().as_bytes()
+      );
+    });
+  };
+
+  @action.bound
+  reset(request: {| justTransaction: boolean |}): void {
```
SebastienGllmt, line 248: make sure things like `catalystPrivateKey` are properly cleared from memory once no longer needed.

SebastienGllmt (approved):

PR #1870, opened 2021-01-15
Merge ITN ToS Stuff

nicarq:
Merge ITN ToS stuff


app/i18n/locales/terms-of-use/itn/cs-CZ.md
```diff
@@ -1,9 +0,0 @@
-## 12. Delegation and Staking
```
SebastienGllmt, line 1: I think these should also be automatically updated by CrowdIn so modifying these files directly will cause merge conflicts

nicarq: @SebastienGllmt yeah I need to change the configuration from Crowdin first -- although stop tracking the changes on those files won't delete them afaik

PR #1872, opened 2021-01-16
fix ToS

nicarq:


ashisherc:
It breaks the terms page in Yoroi, I think #1868 should fix that

ashisherc (approved):

nicarq:
@ashisherc double checked now

Screen Shot 2021-01-16 at 10 58 44 AM
Screen Shot 2021-01-16 at 10 58 51 AM
Screen Shot 2021-01-16 at 10 58 27 AM

PR #1893, opened 2021-01-26
show metadata for transaction

ashisherc:


app/components/wallet/transactions/Transaction.js
```diff
@@ -817,4 +824,33 @@ export default class Transaction extends Component<Props, State> {
...
+  getMetadata: WalletTransaction => ? Node = (data) => {
+    const { intl } = this.context;
+
+    if (data instanceof CardanoShelleyTransaction && data.metadata !== null) {
+      let jsonData = null;
+
+      try {
+        jsonData = parseMetadata(data.metadata);
+      } catch (error) {
+        // discard error
```
SebastienGllmt: maybe if `BasicConversions` fails we should try `DetailedSchema` as a fallback? I think `DetailedSchema` supports all possible metadata but I need to double-check with Rob

ashisherc: `DetailedSchema` can possibly parse any type of metadata into JSON, but very verbose, eg.

```
{"map":[
  {
    "k":{"bytes":"8badf00d"},
    "v":{"bytes":"deadbeef"}
  },
  {
    "k":{"int":9},
    "v":{"int":5}
  },
  {
    "k":{"string":"obj"},
    "v":{"map":[
      {
        "k":{"string":"a"},
        "v":{"list":[
          {"map":[
            {
              "k":{"int":5},
              "v":{"int":2}
            }
          ]},
          {"map":[
          ]}
        ]}
      }
    ]}
  }
]}
```
To make it a bit complicated, should we do this for only Advanced User?

SebastienGllmt: According to Rob (about the DetailedSchema)
> It will fail decoding when you have a negative number between -2^64  + 1  and -2^63 since the json libraries use i64 for negative numbers but the metadata represents signed numbers as either a u64 or a different u64 marked as negative. (int = uint / nint in CDDL).

SebastienGllmt: I think due to the complexity of understanding the default format, only falling back to it for advanced users makes sense 👍

SebastienGllmt (approved):

PR #1897, opened 2021-01-29
make dashboard first tab, fix route crash and fixes

ashisherc:


app/containers/wallet/staking/StakingDashboardPage.js
```diff
@@ -1062,19 +1054,16 @@ export default class StakingDashboardPage extends Component<Props> {
...
-        };
-      }
       if (api === ApiOptions.jormungandr) {
         return {
           getTimeCalcRequests: stores.substores.jormungandr.time.getTimeCalcRequests,
           getCurrentTimeRequests: stores.substores.jormungandr.time.getCurrentTimeRequests,
         };
       }
-      throw new Error(`${nameof(EpochProgressContainer)} api not supported`);
+      return {
```
SebastienGllmt, line 1069: This definitely doesn't look right. Why would you return timestores for ADA for example when you have Ergo selected? I get that this page should never be loaded when Ergo is used, but still this feels like a hacky solution. Maybe it makes more sense to return undefined here and then in `componentDidMount` do nothing if the result is undefined

ashisherc: fixed as we discussed

app/stores/stateless/topbarCategories.js
```diff
@@ -110,13 +110,13 @@ export const SEIZA_STAKE_SIMULATOR: TopbarCategory = registerCategory({
...
-  className: 'cardanoStake',
-  route: ROUTES.WALLETS.CARDANO_DELEGATION,
-  icon: undefined,
-  label: messages.delegationById,
-  isVisible: request => (
-    asGetStakingKey(request.selected) != null &&
-    isCardanoHaskell(request.selected.getParent().getNetworkInfo())
-  ),
-});
+// export const CARDANO_DELEGATION: TopbarCategory = registerCategory({
```
nicarq: I would delete something rather than comment it out -- if we want to re-use it for later I would leave a comment on why this is commented

ashisherc: removed it in this case 👍

ashisherc:
NOTE: This removes the delegation by id tab from Yoroi, and currently there is no other way to access this functionality.

PR #1899, opened 2021-02-01
Ergo connector UI

paulclindo:
TODOs:

- [x] Connect Page w/ functionality
- [x] Connected Websites Page  w/ functionality
- [x] Menu 
- [x]  Sign in Tx Page w/ functionality
- [x] Add Language Selector in Setting Page 
- [x] Refactor the app store, actions, flow types (since I've been reusing the same as Yoroi and we don't need the whole data stored, only for i18n, theme, etc)

Demo Connect, Connected Websites, Signing Tx Confirmation: 
https://share.getcloudapp.com/yAu62lyB

Demo Language Selector:
Screen Recording 2021-02-08 at 12 42 29 AM

Added Read-only info text and No websites connected page 

Screen Shot 2021-02-10 at 9 22 36 PM
Screen Shot 2021-02-10 at 9 20 45 PM
Screen Shot 2021-02-10 at 9 21 21 PM


chrome/extension/ergo-connector/connect.js
```diff
@@ -45,7 +45,8 @@ chrome.runtime.sendMessage({ type: 'connect_retrieve_data' }, async response =>
     connect.onclick = () => {
       const accountsSelected: HTMLCollection<HTMLInputElement> = (document.getElementsByName('account'): any);
       for (let i = 0; i < accountsSelected.length; i += 1) {
-        if (accountsSelected[i].checked === true) {
```
rooooooooob, line 48: Is this file even still being used at all with the new UI?

nicarq: cc @paulclindo

paulclindo: @rooooooooob Good catch! I've just removed the HTML and JS files since we don't need them in the new UI

PR #1929, opened 2021-02-18
Update to webpack v5

SebastienGllmt:
Helps with build speedups and better support for WASM


webpack/commonConfig.js
```diff
@@ -204,7 +218,7 @@ module.exports = {
   plugins,
   rules,
   optimization,
-  node,
   resolve,
   definePlugin,
+  experiments: { syncWebAssembly: true }
```
SebastienGllmt, line 223: I didn't enable async WASM for this initial version bump

PR #1934, opened 2021-02-21
Buy / Sell ADA in Yoroi

nicarq:
Screen Shot 2021-02-20 at 10 25 53 PM

Screen Shot 2021-02-20 at 10 25 44 PM

Screen Shot 2021-02-20 at 10 26 03 PM


app/containers/wallet/MyWalletsPage.js
```diff
@@ -120,10 +150,43 @@ export default class MyWalletsPage extends Component<Props> {
...
+        const settingsCache: ConceptualWalletSettingsCache = this.generated.stores.walletSettings
+          .getConceptualWalletSettingsCache(parent);
+
+        // Currency Name
+        const defaultToken = this.generated.stores.tokenInfoStore.getDefaultTokenInfo(
+          wallet.getParent().getNetworkInfo().NetworkId
+        )
+
+        const currencyName = getTokenName(defaultToken)
+        if (currencyName !== 'ADA') return Promise.resolve(null);
```
SebastienGllmt: What are you trying to check in this case?
If you want to only allow this for the Cardano mainnet, it's probably better to compare the `NetworkId` against `CardanoMainnet` instead of the currency name

app/components/buySell/ChangellyFetcher.js
```diff
@@ -0,0 +1,99 @@
...
+    return (
+      <iframe
+        /**
+          * Rationale for allowing the following:
+          *
+          * allow-scripts:
+          * The iframe uses javascript for:
+          *   - Integrate Changelly Widget
+          *
+          * allow-popups:
```
SebastienGllmt, line 63: Do we need `allow-popups`? Looks like the rationale wasn't updated

app/components/buySell/BuySellDialog.js
```diff
@@ -0,0 +1,166 @@
...
+        .then((resp) => {
+          const wallets = [
+            ...resp,
+            {
+              walletName: intl.formatMessage(messages.dialogManual),
+              currencyName: '',
+              anAddressFormatted: '',
+            }
+          ]
+          this.setState({ walletList: wallets })
```
SebastienGllmt: Feels strange we're updating the state based on the props in the render function -- even if it is in some other async context. 
We may want to use [a different lifecycle function](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) for this

app/components/buySell/ChangellyFetcher.js
nicarq: Yeah I tried removing it but for some flows it opens up a new website to continue the flow and as expected it doesn't work without it

app/containers/wallet/MyWalletsPage.js
nicarq: Will update

app/components/buySell/ChangellyFetcher.js
SebastienGllmt: Okay, maybe just update the comment to mention this then?

app/components/buySell/BuySellDialog.js
nicarq: (conversation in Slack)

SebastienGllmt (approved):

PR #1936, opened 2021-02-22
Finish changelly

SebastienGllmt:
Found a few problems while testing the integration

TODO:
- [x] Finish Storybook integration
- [x] Come up with an E2E test for this
- [ ] Finish testing integration


chrome/content-scripts/3rd-party-trezor/trezor-usb-permissions.js
```diff
@@ -10,7 +10,7 @@ declare var chrome;
 const VERSION = '8.1.22';
 const versionN = VERSION.split('.').map(s => parseInt(s));
 const DIRECTORY = `${ versionN[0] }${ (versionN[1] > 0 ? `.${versionN[1]}` : '') }/`;
-const url = 'https://connect.trezor.io/' + 8;
```
SebastienGllmt, line 13: This is unrelated, but I think we were missing a slash here

package.json
```diff
@@ -203,7 +203,7 @@
     "react-dom": "16.13.1",
     "react-intl": "2.9.0",
     "react-markdown": "5.0.3",
-    "react-polymorph": "0.9.8-rc.6",
```
SebastienGllmt, line 206: Unrelated, but it fixes an issue with NumericInput

app/components/buySell/BuySellDialog.js
```diff
@@ -152,7 +158,6 @@ export default class BuySellDialog extends Component<Props, State> {
           <div className={styles.component}>
             <div className={styles.description}>
               {intl.formatMessage(messages.dialogDescription)}
-              {this.props.walletList}
```
SebastienGllmt, line 155: @nicarq I delete this since it didn't make sense (`this.props.walletList` here is an async function) and it didn't seem necessary for the UI to work properly

nicarq: yeah, I missed it. We definitely don't need this here

nicarq: thanks Seba!

PR #1938, opened 2021-02-23
Add CIP14 and assetName parsing

SebastienGllmt:
image

image

It tries to parse assetName as ASCII and, if it fails, it uses our new CIP14 library as a fallback.


Cryptoking74:
<signing_in>CryptoKing💫

PR #1942, opened 2021-02-26
Update explorer links

ashisherc:
Screenshot 2021-02-27 at 1 11 55 AM

- Makes only the token name part clickable to visit the explorer for token details. When the amount is bigger, this would be neat to click.
- ADA and Ergo tokens will not have the link

### DIFF
--- a/app/api/ada/lib/storage/database/prepackaged/explorers.js
+++ b/app/api/ada/lib/storage/database/prepackaged/explorers.js
@@ -7,6 +7,40 @@ const CardanoMainnetExplorers: Array<$ReadOnly<ExplorerRow>> = [
   {
     ExplorerId: 1_00,
     NetworkId: networks.CardanoMainnet.NetworkId,
+    IsBackup: true,
+    Endpoints: {
+      address: 'https://cardanoscan.io/address/',
+      transaction: 'https://cardanoscan.io/transaction/',
+      pool: 'https://cardanoscan.io/pool/',
+      stakeAddress: 'https://cardanoscan.io/stakeKey/',
+      token: 'https://cardanoscan.io/token/',
+    },
+    Name: 'CardanoScan',
+  },
+  {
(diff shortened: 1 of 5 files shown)


app/api/ada/lib/storage/database/prepackaged/explorers.js
```diff
@@ -7,51 +7,57 @@ const CardanoMainnetExplorers: Array<$ReadOnly<ExplorerRow>> = [
...
     },
-    Name: 'Clio.1',
+    Name: 'ADAex.org',
   },
   {
     ExplorerId: 1_04,
     NetworkId: networks.CardanoMainnet.NetworkId,
     IsBackup: false,
     Endpoints: {
-      address: 'https://explorer.cardano.org/en/address?address=',
```
SebastienGllmt, line 51: Please don't  change it so that ExplorerIds now point to different explorers 🙏 
Better to just leave gaps and add to the list if needed.

app/components/transfer/TransferSummaryPage.js
```diff
@@ -176,7 +176,7 @@ export default class TransferSummaryPage extends Component<Props> {
                       selectedExplorer={this.props.selectedExplorer}
                       light
                       hash={this.props.addressToDisplayString(deregistration.rewardAddress)}
-                      linkType="address"
+                      linkType="stakeAddress"
```
SebastienGllmt, line 179: should we maybe have it default to `address` if `stakeAddress` isn't set for an explorer

SebastienGllmt: Hmm thinking more about this, I'm not sure if we should have this kind of fallback. Instead, @ashisherc can you please add the `stakeAddress` for the other explorers? CardanoScan can't be the only one that supports stake addresses

SebastienGllmt (approved):

PR #1946, opened 2021-03-01
Handled UTXO sums over u64

SebastienGllmt:
This PR improves handling of sending when some token balances go over u64 (max allowed for a UTXO by the protocol)


v-almonacid (approved):
LGTM

PR #1962, opened 2021-03-03
Connector: Whitelist storage fix

rooooooooob:
Via the Local Storage api we are storing as strings and doing json
parse/stringify to get to from that to WhitelistEntry, however we access
the whitelist in 2 areas that are not using this:
1) background.js (does not init any stores/api/etc)
2) ConnectContainer.js - I'm not sure if we have access to
   ConnectorStore from here so I just kept the raw chrome.storage.local
   API call here.

We converted to/from JSON strings in (1) like with the Local Storage API, but not in (2)

### DIFF
--- a/app/ergo-connector/containers/ConnectContainer.js
+++ b/app/ergo-connector/containers/ConnectContainer.js
@@ -50,9 +50,9 @@ export default class ConnectContainer extends Component<
     const chromeMessage = this.generated.stores.connector.connectingMessage;
 
     chrome.storage.local.get('connector_whitelist', async result => {
-      const whitelist = Object.keys(result).length === 0 ? [] : result.connector_whitelist;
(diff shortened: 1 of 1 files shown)


rooooooooob:
Perhaps we should change both to use the LocalStorageApi instead for consistency?

I am needing to initiate the LocalStorageApi (but only temporarily - doesn't look like it initiates anything just a wrapper object) anyway in the background script for syncs.

Can we access the ConnectorStore from within ConnectorContainer easily?

SebastienGllmt:
As background, I noticed we added Ergo-specific code for handling storage inside `app/api/localStorage/primitives.js`. I thought these were just copy-pastes of the existing functionality we already have so I deleted these functions, but maybe it worked slightly differently.
Ideally we would use the existing functions inside `app/api/localStorage/primitives.js` for everything since they're designed to abstract away the fact that the storage API is different for extensions and websites.

paulclindo (approved):

PR #165, opened 2018-12-11
Lock screen functionality

JizKadasha:
@nicarq please check it carefully before merging, need to know if I have to change something

PR #1968, opened 2021-03-04
UI lock

SebastienGllmt:
This is mostly a rebase of #165 but a lot has changed since then so there needs to be more manual work to get this working properly.

TODO:
- [ ] Rename from screen lock to UI lock
- [ ] Update design
- [ ] Double-check if we want `bcryptjs` or not
- [ ] E2E test
- [ ] Storybook


nicarq:
@SebastienGllmt I think this one is worth considering to spend some time to merge. Obviously we need to evaluate that we are not encrypting the public key -for the people reading, the private key to sign is encrypted- I believe that making this transparent should be fine though.

Qwampa:
This feature is very much needed. Need to be able to log out, or lock yoroi UI at will.

PR #1970, opened 2021-03-05
Connector: Syncing before API calls + pending tx filtering

rooooooooob:
Ensures the DB is synced every 30 seconds before API calls.

Any TXs submitted through our submit_tx API are marked as POSSIBLY
pending for up to 10 minutes. This means that inputs in them will not be
returned as part of get_utxos or get_balance.

### DIFF
--- a/chrome/extension/background.js
+++ b/chrome/extension/background.js
@@ -20,6 +20,7 @@ import {
 } from '../../app/api/ada/lib/storage/models/PublicDeriver/traits';
 import type {
   PendingSignData,
+  PendingTransaction,
   RpcUid,
   AccountInfo,
 } from './ergo-connector/types';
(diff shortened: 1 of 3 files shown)


chrome/extension/ergo-connector/api.js
```diff
@@ -139,28 +157,31 @@ export async function connectorGetUtxos(
...
   }
   const utxos = await withUtxos.getAllUtxos();
-  // TODO: more intelligently choose values?
+  const spentBoxIds = pendingTxs.flatMap(pending => pending.tx.inputs.map(input => input.boxId));
+  // TODO: should we use a different coin selection algorithm besides greedy?
   let utxosToUse = [];
   if (valueExpected != null) {
     let valueAcc = new BigNumber(0);
     const target = valueToBigNumber(valueExpected);
     for (let i = 0; i < utxos.length && valueAcc.isLessThan(target); i += 1) {
```
SebastienGllmt, line 166: Why do we use custom code for this instead of the input selection that comes with sigma-rust?

chrome/extension/background.js
```diff
@@ -123,6 +133,50 @@ export async function getWalletsInfo(): Promise<AccountInfo[]> {
...
+}
+
+async function syncWallet(wallet: PublicDeriver<>): Promise<void> {
+  const lastSync = await wallet.getLastSyncInfo();
+  // don't sync more than every 30 seconds
+  const now = Date.now();
+  if (lastSync.Time == null || now - lastSync.Time.getTime() > 30*1000) {
+    const stateFetcher = await getStateFetcher();
+    await RustModule.load();
+    await updateTransactions(
```
SebastienGllmt: FYI this can fail if the server connection fails

chrome/extension/ergo-connector/api.js
rooooooooob: It'd end up being more work to go to/from the connector API types <-> sigma-rust wasm types, and at the time the code was first written we weren't loading the rust module, as we only need it for syncing + signing, not for the other API calls.

SebastienGllmt (approved):

PR #1988, opened 2021-03-09
add favicon from the url in connecting page

paulclindo:
Story details: https://app.clubhouse.io/emurgo/story/8138

Demo (In case url is `yoroi-wallet.com` ) : 
Screen Shot 2021-03-08 at 9 28 10 PM


packages/yoroi-extension/chrome/constants.js
```diff
@@ -64,6 +64,6 @@ export function genCSP(request: {|
     `object-src 'self' ${objectSrc.join(' ')};`,
     `connect-src ${connectSrc.join(' ')};`,
     `style-src * ${evalStyle} 'self' ${styleSrc.join(' ')} blob:;`,
-    `img-src 'self' ${imgSrc.join(' ')} data:;`,
+    `img-src 'self' ${imgSrc.join(' ')} data: https://icons.duckduckgo.com/ip3/ ;`,
```
SebastienGllmt: Although this is definitely an option, I wonder if it's possible to instead have the connector to get the icon for the page it is one and then send that over to Yoroi instead. Maybe this way we wouldn't need to depend on an external service or modify our CSP

paulclindo: @SebastienGllmt I tried to convert from the image url to a data:image format to have like `data:image/png;base64...` from connector, but I got some CORS issues at requesting the image from `icons.duckduckgo` site. Then I tried to give permissions to that url from connector manifest json but still didn't work. Do you have any different idea how to turn off CORS?  
```Access to image at 'https://icons.duckduckgo.com/ip3/localhost.ico' from origin 'http://localhost:8080' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.```

I've tested with different options to get the favicon like  `google.com/favicons`, `api.statvoo.com` but still getting CORS error.

UPDATE: WIP (I found  `services.keeweb.info` and seems to working)

File: packages/yoroi-ergo-connector/src/inject.js [hunk 2]
```diff
@@ -192,17 +210,33 @@ if (shouldInject()) {
...
+            try {
+                if (location.hostname === 'localhost') {
+                    yoroiPort.postMessage({
+                        imgBase64Url: '',
+                        type: "yoroi_connect_request",
+                        url: location.hostname
+                    });
+                    return;
+                }
+                const faviconURL = `https://services.keeweb.info/favicon/${location.hostname}`
```
SebastienGllmt: The connector already has access to the HTML for the page, so I don't think we need this kind of service. We should be able to do the following:

1. Check if an icon exists in the HTML (`document.querySelector("link[rel~='icon']")`)
2. If not, check if a favicon exists at the document root

paulclindo: I tried to do this way but I got CORS error when getting the image to parse it, even when I set `https://*` in permissions `manifest.json`

SebastienGllmt: The code to query the favicon has to be done in the context of the website and not in the context of the extension. The extension can't make requests to other websites, but the page we're injecting code into should be able to.

If the website you're injecting code into has its own CORS policy that is blocking it, that is a separate issue (and entirely possible, but I would think most websites allow querying their own favicon)

paulclindo: @SebastienGllmt I found that cross site request are blocked in content script (context of website) as of chrome 73, even when we allow them in permissions property.  So we need to make the request in the background script. I've just made it, and make it async. Let me know if that works for you 🙌

packages/yoroi-ergo-connector/manifest/manifest.template.js
```diff
@@ -30,7 +30,7 @@ module.exports = ({
     "default_title": displayName,
     default_icon: icons,
   },
-  "content_security_policy": "default-src 'none'; script-src 'self';",
+  "content_security_policy": "default-src 'none'; script-src 'self'; img-src https://* ;",
```
SebastienGllmt: If we can't put it in the content script, maybe we can put it inside `inject.js` inside the function `ergo_request_read_access` so that it still executes in the context of the user's page? 🤔

PR #1991, opened 2021-03-11
Fix multi-background page problem in connector

SebastienGllmt:
Due to the `ConnectorStore` including `background.js`, it was causing every Yoroi connector popup window to re-register all the Chrome listeners.

Problems that were caused by this:
- Clicking on the connector button opened two connector windows if you already had a connector window opened. This is because the onClick register was called twice (since there were two background pages)
- DB requests would cause weird issues. This is because all the DB requests were getting called twice (one for the real background page, one for the connector window's copy)

### DIFF
--- a/packages/yoroi-extension/chrome/extension/utils.js
+++ b/packages/yoroi-extension/chrome/extension/utils.js
@@ -122,14 +122,3 @@ export async function getWalletsInfo(): Promise<AccountInfo[]> {
     throw new GenericApiError();
   }
 }
-
-export function getActiveSites(): Array<string> {
-  const activeSites = []
-  // TODO: can't get connected sites from background.js
-  // because it runs in a different context
-  const connectedSites = [];
-  for (const value of connectedSites.values()){
-    activeSites.push(value);
-  }
-  return activeSites.map(item => item.url);
-}
--- a/packages/yoroi-extension/chrome/extension/utils.js
+++ b/packages/yoroi-extension/chrome/extension/utils.js
@@ -1,0 +1,135 @@
+// @flow
+
+import { schema } from 'lovefield';
+import type {
+  lf$Database,
+} from 'lovefield';
+import {
+  loadLovefieldDB,
+} from '../../app/api/ada/lib/storage/database/index';
+import {
+  getWallets
+} from '../../app/api/common/index';
+import {
+  asGetPublicKey,
+} from '../../app/api/ada/lib/storage/models/PublicDeriver/traits';
+import type {
+  PendingSignData,
+  RpcUid,
+  AccountInfo,
+} from './ergo-connector/types';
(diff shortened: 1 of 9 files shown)


packages/yoroi-extension/chrome/extension/utils.js
```diff
@@ -0,0 +1,135 @@
...
+    }
+    return accounts;
+  } catch (error) {
+    throw new GenericApiError();
+  }
+}
+
+export function getActiveSites(): Array<string> {
+  const activeSites = []
+  // TODO: can't get connected sites from background.js
```
SebastienGllmt: Will need to think of a better way to get this information since the background page is the one managing this. Maybe this will have to be passed as a message or passed through storage.

packages/yoroi-extension/chrome/extension/utils.js
```diff
@@ -0,0 +1,135 @@
...
+    const wallets = await getWallets({ db });
+    // information about each wallet to show to the user
+    const accounts = [];
+    for (const wallet of wallets) {
+      const conceptualWallet = wallet.getParent();
+      const withPubKey = asGetPublicKey(wallet);
+
+      const conceptualInfo = await conceptualWallet.getFullConceptualWalletInfo();
+      if (isErgo(conceptualWallet.getNetworkInfo())) {
+        // TODO: we can't get the pending txs from background.js from here
```
SebastienGllmt, line 109: I think this one is maybe not that big of a deal since I think we want to get rid of the `AccountInfo` type anyway. This `getWalletsInfo` is only used from the `ConnectorStore` so it probably doesn't even belong in this utils file.

packages/yoroi-extension/chrome/extension/utils.js
rooooooooob: When we go to manifest v3 we'll probably end up storing it in local storage since you can't have global variables in the new web workers, but if they change the port API we might not need to.

packages/yoroi-extension/chrome/extension/utils.js
rooooooooob: Pending txs doesn't really matter too much here anyway - at most you just get a different balance, and afaik this is (only?) used on the connect screen, so we probably don't have any anyways. The main reason I had it was to stop spent UTXOs from being given twice unnecessarily.

rooooooooob (approved):
I think we still need that workaround that I made though since even if they weren't registered twice, we'd still have the potential for multiple RPCs to come at us since the injected API is async. The test dApp for example does several at once on startup to test the API calls/cases that weren't covered by the regular flow of the test dApp. It might be more of an issue for syncing though since unless two dApps requested it at once we probably wouldn't run into the double DB load issue, just the double sync.

PR #1993, opened 2021-03-12
Connector: error when trying to connect with no ergo wallets in the Yoroi Extension

paulclindo:
Story details: https://app.clubhouse.io/emurgo/story/8125


paulclindo:
@SebastienGllmt Right now it shows `No wallets found` but I think we can give extra info like `We couldn’t find any wallets. Make sure to have an ergo wallet and try it again`  What do you think? cc @Marta-Ryzhok 
Screen Shot 2021-03-11 at 9 39 09 PM

SebastienGllmt:
I would specify something like "No Ergo mainnet wallets found", otherwise people may be confused if we tell them "no wallets found" but they have Cardano wallets.
Probably the best way to do this is to avoid hard-coding "Ergo mainnet" as part of the string, and instead add the network name as a new field in `networks.js`

Marta-Ryzhok:
@paulclindo yeah agree with you both, explanation better be more explicit, what @SebastienGllmt suggested

PR #1994, opened 2021-03-12
Connector: disconnects

rooooooooob:
Port disconnects cause the disconnect event to be dispatched to the
dApp.

Removing a site from the whitelist disconnects the associated port, if
it exists (e.g. actively connected).

Ports are now only initiated upon a request for read access to avoid
creating many ports for non-dApp websites.

### DIFF
--- a/packages/yoroi-ergo-connector/src/inject.js
+++ b/packages/yoroi-ergo-connector/src/inject.js
@@ -1,15 +1,13 @@
 // sets up RPC communication with the connector + access check/request functions
 const initialInject = `
-var timeout = 0;
-
-var connectRequests = [];
+var ergoConnectRequests = [];
 
 window.addEventListener("message", function(event) {
     if (event.data.type == "connector_connected") {
         if (event.data.err !== undefined) {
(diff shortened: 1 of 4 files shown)


packages/yoroi-ergo-connector/src/inject.js
```diff
@@ -139,6 +114,8 @@ class ErgoAPI {
 
 const ergo = Object.freeze(new ErgoAPI());
 `
+const API_INTERNAL_ERROR = -2;
```
SebastienGllmt, line 115: Three things:
- Maybe these error codes should be documented somewhere? 
- Maybe these should should be in an enum instead of two separate variables
- Why does it start at -2 instead of -1? Is -1 used for something else? Maybe document this

rooooooooob: https://github.com/ergoplatform/eips/pull/23 they're documented here in the EIP-0012 spec. They do start at -1, but we just don't reject things for `InvalidRequest: -1` in the connector since input data verification happens elsewhere.

SebastienGllmt (approved):

PR #2001, opened 2021-03-19
Connector: popup location

paulclindo:
Story details: https://app.clubhouse.io/emurgo/story/8144
Screen Recording 2021-03-18 at 08 13 25 PM

### DIFF
--- a/packages/yoroi-ergo-connector/src/inject.js
+++ b/packages/yoroi-ergo-connector/src/inject.js
@@ -233,22 +233,11 @@ if (shouldInject()) {
                     createYoroiPort();
                 }
 
-                // Get measures from DOM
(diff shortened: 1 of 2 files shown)


SebastienGllmt:
It seems this always pops up at the top-right of your screen regardless of where your Chrome window actually is. It should appear at the top-right of wherever your Chrome browser is on your screen

paulclindo:
Yeah, it's because I need to get the width, height, and position from the content script and send it once it connects to yoroi background. So after you open `localhost:8080`, it's gonna be below the browser icon. 
Here's the demo: https://share.getcloudapp.com/4gu1kGxy
I'm not quite sure if there's another solution to

PR #2006, opened 2021-03-22
Handle wallets being deleted while they are being used by the connector

SebastienGllmt:
Before this PR:

 1. Create Ergo wallet
 1. Connect it to a dApp
 1. Open Yoroi and delete the wallet
 1. Open the connector and look at the connected sites. It will crash

### DIFF
--- a/packages/yoroi-extension/app/ergo-connector/components/connect-websites/ConnectWebsitesPage.js
+++ b/packages/yoroi-extension/app/ergo-connector/components/connect-websites/ConnectWebsitesPage.js
@@ -17,7 +17,7 @@ import type { TokenLookupKey } from '../../../api/common/lib/MultiToken';
 import type { TokenRow } from '../../../api/ada/lib/storage/database/primitives/tables';
 
 type Props = {|
-  +accounts: ?Array<WhitelistEntry>,
+  +whitelistEntries: ?Array<WhitelistEntry>,
   +activeSites: Array<string>,
   +wallets: ?Array<PublicDeriverCache>,
   +onRemoveWallet: ?string => void,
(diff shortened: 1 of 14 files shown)


packages/yoroi-extension/app/ergo-connector/components/connect-websites/ConnectWebsitesPage.js
```diff
@@ -63,19 +63,27 @@ export default class ConnectWebsitesPage extends Component<Props> {
...
-              onRemoveWallet={this.props.onRemoveWallet}
-              getTokenInfo={this.props.getTokenInfo}
-              shouldHideBalance={this.props.shouldHideBalance}
-            />
-          ))}
+          {whitelistEntries.map(({ url, publicDeriverId }) => {
+            const wallet = wallets.find(
+              cacheEntry => cacheEntry.publicDeriver.getPublicDeriverId() === publicDeriverId
+            );
+            // note: store should make sure that deleted wallets got removed from the whitelist
```
SebastienGllmt, line 70: TODO: I haven't done this yet

packages/yoroi-extension/chrome/extension/background.js
```diff
@@ -272,10 +297,16 @@ chrome.runtime.onMessage.addListener(async (request, sender, sendResponse) => {
...
     const connection = connectedSites.get(request.tabId);
     if (connection && connection.status != null && typeof connection.status === 'object') {
-      connection.status.resolve(request.accepted ? request.account : null);
-      connection.status = request.account;
+      if (request.accepted === true) {
+        connection.status.resolve(request.publicDeriverId);
+        connection.status = request.publicDeriverId;
+      } else {
+        connection.status.resolve(null);
+        // TODO: what to do here? Delete entry from connectedSites? Do nothing?
```
SebastienGllmt: @rooooooooob do you know how this is supposed to be handled?

rooooooooob: Yes, remove the entry from `connectedSites`. it will get re-added the next time the dApp tries to connect and there's no point keeping it around. It might be a good idea to also update how `'get_connected_sites'` is handled to check if the site is connected yet since it could be in `connectedSites` but have the status be pending.

packages/yoroi-extension/chrome/extension/background.js
```diff
@@ -295,10 +326,17 @@ chrome.runtime.onMessage.addListener(async (request, sender, sendResponse) => {
           }
           break;
         case 'tx_input':
+          // mocked data sign
```
rooooooooob: Why does it say mocked data sign?

PR #2025, opened 2021-04-28
Update ledger

SebastienGllmt:
Ledger API was rewritten in Typescript and the API changes as well.

This PR updates to the Yoroi code & dependencies that support this refactor

### DIFF
--- a/packages/yoroi-extension/chrome/manifest.template.js
+++ b/packages/yoroi-extension/chrome/manifest.template.js
@@ -66,7 +66,7 @@ export default ({
     permissions: [
       'storage',
       '*://connect.trezor.io/*',
-      'https://emurgo.github.io/yoroi-extension-ledger-connect/*'
+      'https://emurgo.github.io/yoroi-extension-ledger-connect-vnext/*'
     ],
     content_scripts: [
       {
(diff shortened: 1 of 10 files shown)


packages/yoroi-extension/chrome/manifest.template.js
```diff
@@ -66,15 +66,15 @@ export default ({
     permissions: [
       'storage',
       '*://connect.trezor.io/*',
-      'https://emurgo.github.io/yoroi-extension-ledger-connect/*'
+      'https://emurgo.github.io/yoroi-extension-ledger-connect-vnext/*'
```
SebastienGllmt, line 69: I'm hoping that since this is the same domain as the previous URL that it doesn't re-prompt users to accept this extension's permission :/
We have no good way to version this library, so deploying to a different URL is the only clean way to do it.

packages/yoroi-extension/package.json
```diff
@@ -149,13 +149,13 @@
     "ws": "7.4.3"
   },
   "dependencies": {
-    "@cardano-foundation/ledgerjs-hw-app-cardano": "2.2.1",
+    "@cardano-foundation/ledgerjs-hw-app-cardano": "git+https://github.com/SebastienGllmt/ledgerjs-hw-app-cardano.git",
```
SebastienGllmt, line 152: We can switch to the main repo once https://github.com/cardano-foundation/ledgerjs-hw-app-cardano/pull/20 is deployed

PR #2052, opened 2021-04-30
fix typo in translation

chrissiwaffler:
fixed small typo in the german translation


zuzunker:
Hi, @CodingLegend27 ! Thank you for this contribution! But unfortunately the translation files cannot be edited directly, in order to help with any translation - you need to do it via the Crowdin platform, we have a lot of community members there helping us with supporting various languages.

Link to this specific message there: https://crowdin.com/translate/yoroi-app/34/en-de#18992

I have to close this PR, because if we change it directly - it will cause conflicts with the Crowdin.

PR #2081, opened 2021-05-04
Data inputs + P2S support

rooooooooob:
Data inputs have been accepted by the network
P2S inputs seem to work as both inputs and outputs.

Tested:
`HEIGHT > 1337`
`OUTPUTS(0).R4[Int].get == 12345`


packages/yoroi-extension/chrome/extension/ergo-connector/api.js
```diff
@@ -262,13 +266,36 @@ export async function connectorSignTx(
...
+  console.log('before signing');
   const signedTx = RustModule.SigmaRust.Wallet
     .from_secrets(wasmKeys)
     .sign_transaction(
-      RustModule.SigmaRust.ErgoStateContext.dummy(), // TODO? Not implemented in sigma-rust
+      new RustModule.SigmaRust.ErgoStateContext(preHeader),
       wasmTx,
       txBoxesToSign,
-      RustModule.SigmaRust.ErgoBoxes.from_boxes_json([]), // TODO: not supported by sigma-rust
+      RustModule.SigmaRust.ErgoBoxes.from_boxes_json([]),
```
SebastienGllmt: I think the connector doesn't rely on [utxoTransactions.js](https://github.com/Emurgo/yoroi-frontend/blob/develop/packages/yoroi-extension/app/api/ergo/lib/transactions/utxoTransaction.js#L469), right?

rooooooooob: It uses `generateKeys()` but not `signTransaction()` so we don't need to worry about that TODO there for the connector.

packages/yoroi-extension/chrome/extension/ergo-connector/api.js
```diff
@@ -262,13 +266,40 @@ export async function connectorSignTx(
   const jsonBoxesToSign = utxosToSign.map(formatUtxoToBox);
   processBoxesForSigmaRust(jsonBoxesToSign);
   const txBoxesToSign = RustModule.SigmaRust.ErgoBoxes.from_boxes_json(jsonBoxesToSign);
+  console.log('data inputs');
+  const dataBoxIds = tx.dataInputs.map(box => box.boxId);
```
rooooooooob, line 268: I think we might have to update the spec to include the full ErgoBox for data inputs too to match the unsigned inputs. We never updated the signing code since we primarily made that change for the UI side to display the data. I'm guessing we want to take out these yoroi-supplied utxos and directly use the data from `UnsignedInput` here too? Likewise for `DataInput` if we do the same there.

packages/yoroi-ergo-connector/example/index.js
```diff
@@ -104,15 +123,34 @@ function initDapp() {
...
+                    txBuilder.set_data_inputs(dataInputs);
                     const tx = txBuilder.build().to_json();
-                    console.log(`tx: ${tx}`);
+                    console.log(`tx: ${JSON.stringify(tx)}`);
+                    console.log(`original id: ${tx.id}`);
+                    // sigma-rust doesn't support most compilation so manually insert it here
+                    // this is HEIGHT > 1337 but in hex and without the checksum/etc for the address of the contract
+                    //tx.outputs[0].ergoTree = "100104f214d191a37300";
+                    // and we rebuild it using
+                    const correctTx = wasm.UnsignedTransaction.from_json(JSON.stringify(tx)).to_json();
```
greenhat, line 124: What is the reason for JSON roundtrip for UnsignedTransaction? Do `original id:` and `new id:` have different ids?

rooooooooob: It was for line 135 - I was adding a P2S address as an output to test something, and wasn't able to do it via sigma-rust directly so I just manually changed one of the ergoTrees then using my modified fork used just for the dapp for testing of sigma-rust that ignores the txid and recalculates it upon from_json I do the round-trip to recalculate the txid to be the proper one with that new ergotree. The resulting TX was accepted by the network, the problem was when we tried to spend it. I think I've found the problem for the different IDs though, I'll try to push a new commit soon. It was inside of Yoroi not the dApp.

greenhat: > Do `original id:` and `new id:` have different ids?

When UnsignedTransaction is parsed from JSON it recalculates the tx id and if it differs from the one provided in JSON it fails the parsing process. So I wonder how they could be different in the logs.

rooooooooob: They aren't different between the two logs here, it was different between here and the resulting signed tx spit out by Yoroi. What was happening was when the signing logic was written we needed some internal datastructure in Yoroi for unspent boxes to get the keys, and to do that we filtered the user's UTXOs based on the inputs we needed to sign, which since a P2S address would never be inside the user's tracked UTXOs would silently filter that out during signing. It's fixed now though! The TX is accepted by the network: https://explorer.ergoplatform.com/en/transactions/59b94cfc3244d1358a2b06a1abaab2ee16884e2589554c42a593cf8ecffbb33e

packages/yoroi-extension/chrome/extension/ergo-connector/api.js
```diff
@@ -259,16 +263,43 @@ export async function connectorSignTx(
     keyLevel: wallet.getParent().getPublicDeriverLevel(),
```
rooooooooob, line 262: What about P2S inputs that include keys we might need to be able to sign with? Would we have to look into the context extension or into the ergoscript to be able to figure out what those keys are to be able to sign those inputs?

rooooooooob:
I tried testing P2S with registers to see if it works and I can create it (albeit with a bit of difficulty since I didn't see register/constant support in the bindings so I manually created a `ergotree-ir::mir::constant::Constant` in rust to `sigma_serialize_bytes()` then hex encode then reparse that via JSON using my sigma-rust fork which ignores provides tx ids to correct them.

Here's the TX it's created in:
https://explorer.ergoplatform.com/en/transactions/c8bca66018be7b83247c3508a1b32d247fb9d6deebe2b2fd8c6e119f165f0412

(0 index output)

with JSON:

```json
{
   "id":"c189e6b13b497d3f538249a9f16aad233ad9f45b4577901c00124e980da26ad9",
   "inputs":[
      {
         "boxId":"c1e9ba12a5ab146c8ee81afe45db8fb18cca2da1912fdad7a15fff09fa920864",
         "value":16427378,
         "ergoTree":"0008cd03622070184643e8089c6ff367dc648eacc6bcf9ee34c07fafa7e61075df25b58b",
         "assets":[
            
         ],
         "creationHeight":398959,
         "additionalRegisters":{
            
         },
         "transactionId":"bc90e638ce91180be8d6f63dbc6ccf30fc68a7bf093740a2415a57533be3a3da",
         "index":1,
         "extension":{
            
         }
      },
      {
         "boxId":"db437a5e0399a031b8fd77c3c176011561f1e0c80aee59fed6b0364dec5c752e",
         "value":1659312,
         "ergoTree":"1002040004a48bb099f9ffffffff01d193e4c6b2a573000004047301",
         "assets":[
            
         ],
         "additionalRegisters":{
            
         },
         "creationHeight":398959,
         "transactionId":"c8bca66018be7b83247c3508a1b32d247fb9d6deebe2b2fd8c6e119f165f0412",
         "index":0,
         "extension":{
            
         }
      }
   ],
   "dataInputs":[
      
   ],
   "outputs":[
      {
         "boxId":"7dbc699d8012e172ef09fa06609c72f0829247aba109fa69ac4430285181f9d8",
         "value":16986690,
         "ergoTree":"0008cd03622070184643e8089c6ff367dc648eacc6bcf9ee34c07fafa7e61075df25b58b",
         "assets":[
            
         ],
         "additionalRegisters":{
            "R4":"04a48bb09909"
         },
         "creationHeight":398959,
         "transactionId":"c189e6b13b497d3f538249a9f16aad233ad9f45b4577901c00124e980da26ad9",
         "index":0
      },
      {
         "boxId":"16191dba1a1e082a330b05bb89e24f42f6561ea8c86e99eb61a8da31c5c68694",
         "value":1100000,
         "ergoTree":"1005040004000e36100204a00b08cd0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798ea02d192a39a8cc7a701730073011001020402d19683030193a38cc7b2a57300000193c2b2a57301007473027303830108cdeeac93b1a57304",
         "assets":[
            
         ],
         "additionalRegisters":{
            
         },
         "creationHeight":398959,
         "transactionId":"c189e6b13b497d3f538249a9f16aad233ad9f45b4577901c00124e980da26ad9",
         "index":1
      }
   ]
}
```

Signing gives the error:
```
Prover error (tx input index 1): Ergo tree error: Whole ErgoTree parsing (deserialization) error
undefined
```

With the following ergoscript:
https://wallet.plutomonkey.com/p2s/?source=T1VUUFVUUygwKS5SNFtJbnRdLmdldCA9PSAxMjM0NTY3ODkw

`OUTPUTS(0).R4[Int].get == 1234567890`

@greenhat In the `ergotree-ir` crate it mentions `R0 .. R9` being supported in `Box`, as well as `OUTPUTS` in `Context` and `==` operator in `Operations` so I thought it should be fine. What am I doing wrong?

It gets the exact same error even without the register (although then the input shouldn't validate) so I don't think it's that. It really doesn't like that P2S ergotree.

Or if this is just an issue with it not being supported by `sigma-rust`, then should we just merge this PR and mention that until `sigma-rust` is updated we only support signing certain P2S inputs?

greenhat:
> I tried testing P2S with registers to see if it works and I can create it (albeit with a bit of difficulty since I didn't see register/constant support in the bindings so I manually created a `ergotree-ir::mir::constant::Constant` in rust to `sigma_serialize_bytes()` then hex encode then reparse that via JSON using my sigma-rust fork which ignores provides tx ids to correct them.

There is an `ErgoBoxCandidateBuilder` API with register and constant support. See the following example:
https://github.com/ergoplatform/sigma-rust/blob/c6aa36be641fae1ecbf0fe93b8e7b395e759dec8/bindings/ergo-lib-wasm/tests/test_box_builder.js#L17-L26

 
> Signing gives the error:
> 
> ```
> Prover error (tx input index 1): Ergo tree error: Whole ErgoTree parsing (deserialization) error
> undefined
> ```
> 
> With the following ergoscript:
> https://wallet.plutomonkey.com/p2s/?source=T1VUUFVUUygwKS5SNFtJbnRdLmdldCA9PSAxMjM0NTY3ODkw
> 
> `OUTPUTS(0).R4[Int].get == 1234567890`
> 
> @greenhat In the `ergotree-ir` crate it mentions `R0 .. R9` being supported in `Box`, as well as `OUTPUTS` in `Context` and `==` operator in `Operations` so I thought it should be fine. What am I doing wrong?
> 
> It gets the exact same error even without the register (although then the input shouldn't validate) so I don't think it's that. It really doesn't like that P2S ergotree.
> 
> Or if this is just an issue with it not being supported by `sigma-rust`, then should we just merge this PR and mention that until `sigma-rust` is updated we only support signing certain P2S inputs?

It sounds silly, but I see we fail to parse `1234567890` int constant. :) I'm digging into our VLQ decoding and planning to fix it ASAP. 

EDIT: I mean we fail to parse the constant as part of ergo tree parsing. Register value parses just fine.

greenhat:
@rooooooooob From what I found so far integer value larger than 1073741823 fails to decode from VLQ form. Also encoding of these values seems to be incorrect in sigma-rust. I'm still digging in and extending test suite in sigma-rust with reference values from scala version to make sure that correct encoding/decoding. 

Meanwhile, you can test this PR with a smaller value (than 1073741823) in the contract to be sure it works fine.

rooooooooob:
I tried it now with just `12345` and it seems to work and was accepted by the network!

I cleaned up a few of the log statements so I think this PR is final on my end if you guys think the code changes are fine.

zuzunker (approved):

PR #2089, opened 2021-05-12
Trezor wallet Catalyst voting registration support

yushih:
[ch8805]
Only the API level work. UI will be updated later.
### `VotingStore.js`:
1. When the "PIN confirmation" step is submitted, if it is Trezor wallet, just call `createTransaction` action and skip the next step (asking for spending password).

2. `_createTransaction`: For hardware wallets, do not call `generateRegistration` to generate the tx metadata, because the metadata is generated by the wallet. However, we do need to pass the public voting key down (instead of the generated metadata).

###  `api/ada/transactions/shelley/HaskellShelleyTxSignRequest.js` and `api/ada/index.js`
`createVotingRegTx:` Pass the public voting key and nonce down instead of the generated metadata.

### `api/ada/transactions/shelley/trezorTx.js`
If it is a catalyst voting registration tx, construct the `auxiliaryData` of the signing request using the public voting key and nonce passed down.

### DIFF
--- a/packages/yoroi-extension/app/api/ada/index.js
+++ b/packages/yoroi-extension/app/api/ada/index.js
@@ -374,11 +374,23 @@ export type CreateDelegationTxRequest = {|
   valueInAccount: MultiToken,
 |};
 
-export type CreateVotingRegTxRequest = {|
+type CreateVotingRegTxRequestCommon = {|
   publicDeriver: IPublicDeriver<ConceptualWallet> & IGetAllUtxos & IHasUtxoChains,
   absSlotNumber: BigNumber,
-  metadata: RustModule.WalletV4.GeneralTransactionMetadata,
 |};
+
+export type CreateVotingRegTxRequest = {|
+  ...CreateVotingRegTxRequestCommon,
+  normalWallet: {|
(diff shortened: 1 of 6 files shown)


packages/yoroi-extension/app/api/ada/index.js
```diff
@@ -1431,6 +1451,12 @@ export default class AdaApi {
           neededHashes: new Set(),
           wits: new Set(),
         },
+        trezorTCatalystRegistrationTxSignData:
+          request.trezorTWallet &&
+            {
+              publicVotingKey: request.trezorTWallet.publicVotingKey,
+              nonce: request.absSlotNumber,
+            },
```
zuzunker: @yushih , you are using `.flag && value` here, which I think might leave the `trezorTCatalystRegistrationTxSignData` to have value `false` or something like that, in cases when the flag is not truthy. Let's use
```
request.trezorTWallet ? { ... } : undefined,
```

Because in some other place eventually someone might use a `!= null` check and then we'll have a problem.

zuzunker (changes requested):

zuzunker (approved):

PR #2096, opened 2021-05-17
Ledger wallet Catalyst voting registration support

yushih:
[ch8804]
*Prerequisite*: this change https://github.com/Emurgo/yoroi-extension-ledger-connect-handler/pull/51/commits/3ebe9bc70c62b551141b62147d09b60e86bef5d0 be published to NPM as version 5.0.0 

Because when Ledger signs the tx, it only returns the signature part of the CIP-15 aux data, in `VotingStore._createTransaction` I have to pass the staking pub key and reward address along with the staking key path down the call chain. And in `LedgerSendStore.signAndBroadcast`, I construct the aux data (aka `metadata`) using the data passed down and the signature returned from Ledger and put it into the tx body.


zuzunker:
@yushih , plz resolve conflicts

zuzunker (approved):

zuzunker (commented):
/check

PR #2100, opened 2021-05-20
UI updates for hardware wallet Catalyst registration

yushih:
[ch8806]
Delegation status warning are moved to the top of the dialog according to the design.
The round number is fixed by using the constant defined in code instead of the hardcoded value in the message.

Add specific instructions for using hardware wallet:
trezor0
ledger0

Fix the progress bar:
progress

Add specific instructions for hardware wallet before sending the tx:
tx-trezor


yushih:
Update: since the round number has been fixed by Vicente and merged first, the commit is removed.

zuzunker (commented):
/check

zuzunker (approved):

PR #2103, opened 2021-05-25
Migration to SigmaRust 0.13.x

oskin1:
Recently several mistakes in ErgoRust data model were fixed. Namely `UnsignedTransaction` now requires `ErgoBoxCandidate[]` rather than `ErgoBox[]`. Also `UnsignedTransaction` no longer requires an `id` field.


zuzunker:
@oskin1 , thank you for the contribution! Can you plz resolve the conflict with the package-lock? And then we can get this diff in

oskin1:
> @oskin1 , thank you for the contribution! Can you plz resolve the conflict with the package-lock? And then we can get this diff in

Done👌

packages/yoroi-extension/chrome/extension/ergo-connector/types.js
```diff
@@ -270,10 +301,9 @@ export function asTx(
         Array.isArray(tx.dataInputs) &&
         Array.isArray(tx.outputs)) {
       return {
-        id: asTxId(tx.id),
```
rooooooooob, line 273: I'm totally fine with removing id since it's honestly more annoying than useful IMO for unsigned tx so that's nice that it's no longer required by sigma-rust. We should remove `id` from the `Tx` type above too then. I'll need to update the EIP-0012 spec as well.

rooooooooob (changes requested):

packages/yoroi-extension/chrome/extension/ergo-connector/types.js
oskin1: Done

rooooooooob (approved):
LGTM

oskin1:
@rooooooooob it seems https://github.com/rooooooooob/eips/blob/eip-0012/eip-0012.md#unsignedinput is inconsistent with what Yoroi actually expects as an `UnsignedInput` https://github.com/Emurgo/yoroi-frontend/blob/583ec3d2e825dbe6b4d19ece6246f116d82f7502/packages/yoroi-extension/chrome/extension/ergo-connector/types.js#L145

zuzunker (commented):
/check

rooooooooob:
> @rooooooooob it seems https://github.com/rooooooooob/eips/blob/eip-0012/eip-0012.md#unsignedinput is inconsistent with what Yoroi actually expects as an `UnsignedInput`
> 
> https://github.com/Emurgo/yoroi-frontend/blob/583ec3d2e825dbe6b4d19ece6246f116d82f7502/packages/yoroi-extension/chrome/extension/ergo-connector/types.js#L145

Thanks for reminding me. We added in that extra information for #2081 but I didn't get around to updating the spec yet. I just updated the spec for both that change and the one in this PR about box candidates.

zuzunker (commented):
/check

zuzunker (approved):

rooooooooob:
@oskin1 I'm actually getting issues signing TXs (at least via our example dApp in `packages/yoroi-ergo-connetor/example/`). I'm not sure if it's an issue with something being cached locally for me, or our test dApp, or something else so I'd like to be able to see if that's the case.

We hadn't been able to test this before merging since we had assumed it was working fine on your end and Yoroi's `develop` was experiencing an issue at the time which made testing impossible when I tried to check out this PR and run that. That's resolved with #2151 now. cc @vantuz-subhuman 

However, I get an `unknown field "tx_id"` error when trying to sign the tx on `develop` but when I revert to a commit before this PR it seems to work. I ran `npm install` on both Yoroi and the dApp to make sure they were both on sigma-rust `0.13.1`, and likewise when I reverted for 0.10.0 on both. I checked in `node_modules/` too and the versions line up.

Can you please check with whatever you were using to send the connector txs made with this PR? Thanks. Hopefully it's just something on my local machine.

rooooooooob:
Okay I solved the underlying issues there. We had a few issues related to versions which are fixed in #2166 and then there was also an issue with sigma-rust's json deserialization but I made a PR for sigma-rust here ergoplatform/sigma-rust#304 which after testing locally fixes that `tx_id` unknown field issue in the connector, but we'll have to wait for that to get finalized before updating the connector to fix that issue.

PR #2151, opened 2021-06-01
Fix for Ergo unknown address type error in develop

rooooooooob:
When we updated sigma-rust the flow types weren't updated which caused
https://github.com/ergoplatform/sigma-rust/commit/30be67d3b748bf73ac0024f58827d3c86f051241
to give us undefined instead of 1 or 2.


zuzunker (approved):

zuzunker (commented):
/check

PR #2167, opened 2021-06-09
Improve and fix transaction display

yushih:
#### More specific intrawallet transaction types
1. If the intrawallet transaction has withdrawal, it is shown as "Reward withdrawn"
reward

2. If the intrawallet transaction has catalyst registration metadata, it is shown as "Catalyst voting registered"
catalyst

3. If the intrawallet transaction has delegation certificate, it is shown as "Stake delegated"
delegated

#### Fix withdraw transaction amount
Previously, withdraw transaction amount does not account for the amount of rewards withdrawn, thus its value could be `-fee`, which is misleading. Now the reward amount is added.
Before:
before
After:
after


SebastienGllmt:
> More specific intrawallet transaction types

The problem is that theoretically a transaction could be multiple things. For example, you can have a tx that both registers for Catalyst and withdraws rewards. To address this, we were thinking about adding badges next to a transaction with each badge representing some action included in this transaction. We had Marta make designs for this, but we never had enough developers to implement it

> Fix withdraw transaction amount

This was by design because if you include the reward amount, the sum of the transaction history no longer equals your wallet balance which is also counter-intuitive and also inaccurate for tx history export for tax purposes.

v-almonacid:
Yeah it's a bit trickier, the idea of having badges sounds good though.

>  the sum of the transaction history no longer equals your wallet balance 

@SebastienGllmt Not sure if I understand here. Why rewards would mess up the balance? 🤔

SebastienGllmt:
Imagine a MIR certificate gives 5 ADA to your wallet (reminder: MIR certificate goes to your reward address)
Then you withdraw the 5 ADA to your spendable UTXO balance.

Now your transaction history will show two transactions with the summary being +5 and +5 for both these transactions, but your wallet only actually has 5 ADA in it.

v-almonacid:
> Imagine a MIR certificate gives 5 ADA to your wallet (reminder: MIR certificate goes to your reward address)
> Then you withdraw the 5 ADA to your spendable UTXO balance.
> 
> Now your transaction history will show two transactions with the summary being +5 and +5 for both these transactions, but your wallet only actually has 5 ADA in it.

Great explanation, thanks 😄 If it's only just that edge case that messes everything up, we can probably just assume that MIR transactions do not add coins to your balance until those coins are withdrawn 🤔

vantuz-subhuman:
> Now your transaction history will show two transactions with the summary being +5 and +5 for both these transactions, but your wallet only actually has 5 ADA in it.

🤔 For this case it seems to me like the summary for the reward-to-utxo tx should be `+0` or something like `-5 +5` instead of `+5` cuz you did not gain any funds, you just moved them within the same wallet

SebastienGllmt:
@ruslan the summary for the reward-to-utxo tx (AKA withdrawal) already is +0. This PR is changing it to be +5

That's why I'm saying we didn't make this change on purpose

v-almonacid:
IMO is the MIR tx that should be +0, since the funds are kept in your rewards account, IIRC. So only withdrawals from reward account to utxo should add to the balance.

yushih:
Udate:
For each tx, get an array of features:
```
  | 'CatalystVotingRegistration'
  | 'Withdrawal'
  | 'StakeDelegation'
  | 'StakeRegistration'
  | 'StakeDeregistration'
  | 'PoolRegistration'
  | 'PoolRetirement'
  | 'GenesisKeyDelegation'
  | 'MoveInstantaneousRewards'
```
and be strict in labeling intrawallet tx:
- withdrawal tx either contains only withdrawal (user "keep registered") or withdrawal and deregistration (user "deregister")
- catalyst registration tx contains only catalyst registration metadata
- stake delegation tx either contains only stake delegation or stake delegation and stake key registration

In the future, the set of features can be mapped to badges (when the design is ready).

For now, reverted the change on withdraw tx amount.

zuzunker (approved):
This is a good improvement, let's get this in

zuzunker (commented):
/check

PR #2170, opened 2021-06-11
Ergo Connector: Fix crash when signing a TX that mints new assets

### LINKED ISSUE #2162: Getting a blank popup window when trying to sign a transaction with ergo-dapp-connector

rooooooooob:
We rely on the ergo explorer backend to look up token info, but this
won't work when we're minting a new token.

So instead, we will check if the tx minted a token and create the token
info from the registers according to [EIP-0004](https://github.com/ergoplatform/eips/blob/master/eip-0004.md)

This will fix #2162


packages/yoroi-extension/chrome/extension/ergo-connector/utils.js
```diff
@@ -0,0 +1,66 @@
...
+// you should not be able to mint more than 1
+export function mintedTokenInfo(tx: Tx): $ReadOnly<TokenRow>[] {
+  const tokens = []
+  for (const output of tx.outputs) {
+    const name = parseEIP0004Data(output.additionalRegisters.R4);
+    const description = parseEIP0004Data(output.additionalRegisters.R5);
+    const decimals = parseEIP0004Data(output.additionalRegisters.R6);
+    if (name != null && description != null && decimals != null) {
+      tokens.push({
+        TokenId: 0,
```
rooooooooob, line 45: I don't think we use the digest/networkid/tokenid in the tx signing page. I could try and figure out the data for them but we only use this for tx signing.

Should we bother getting this info? I think we can get the network id from looking at which addresses are used in the tx but it's a bit indirect and I'd need to use the rust module, or we could pass that in. I'm not sure where to get the digest from. The tokenId thing afaik is something tracked just within Yoroi so I don't think we can easily create something valid here.

We also have both a ticker and a longName here but there is only one name in the EIP-0004 spec so I'm not sure where we get those from.

packages/yoroi-extension/app/ergo-connector/containers/SignTxContainer.js
```diff
@@ -101,6 +101,7 @@ export default class SignTxContainer extends Component<
                 ? null
                 : uiNotifications.getTooltipActiveNotification(this.notificationElementId)
             }
+            tx={signingMessage.sign.tx}
```
rooooooooob, line 104: @paulclindo The txData is used by the UI, by the problem I ran into was whoever was using it assumed all assets were pre-existing, so I also passed in the JSON tx sent to use which will contain the information about newly minted assets.

packages/yoroi-extension/chrome/extension/ergo-connector/utils.js
```diff
@@ -0,0 +1,66 @@
+// @flow
+
+import type { Tx } from './types';
+import type { TokenRow } from '../../../app/api/ada/lib/storage/database/primitives/tables';
+import { Logger } from '../../../app/utils/logging';
+
+function parseEIP0004Data(input: any): ?string {
```
rooooooooob, line 7: I didn't use the vlq npm package here since:
1) it uses base64url
2) we need to partially read the string, not the entire thing, and the use the intermediate info of how far we parsed into the vlq-encoded length to then read the remainder of the string

I tested the vlq part of it against the examplet test vectors on wikipedia + the entire thing on the ones in the EIP-0004 standard

packages/yoroi-extension/app/ergo-connector/containers/SignTxContainer.js
zuzunker: Sounds good

packages/yoroi-extension/chrome/extension/ergo-connector/utils.js
zuzunker: I wouldn't spend time on this until we absolutely need it. We'll put a note about it and do later when actually needed.

zuzunker (approved):

zuzunker (commented):
/check

PR #2171, opened 2021-06-11
change icon "transactions" 

neuodev:
- Task Id: 8732
- The old images removed


packages/yoroi-extension/app/components/wallet/navigation/WalletNavButton.js
```diff
@@ -35,7 +36,11 @@ export default class WalletNavButton extends Component<Props> {
       <button type="button" className={componentClasses} onClick={onClick}>
         {IconComponent != null &&
           <div className={styles.icon}>
-            <IconComponent />
+            {
+              label === 'Transactions' && isActive ?
+              <TransactionImageActive />
+              : <IconComponent />
+            }
```
zuzunker: @AhmedIbrahim336 , this is not a great way to add some logic:

1. A component like `WalletNavButton` is very abstract and should not be aware of a specific icon like `TransactionImageActive`. It works only with generic properties and should be kept this way.

2. You cannot use label for a hardcoded check to decide which tab is it. Not only your logic will only work if no one ever changes the text, but also all labels we have are `intl` rendered strings, which means that it will be different for different languages.

You need to change this. This logic must be moved to another place where the different tabs are actually handled. A button shouldn't do that.

zuzunker (changes requested):

packages/yoroi-extension/app/components/wallet/navigation/WalletNavButton.js
neuodev: I got your point 
so the button is a reusable component it shouldn't have this logic in it.
Also this will not work with different languages and when we change the text.
Okay I will try to think for another way to do it

neuodev:
@vantuz-subhuman 
So the new updates are 
- I moved the logic for the parent component the `main Nav`
- I checked for the `ID` no the label the advantages for this is it's unique, working with all languages ( I test it )
- When you change the text `Transactions` in the future it will work 
- When you update the` id` it will work as will

packages/yoroi-extension/app/components/wallet/navigation/WalletNavigation.js
```diff
@@ -38,7 +39,7 @@ export default class WalletNavigation extends Component<Props> {
         <WalletNavButton
           className={category.className}
           label={intl.formatMessage(category.label)}
-          icon={category.icon}
+          icon={category.label.id === WALLET_TRANSACTION_ID && category.isActive ? TransactionImageActive  : category.icon}
```
vantuz-subhuman, line 42: @AhmedIbrahim336 , we have multiple tabs in the interface and all of them have icons that change color, but other tabs don't declare a check like this here. Seems like there must be another solution for it.

Also I see that this code will use the new icon when the tab is active, but when the tab is not active - we will see the old icon. The task is not resolved.

vantuz-subhuman (changes requested):

zuzunker (approved):

zuzunker (commented):
/check

PR #2174, opened 2021-06-12
Release / 4.5.5 (Tokens metadata integration v1)

zuzunker:
Milestone: https://github.com/Emurgo/yoroi-frontend/milestone/9?closed=1

New feature is implemented so that when the wallet history is being restored, for each token that has ever been mentioned - we query the off-chain metadata registry and if some metadata is found there - we store it in the local storage (ought to be changed to indexed DB) and then use it on the sending page to display the name of the token and the correct value, considering the decimals value.

### DIFF
--- a/packages/yoroi-extension/app/api/ada/lib/storage/database/prepackaged/networks.js
+++ b/packages/yoroi-extension/app/api/ada/lib/storage/database/prepackaged/networks.js
@@ -33,6 +33,8 @@ export const networks = Object.freeze({
       WebSocket: environment.isTest()
         ? 'ws://localhost:21000'
         : 'wss://iohk-mainnet.yoroiwallet.com:443',
+      TokenInfoService:
+        'https://tokens.cardano.org',
     },
     BaseConfig: ([
       Object.freeze({
@@ -127,6 +129,8 @@ export const networks = Object.freeze({
       WebSocket: environment.isTest()
         ? 'ws://localhost:21000'
         : 'wss://testnet-backend.yoroiwallet.com:443',
+      TokenInfoService:
+        'https://stage-cdn.yoroiwallet.com',
     },
     BaseConfig: ([
       Object.freeze({
--- a/packages/yoroi-extension/app/api/ada/lib/storage/database/prepackaged/networks.js
+++ b/packages/yoroi-extension/app/api/ada/lib/storage/database/prepackaged/networks.js
@@ -34,7 +34,7 @@ export const networks = Object.freeze({
         ? 'ws://localhost:21000'
         : 'wss://iohk-mainnet.yoroiwallet.com:443',
       TokenInfoService:
-        'https://tokens.cardano.org',
+        'https://cdn.yoroiwallet.com',
     },
     BaseConfig: ([
       Object.freeze({
--- a/packages/yoroi-extension/app/stores/ada/AdaTransactionsStore.js
+++ b/packages/yoroi-extension/app/stores/ada/AdaTransactionsStore.js
@@ -11,7 +11,7 @@ import type { ActionsMap } from '../../actions/index';
 import type { StoresMap } from '../index';
 import WalletTransaction from '../../domain/WalletTransaction';
 import CardanoShelleyTransaction from '../../domain/CardanoShelleyTransaction';
-import type { RemoteTokenInfo } from '../../api/ada/lib/state-fetch/types';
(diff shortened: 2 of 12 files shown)


yushih (approved):
We might want to put the data fetching functions in `api/common` instead of `api/ada` but let's deal with that later with refactorings.

packages/yoroi-extension/app/stores/ada/AdaTransactionsStore.js
```diff
@@ -2,23 +2,58 @@
...
+      const missingMetaTokenIds = [...tokenIds]
+        .filter(tokenId => !localStorage.getItem(`token-metadata-${tokenId}`))
+        .map(id => id.split('.')[0]);
+      const tokenInfo = await stateFetcher.getTokenInfo({
+        network: request.publicDeriver.getParent().getNetworkInfo(),
+        tokenIds: missingMetaTokenIds,
+      });
+      // $FlowFixMe[incompatible-call]
+      Object.entries(tokenInfo).forEach(([tokenId, tokenMeta]: [string, RemoteTokenInfo]) => {
+        localStorage.setItem(`token-metadata-${tokenId}`, JSON.stringify({
```
SebastienGllmt: Presumably your key needs the network id as well to avoid collisions with the testnet.
In general it's usually better to use IndexedDB for this kind of stuff, but I understand the IndexedDB setup for the Yoroi Extension is quite complicated

packages/yoroi-extension/app/api/ada/lib/storage/database/prepackaged/networks.js
```diff
@@ -33,6 +33,8 @@ export const networks = Object.freeze({
       WebSocket: environment.isTest()
         ? 'ws://localhost:21000'
         : 'wss://iohk-mainnet.yoroiwallet.com:443',
+      TokenInfoService:
+        'https://tokens.cardano.org',
```
v-almonacid: We might need to whitelist this in the manifest

packages/yoroi-extension/app/stores/ada/AdaTransactionsStore.js
zuzunker: @SebastienGllmt , good point about the network ID, absolutely! The plan is to move all this into the indexed DB in the next version as we just need to get this working for now, and indexed DB needs a bit more time, yes.

packages/yoroi-extension/app/api/ada/lib/storage/database/prepackaged/networks.js
zuzunker: @v-almonacid , this is a temporary value for draft only, supposed to be replaced by a `yoroiwallet.com` address once we get it set up

zuzunker:
> We might want to put the data fetching functions in `api/common` instead of `api/ada` but let's deal with that later with refactorings.

@yushih, Yes, decided to specialize it for ADA only locally for now, as I am not sure yet if Ergo will have anything similar at all with the decimals being a part of tokens metadata that needs to be fetched. Should be easy to generalize if we are confident it can and should be done for Ergo as well

packages/yoroi-extension/app/stores/ada/AdaTransactionsStore.js
```diff
@@ -11,6 +11,11 @@ import type { StoresMap } from '../index';
 import WalletTransaction from '../../domain/WalletTransaction';
 import CardanoShelleyTransaction from '../../domain/CardanoShelleyTransaction';
 import type { RemoteTokenInfo } from '../../api/ada/lib/state-fetch/types';
+import type { NetworkRow } from '../../api/ada/lib/storage/database/primitives/tables';
+
+function createTokenLocalStorageKey(tokenId: string, network: $ReadOnly<NetworkRow>): string {
+  return `token-metadata-${network.NetworkName}-${tokenId}`;
```
yushih: Maybe better to use `NetworkId` instead of `NetworkName`?

yushih (changes requested):

packages/yoroi-extension/app/stores/ada/AdaTransactionsStore.js
zuzunker: Done 👌

zuzunker (commented):
/check

zuzunker (commented):
/check

PR #2178, opened 2021-06-16
Tab icons css tiny fix

zuzunker:
Updated the CSS for new transactions icon a bit to make it more generic than targeting `stroke` property in all `g` tags


paulclindo (approved):
Nice! I thought that currentcolor should be in camelcase but that works too. Seems that both are fully supported and  `currentColor` is in the Color Module Level 3 spec and `currentcolor` in the Color Module Level 4 spec

PR #2180, opened 2021-06-17
api integration for catalyst round info

nicetech-creator:
Please check if the work is in the right location. It seems un-necessarily complicate.


packages/yoroi-extension/app/api/ada/lib/state-fetch/types.js
```diff
@@ -339,3 +339,20 @@ export type TokenInfoResponse = {|
...
+  +id: number,
+  +registrationStart: string,
+  +registrationEnd: string,
+  +votingStart: string,
+  +votingEnd: string,
+  +votingPowerThreshold: string,
+|};
+
+export type CatalystRoundInfoResponse = {|
+  [key: string]: (CatalystRound | null),
```
zuzunker: Here I think it's better to specifically list the allowed keys, instead of just a generic string key type: `currentFund ` and `nextFund `.

packages/yoroi-extension/app/api/ada/lib/state-fetch/remoteFetcher.js
```diff
@@ -407,4 +410,23 @@ export class RemoteFetcher implements IFetcher {
       return res;
     }, {});
   }
+
+  getCatalystRoundInfo: CatalystRoundInfoRequest => Promise<CatalystRoundInfoResponse> = (body) => {
+    const { BackendService } = body.network.Backend;
+    if (BackendService == null) throw new Error(`${nameof(this.getCatalystRoundInfo)} missing backend url`);
+    return axios(
+      `${BackendService}/api/v0/catalyst/fundInfo`,
```
zuzunker: The `/api` prefix in the endpoint is conditional and only exists on the testnet. Must add additional check to not add it in case it's mainnet. See example in the mobile:

https://github.com/Emurgo/yoroi-mobile/blob/758a030ef584c18cccc266b2b65e152666890bbd/src/api/shelley/api.js#L176-L179

neuodev:
@yushih Please review the code. I also added some comments. please answer them

packages/yoroi-extension/app/api/ada/lib/state-fetch/remoteFetcher.js
```diff
@@ -407,4 +410,21 @@ export class RemoteFetcher implements IFetcher {
...
+
+  getCatalystRoundInfo: CatalystRoundInfoRequest => Promise<CatalystRoundInfoResponse> = async (body) => {
+    const { BackendService } = body.network.Backend;
+    if (BackendService == null) throw new Error(`${nameof(this.getCatalystRoundInfo)} missing backend url`);
+    const isMainnet = /Mainnet/g.test(body.network.NetworkName)
+    const prefix = isMainnet ? '' : 'api/'
+    return await axios(
+      `${BackendService}/${prefix}v0/catalyst/fundInfo`,
+      {
+        method: 'get',
```
neuodev, line 422: the prefix `api/` only exist on `testnet` but when I remove the prefix on mainnet the response is 404 this mean that the route isn't exist 
on testnet `https://testnet-backend.yoroiwallet.com/api/v0/catalyst/fundInfo` => Working
on mainnet `https://iohk-mainnet.yoroiwallet.com/v0/catalyst/fundInfo/` => not working

packages/yoroi-extension/app/api/ada/lib/state-fetch/types.js
```diff
@@ -340,3 +340,24 @@ export type TokenInfoResponse = {|
...
+export type CatalystRound = {|
+  +id: number,
+  +registrationStart: string,
+  +registrationEnd: string,
+  +votingStart: string,
+  +votingEnd: string,
+  +votingPowerThreshold: string,
+|};
+
+export type CatalystRoundInfoResponse = {|
```
neuodev, line 357: The API response has the `currentFund` and the `nextFund` but both of them are optional
if both of them are optional what should be the default. The app complaining this it needs a default values.

packages/yoroi-extension/app/containers/wallet/voting/VotingPage.stories.js
```diff
@@ -67,6 +67,22 @@ const defaultProps: ({|
     delegation: {
       getDelegationRequests: walletLookup([request.wallet]).getDelegation,
     },
+    substores: {
+      ada: {
+        votingStore: {
+          catalystRoundInfo: {
+            currentFund: {
```
neuodev: I added this because `flow`  complaining as the `catalystRounInfo` need a default value

zuzunker:
@AhmedIbrahim336 , resolve conflicts please

packages/yoroi-extension/app/api/ada/lib/state-fetch/types.js
```diff
@@ -340,3 +340,24 @@ export type TokenInfoResponse = {|
...
+  currentFund: CatalystRound,
```
yushih: This needs to be changed to `currentFund?: CatalystRound,` and the case that this field is null must be handled (see the comment is `VotingPage.js`).

packages/yoroi-extension/app/api/ada/lib/state-fetch/batchedFetcher.js
```diff
@@ -422,6 +427,19 @@ export function batchGetAccountState(
     }
   };
 }
+export function batchGetCatalystRoundInfo(
+  getCatalystRoundInfo: CatalystRoundInfoFunc,
+): CatalystRoundInfoFunc {
+  return async function (body: CatalystRoundInfoRequest): any {
```
yushih, line 433: Please use specific type and avoid `any` as much as possible.

packages/yoroi-extension/app/containers/wallet/voting/VotingPage.js
```diff
@@ -107,21 +102,19 @@ export default class VotingPage extends Component<Props> {
...
         selected.getParent().getNetworkInfo().NetworkId === networks.CardanoMainnet.NetworkId &&
         (isEarly || isLate)
       ) {
         return (
-          <RegistrationOver roundNumber={
-              isLate
-                ? roundInfo.nextRound
-                : roundInfo.nextRound - 1
-            }
+          <RegistrationOver roundNumber={roundNumber}
```
yushih: I think we should do this:
```
if `currentFund` is available {
 if it is before `registrationStart` {
  show 'round {currentFund.id} registration starts at {registrationStart}'
 }
 if it is after `registrationEnd` {
  if it is before `votingStart` {
    show "Registration has ended. Voting starts at {votingStart}"
  }
  if it is between `votingStart` and `votingEnd` {
   show "Registration has ended.  Voting ends at  {votingEnd}"
  }
  if it is after `votingEnd` {
    show "Round {currentFund.id} ended.
    if `nextFund` is available {
      show "Round {nextFund.id} starts at {nextFund.registrationStart}"
    }
  }
 }
} else { // `currentFund` is not available
 if nextFund is available {
   show "Round {nextFund.id} starts at {nextFund.registrationStart}"
 } else {
   show a generic "Catalyst Round Info unavailable please check the Catalyst app for more info" page with no specific info
 }
}
``` 
cc @vsubhuman

yushih (changes requested):
Please end each line with a semicolon for consistency.

When you make changes to this PR, please do not rewrite history and `git push -f`. Just add commits. It makes reviewing easier. Thanks.

packages/yoroi-extension/app/api/ada/lib/state-fetch/remoteFetcher.js
yushih: `curl https://iohk-mainnet.yoroiwallet.com/api/v0/catalyst/fundInfo/` works.

packages/yoroi-extension/app/stores/ada/VotingStore.js
```diff
@@ -63,7 +64,7 @@ export default class VotingStore extends Store<StoresMap, ActionsMap> {
   @observable catalystPrivateKey: RustModule.WalletV4.PrivateKey | void;
   @observable pin: Array<number>;
   @observable error: ?LocalizableError;
-
+  @observable catalystRoundInfo: CatalystRoundInfoResponse
```
yushih, line 67: Please change the type to `?CatalystRoundInfoResponse` and handle the null case in `VotingPage.js`. This is necessary because network request is async and `VotingPage.render` may be called before the request returns and this variable is assigned value.

Here flow failing to detect the nullability is a known issue: https://github.com/facebook/flow/issues/650 .

packages/yoroi-extension/app/containers/wallet/voting/VotingPage.js
neuodev: This is what I am talking about
@vsubhuman  Can you please confirm this so I can start

packages/yoroi-extension/app/api/ada/lib/state-fetch/batchedFetcher.js
neuodev: Sorry, I just put it as a placeholder in development. Will update it 👌

yushih (changes requested):

packages/yoroi-extension/app/api/ada/lib/state-fetch/remoteFetcher.js
zuzunker: @AhmedIbrahim336 , it's because the proxy of `BackendService` itself includes the `/api` part for mainnet. So both URLs contain `/api`, but in the code it has to be adapted for the values. We might fix it later, bit not a priority right now

zuzunker:
Replaced by: https://github.com/Emurgo/yoroi-frontend/pull/2296

PR #2181, opened 2021-06-18
Token metadata revamp

yushih:
Originally in transaction-update code, there was commented-out stub that deals with token info.

Previously Ruslan implemented the fetch functions for token metadata and stored the data in local storage as a temporary measure. 

This PR builds on top of Ruslan's fetch function and calls it in the originally designed place and stores the token metadata properly in the storage layer.

A new field `lastUpdatedAt` is added to the token row. its purpose is to distinguish the token data fetched from the network from a placeholder row so that it would be updated when the token metadata become available.

### DIFF
--- a/packages/yoroi-extension/app/api/ada/lib/storage/bridge/updateTransactions.js
+++ b/packages/yoroi-extension/app/api/ada/lib/storage/bridge/updateTransactions.js
@@ -130,6 +130,7 @@ import type {
   RemoteTxState,
   RemoteTransaction,
   RemoteCertificate,
+  TokenInfoFunc,
 } from '../../state-fetch/types';
 import {
   ShelleyCertificateTypes,
@@ -873,6 +874,7 @@ export async function updateTransactions(
   checkAddressesInUse: FilterFunc,
   getTransactionsHistoryForAddresses: HistoryFunc,
   getBestBlock: BestBlockFunc,
+  getTokenInfo: TokenInfoFunc,
 ): Promise<void> {
   const withLevels = asHasLevels<ConceptualWallet>(publicDeriver);
   const derivationTables = withLevels == null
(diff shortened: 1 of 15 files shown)


packages/yoroi-extension/app/api/ada/lib/storage/bridge/updateTransactions.js
```diff
@@ -1930,26 +1936,53 @@ async function genCardanoAssetMap(
...
-  //   assetIds: tokenIds.filter(token => !existingTokens.has(token))
-  // });
 
-  const databaseInsert = tokenIds
-    .filter(token => !existingTokens.has(token))
+  const missingTokenIds = tokenIds.filter(token => !existingTokens.has(token));
+
+  const tokenInfoResponse = await getTokenInfo({
+    network,
+    tokenIds: missingTokenIds.map(id => id.split('.').join(''))
```
v-almonacid, line 1952: I'd just suggest that instead of only querying for missing token Ids, we also query for existing tokens which haven't been updated for some time, as the token metadata can change.

v-almonacid (approved):
LGTM

packages/yoroi-extension/app/api/ada/lib/storage/bridge/updateTransactions.js
yushih: What's the policy we use in mobile? What expiration time do you use?

v-almonacid: In mobile is kind of the opposite: We fetch the server every time a wallet is opened and every time a wallet receives a new transaction, but we should add an expiration date instead.

zuzunker (approved):
Purrrfecto 👌

zuzunker (commented):
/check

zuzunker:
@yushih , Flow was giving some errors, had to fix some types. Check the commit plz when you available: https://github.com/Emurgo/yoroi-frontend/pull/2181/commits/80f40046f7250e400af8e6b4b50b46c518e5999e

Good to go, from my side 👌

PR #2185, opened 2021-06-22
Change the close icon for the verify address dialog

neuodev:
-  Added the new cross icon 
- Fix the alignment with the dialog title
- The old Icon is still thair  ( maybe another component using it )
- the task link on [asana](https://app.asana.com/0/1200458692294040/1200479963350777/f)


yushih (changes requested):
The position of the cross is still too low on my machine.

To align the cross with the title I need to change `top` to `0px` and `margin` of `svg` to `0px`.

But then the background shade when hovering needs fixing.

The old icon can be deleted. There are two ways to verify that it is not used:
1. by doing a `grep -r` search of the code
2. by observing that webpack doesn't complain after deleting it.

neuodev:
> The position of the cross is still too low on my machine.
> 
> To align the cross with the title I need to change `top` to `0px` and `margin` of `svg` to `0px`.
> 
> But then the background shade when hovering needs fixing.
> 
> The old icon can be deleted. There are two ways to verify that it is not used:
> 
> 1. by doing a `grep -r` search of the code
> 2. by observing that webpack doesn't complain after deleting it.

1. Deleted the Icon 
2. you can't make the margin of the svg 0px the close button should be moved up as a single component. So what I did. I applied negative margin on the button so it can be up and align with the title

yushih (approved):
It works. Thanks!

zuzunker (commented):
/check

PR #2187, opened 2021-06-23
Add success dialog after successful transaction

neuodev:
- Task link on [asan](https://app.asana.com/0/1200458692294040/1200479963350806/f)
- The current implementation includes redirecting after making a successful transaction. I moved the redirection logic into the new dialog as it will be directly after closing the dialog to the transaction screen
- it is better to test yourself before start reviewing **check out my branch and make a test transaction**


yushih (changes requested):
1. Please do not simply add a state to `WalletSendPage`. As a convention of this whole project, none of the containers has a state. Instead, states are managed by store classes. Please see how `WalletSendConfirmationDialog` is opened by triggering the `dialogs.push` action and follow the pattern (FYI, the dialog state is managed by `UiDialogsStore`).
2. A new property `openTransactionDialog` is added to `WalletSendConfirmationDialog` without changing the prop type definition. This breaks flow. But the property should not be added in the first place because of point 1. Anyway, please make sure `npm run flow` passes.
3. Please make sure the hardware wallet send process is changed too.
4. Please update the e2e test too.
5. (Bonus) Please take care of storybook.

yushih (changes requested):
There are some problems:
1. You can't simply remove the line that changes route to the transactions page in `AdaMnemonicSendStore._sendMoney`, because this function is used elsewhere other than sending a tx, and the caller there actually expects that behavior. You can check this with a `grep -r wallets.sendMoney.trigger` over the code base. This is fixed by 4c5a3f4da4a5ed1e5b6280e87b87db16b0be469d.

2. Need to change to hardware wallet send process too. This is done by ce9689b4cfbe9462f18861fb38f018102e43183e.

3. The way i18n messages are added must follow a pattern so that they can be translated. Please see commit cae913b02aac51dde4ebca3939802930472e280e.

4. Please change e2e tests accrodingly, for example, in `transactions.feature`, tests like this:
```
    And I submit the wallet send form
    Then I should see the summary screen
```
should be changed to something like this:
```
    And I submit the wallet send form
    Then I should see the "successfully sent" page
    And I click the "transaction page" button
    Then I should see ...
```

yushih (changes requested):
Please also update `hardware.feature`. Thanks.

neuodev:
> Please also update `hardware.feature`. Thanks.

@yushih I have a problem running testing can you please update it.

yushih (approved):

packages/yoroi-extension/app/containers/wallet/WalletSendPage.js
```diff
@@ -91,16 +104,11 @@ export default class WalletSendPage extends Component<InjectedOrGenerated<Genera
 
     const { transactionBuilderStore } = this.generated.stores;
 
-    const { uiDialogs, profile, } = this.generated.stores;
+    const { profile, } = this.generated.stores;
```
zuzunker: Why this piece about blocking during a pending tx got removed? I don't see a mention of that in the PR description or in the task 🤔 

image

zuzunker (changes requested):

packages/yoroi-extension/app/containers/wallet/WalletSendPage.js
zuzunker: In discussion with @yushih now

zuzunker (approved):

PR #2213, opened 2021-07-05
Cardano dApp-connector : Add read-access functions for cardano

neuodev:
[Task on Asana](https://app.asana.com/0/1200458692294040/1200543243099078/f)


packages/yoroi-ergo-connector/src/inject.js
```diff
@@ -28,10 +28,28 @@ function ergo_check_read_access() {
...
+        return 'Under development'
+    } else {
+        return Promise.resolve(false);
+    }
+}
 `
 
 // client-facing ergo object API
-const apiInject = `
+const ergoApiInject = `
```
zuzunker, line 52: Good call to rename this 👍

packages/yoroi-ergo-connector/src/inject.js
```diff
@@ -108,6 +126,24 @@ class ErgoAPI {
...
+    constructor(){
+        this.getInitTimestamp()
+    }
+
+    getInitTimestamp(){
+        return Date.now()
+    }
+}
+
+const cardano = Object.freeze(new CardanoAPI)
```
zuzunker: @AhmedIbrahim336, missing brackets when calling constructor: `new CardanoAPI()`

packages/yoroi-ergo-connector/src/inject.js
```diff
@@ -108,6 +126,24 @@ class ErgoAPI {
...
+// @todo add cardano rpc resolver when it is done
+
+class CardanoAPI {
+    constructor(){
+        this.getInitTimestamp()
+    }
+
+    getInitTimestamp(){
+        return Date.now()
+    }
```
zuzunker, line 141: @AhmedIbrahim336 This function should always return the same date, when the class instance was created. You need to do something like `this.initTimestamp = Date.now()` in the constructor and then `return this.initTimestamp` inside the function.

packages/yoroi-ergo-connector/src/inject.js
```diff
@@ -191,6 +228,28 @@ function createYoroiPort() {
                 type: "connector_connected",
                 success: message.success
             }, location.origin);
+        }else if (message.type == "yoroi_connect_response/cardano") {
```
zuzunker:
```suggestion
        } else if (message.type == "yoroi_connect_response/cardano") {
```

packages/yoroi-ergo-connector/src/inject.js
```diff
@@ -191,6 +228,28 @@ function createYoroiPort() {
...
+                    // inject full API here
+                    if (injectIntoPage(cardanoApiInject)) {
+                        cardanoApiInjected = true;
+                    } else {
+                        console.error()
+                        window.postMessage({
+                            type: "connector_connected",
+                            err: {
+                                code: API_INTERNAL_ERROR,
+                                info: "failed to inject cardano API"
```
zuzunker:
```suggestion
                                info: "failed to inject Cardano API"
```

zuzunker (changes requested):

packages/yoroi-ergo-connector/src/inject.js
rooooooooob: What are we using the init timestamp for? It's not in the spec. Is this something to show on the UI side? Do we want to do this for Ergo too?

packages/yoroi-ergo-connector/src/inject.js
```diff
@@ -108,6 +126,24 @@ class ErgoAPI {
 
 const ergo = Object.freeze(new ErgoAPI());
 `
+
+const cardanoApiInject = `
+
+// @todo add cardano rpc resolver when it is done
+
```
rooooooooob, line 133: Do we not want to reuse the RPC relay instead? They're both going to the same extension and for RPCs should be basically the same and afaik there's no naming conflicts (we can always add in a type: ' cardano' | 'ergo'  too which could be useful regardless).

Although this would also depend how we're handling it on the Yoroi side. There should be a LOT of things in common between the two, and most of the code here + the RPC stuff on both sides is pretty independent from the actual API that's being implemented across the RPCs.

It's worth noting that having the RPC code being injected/existing != you are connected. You could have been disconnected at some point anyway so the RPC already handles the case where you no longer have permissions to access the API so that shouldn't be an issue reusing it. It's all async so it shouldn't block interaction anywhere between the two.

packages/yoroi-ergo-connector/src/inject.js
neuodev: it just a place holder for now I think. @vantuz-subhuman  I think you need to answer this

zuzunker: Yes

packages/yoroi-ergo-connector/src/inject.js
zuzunker: Yes, this is definetely what we would want to achieve, either right away from the beginning or as a reworking at a later stage. Thank you, for the details, @rooooooooob ! 🙌

packages/yoroi-ergo-connector/src/inject.js
```diff
@@ -108,6 +126,25 @@ class ErgoAPI {
...
+class CardanoAPI {
+    constructor(){
+        this.initTimestamp = Date.now()
+        this.getInitTimestamp()
```
zuzunker: No reason to call this function in the constructor, it's just a getter with no side-effects

```suggestion
```

zuzunker (approved):

zuzunker (commented):
/check

rooooooooob (approved):
LGTM for now for a start, but we should try and share as much of the core RPC framework as reasonable in the future PRs when it's implemented since it should be the same in both specs.

PR #2214, opened 2021-07-06
Fix Ergo connector crash

yushih:
The Ergo connector app (`/main_window_ergo.html`) has a different action map than the main Yoroi extension app. Accessing to `actions.wallets.setActiveWallet` in the Ergo connector app leads to crash. It is fixed by wrapping in a check.


zuzunker (approved):

rooooooooob (approved):
LGTM, thanks Yushi! I did a similar fix locally and had no problems with the connector main flow (connect + sign tx + send)

zuzunker (commented):
/check

PR #2223, opened 2021-07-08
Remove the complexity page 

neuodev:
[Task link](https://app.asana.com/0/1200458692294040/1200479963350821/f)


packages/yoroi-extension/app/Routes.js
```diff
@@ -220,11 +213,6 @@ const SettingsSubpages = (stores, actions) => (
       path={ROUTES.SETTINGS.SUPPORT}
       component={(props) => <SupportSettingsPage {...props} stores={stores} actions={actions} />}
     />
-    <Route
-      exact
-      path={ROUTES.SETTINGS.LEVEL_OF_COMPLEXITY}
-      component={(props) => <ComplexityLevelSettingsPage {...props} stores={stores} actions={actions} />}
-    />
```
zuzunker: @AhmedIbrahim336 , I have confirmed it when you asked if we also need to remove the settings complexity page - this is my bad, but this was incorrect. We need only to remove the initial complexity screen in the first lauch process, but leave the settings-complexity page so users can change it later, if needed.

Let's plz revert the changes on the settings page and bring it back.

zuzunker (changes requested):

packages/yoroi-extension/app/Routes.js
neuodev: Okay

packages/yoroi-extension/app/containers/profile/TermsOfUsePage.js
```diff
@@ -34,6 +36,11 @@ export default class TermsOfUsePage extends Component<InjectedOrGenerated<Genera
     intl: intlShape.isRequired,
   };
 
+  onSubmit: ((acceptance: any) => void) = (acceptance: any) => {
+    this.generated.actions.profile.acceptTermsOfUse.trigger(acceptance)
+    this.generated.actions.profile.selectComplexityLevel.trigger(ComplexityLevels.Simple)
+  }
```
zuzunker: Nice!

zuzunker (approved):

yushih (changes requested):
I think this solution is a bit problematic in that:
1. It couples the terms of use page with the knowledge of the complexity page, while these two should be independent. I think a better solution would be to change the complexity page step action in `ProfileStore.SETUP_STEPS` from going to the complexity page route to just directly settiing the complex level to simple.
2. Notice how the e2e tests expect the complexity page: https://github.com/Emurgo/yoroi-frontend/blob/develop/packages/yoroi-extension/features/step_definitions/common-steps.js#L250. You need to remove these lines now that there is no complexity page. But then there's another problem: originally the e2e tests set the complexity level to the most complex one but now it defaults to simple. I think this might break the following tests. Please check and workaround (maybe by defaulting the complexity level for test build to complex?).

neuodev:
@yushih  I removed the `complexity page` from testing and run it. And it's working.
So it will work like this. The `SIMPLE` mode will be the default this will work for every feature that exist on both the SIMPLE and COMPLEX mode. I asked Ruslan about the difference. he said their is no big difference.
For features that exist on COMPLEX but not on SIMPLE mode we can add one extra step to go to the settings and change it to complex and then complete the testing normally. 
So can you identify which features exit on COMPLEX mode and not on the SIMPLE one?

packages/yoroi-extension/app/stores/toplevel/ProfileStore.js
```diff
@@ -48,6 +49,7 @@ export default class ProfileStore extends BaseProfileStore<StoresMap, ActionsMap
         if (this.stores.app.currentRoute === route) {
           return;
         }
+        this.actions.profile.selectComplexityLevel.trigger(ComplexityLevels.Simple);
         this.actions.router.goToRoute.trigger({ route });
```
yushih: Remove this line : `this.actions.router.goToRoute.trigger({ route });` ?

zuzunker (approved):
@yushih , double-check plz

yushih:
@AhmedIbrahim336 please fix e2e tests.There are some errors but I think they should be obvious.

packages/yoroi-extension/features/memo.feature
```diff
@@ -37,6 +37,6 @@ Feature: Memos
       | password   |
       | asdfasdfasdf |
     And I submit the wallet send form
+    Then I should see the successfully sent page
+    And I click the transaction page button
     Then I should see the summary screen
-    And I expand the top transaction
-    Then The memo content says "my awesome memo"
```
yushih, line 42: Please do not delete these two lines. Because the purpose of this test is to check that these lines are true.

yushih (changes requested):

packages/yoroi-extension/features/memo.feature
neuodev: Okay, I will get them back. And run the test again

yushih (approved):

PR #2224, opened 2021-07-08
Add cardano RPC

neuodev:
[Task Link](https://app.asana.com/0/1200458692294040/1200567965519644/f)

### DIFF
--- a/packages/yoroi-ergo-connector/src/inject.js
+++ b/packages/yoroi-ergo-connector/src/inject.js
@@ -1,13 +1,14 @@
 // sets up RPC communication with the connector + access check/request functions
 const initialInject = `
-var ergoConnectRequests = [];
+var connectRequests = [];
 
+var cardanoConnectRequests = [];
 window.addEventListener("message", function(event) {
     if (event.data.type == "connector_connected") {
         if (event.data.err !== undefined) {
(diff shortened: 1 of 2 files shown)


packages/yoroi-ergo-connector/src/inject.js
```diff
@@ -1,13 +1,14 @@
 // sets up RPC communication with the connector + access check/request functions
 const initialInject = `
-var ergoConnectRequests = [];
+var connectRequests = [];
 
+var cardanoConnectRequests = [];
```
vantuz-subhuman, line 5: @AhmedIbrahim336 , I think you don't need this `cardanoConnectRequests` at all

vantuz-subhuman: Everything else is good 👌

zuzunker:
```suggestion
```

zuzunker (approved):

cvetelinandreevdreamix:
Hey @vsubhuman,

We have an existing iOS app which we want to integrate with a Cardano wallet. Basically we want to open a wallet app to sign a transaction we've prepared in advance. We'll use a testnet initially.

Is that possible now?

Will be there nightly builds for the iOS app we can test (TestFlight)?

Thank you!

PR #2228, opened 2021-07-09
Ruslan / Revert my own incorrect flow fix

zuzunker:
There was a PR from @yushih : https://github.com/Emurgo/yoroi-frontend/pull/2214

Which I have ruined with an incorrect Flow fix.
Reverting my incorrect changes now.


zuzunker (commented):
/check

yushih (approved):
LGTM. This does lead to Flow warning. But the runtime behavior is correct - `this.actions.wallets` is indeed undefined for Ergo connector app. For now let's ignore the warning until we fix the root cause.

PR #2229, opened 2021-07-13
Add the get_balance function to the CardanoAPI class

neuodev:
I don't think we need to update any thing from the Yoroi side. This can be handled in the same manner.
As the flow go like this 
 1. The `cardano_rpc_call` sends message with the type `connector_rpc_request` 
 2. In the Yoroi side we handle this message type
 3.  Start checking for the function name in this situation it is `get_balnce`  
 4. Validate the input 
 5. Query the local storage for with the token id this can be `ERG` or `ADA`.
  I thing this function can work for both `ERG` and `ADA`


zuzunker:
in testing

zuzunker (approved):

PR #2230, opened 2021-07-13
bump chromedriver version to 91

yushih:
The original chromedriver version 87 requires Chrome version 87, which is not available for Mac with M1. Hence bump it to version 91.

Also, fix a problem that fails some e2e tests.


neuodev:
@yushih @vantuz-subhuman I test it. Can we merge it quickly so I can use it with the other tasks.

zuzunker (approved):

PR #2232, opened 2021-07-13
Ergo Connector: Fix for large token amounts

### LINKED ISSUE #2209: Ergo connector: Large (>Number.MAX_SAFE_INTEGER) quantities in txs render them unsignable.
Hey guys. I want to use `sign_tx` method for send 2 tokens in one TX. After I enter the spending password, nothing happens. Promise just has 'pending' state.

Screenshot 2021-07-03 at 16 49 39

If I reject this TX, I can catch 'reject' state.

> 1. State after opened yoroi-dapp-connector window.
> image
> 2. State after close yoroi-dapp-connector window.
> image
> 

>  1. State after opened yoroi-dapp-connector window.
> image
>  2. State after entered and confirmed spending password.
> image
> image
>

rooooooooob:
The previous definition of the TxJson we had amount/value as just
number, but now it accepts number/string so we can fully support values
too big for a 53-bit JS number. The dApp example was updated to fully
use BigInt as well.

Fixes #2209

### DIFF
--- a/packages/yoroi-ergo-connector/example/index.js
+++ b/packages/yoroi-ergo-connector/example/index.js
@@ -58,25 +58,23 @@ function initDapp() {
                     status.innerText = "Creating transaction";
                     const donationAddr = "9hD2Cw6yQL6zzrw3TFgKdwFkBdDdU3ro1xRFmjouDw4NYS2S5RD";
                     const creationHeight = 398959;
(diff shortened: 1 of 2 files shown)


zuzunker (approved):

zuzunker (commented):
/check

PR #2237, opened 2021-07-16
dapp-connector: support json-bigint

zuzunker:
## Intro
After investigating the issue https://github.com/Emurgo/yoroi-frontend/issues/2209 and removing the stub that was parsing number strings, and after some additional testing after that it turned out that sigma-rust does not support numbers being passed as strings after all, but it does parse u64 numbers safely, unlike JS by default.

This meant that Yoroi now does not block or modifies the passed numbers in any way, and the sigma-rust is able to receive and parse JSON with u64 numbers, but that JS provides no safe way for dapp devs to pass some of the number values to Yoroi, unless Yoroi would know and explicitly map or convert every single possible value, which we want to avoid to not block the support of any new field by having first to add a new mapper on the Yoroi side.

The solution is simply to add the json-bigint library that supports parsing and serialising JSON with BigInt values (both native and `bignumber.js`) as proper JSON numbers. So dapp devs just need to make sure they are using safe bigint numbers in objects sent to the dapp-connector API and the Yoroi will make sure to serialize them with the proper bigint support.

👉  Thanks to @oskin1 for contributions with this issue!

## Changes
Added json-biging and replaced built-in `JSON` with the library in ergo-connector related files


packages/yoroi-extension/chrome/extension/ergo-connector/types.js
```diff
@@ -83,16 +85,16 @@ export function asBox(
 
       {
         // we just want to validate that the boxID matches the content
-        const wasmBox = wasmInstance.ErgoBox.from_json(JSON.stringify(boxJson));
+        const wasmBox = wasmInstance.ErgoBox.from_json(JSONBigInt.stringify(boxJson));
```
rooooooooob, line 88: Does the sigma-rust JSON parser interpret this correctly? From what I've read JSON doesn't specify the range but a lot of parsers will assume only up to a regular JS double. Just to be sure since I thought my change worked before, but somewhere along the line it was getting truncated to JS double range and losing a tiny bit of precision: `1234567890123456789` -> `1234567890123456800`

This seems to be the only place where parsing is done of the JSON (albeit indirectly inside of sigma-rust), and the json-bigint uses bignumber.js instead of bigint like us but I guess that's not relevant since we never call `JSONbig.parse()` anywhere.

rooooooooob: There is an option to get bigint instead of bignumber.js now that it's natively supported but I think we just read it all per-field into bigints throughout the connector code when we need to operate on it rather than converting the whole thing.

rooooooooob:
>The solution is simply to add the json-bigint library that supports parsing and serialising JSON with BigInt values (both native and bignumber.js) as proper JSON numbers. So dapp devs just need to make sure they are using safe bigint numbers in objects sent to the dapp-connector API and the Yoroi will make sure to serialize them with the proper bigint support.

But are we keeping the *ability* to pass things as raw strings for value? Or are we just passing in json integers which are high-precision and out of the normal JS value then  parsing them via this lib only? We' should update the spec to clarify these things.

packages/yoroi-extension/chrome/extension/ergo-connector/api.js
```diff
@@ -227,7 +229,7 @@ export async function connectorSignTx(
   await RustModule.load();
   let wasmTx;
   try {
-    wasmTx = RustModule.SigmaRust.UnsignedTransaction.from_json(JSON.stringify(tx));
+    wasmTx = RustModule.SigmaRust.UnsignedTransaction.from_json(JSONBigInt.stringify(tx));
```
rooooooooob, line 232: If we want to keep the string values (which is what the standard specifies) we're going to want to handle that here (+ the other from_json place for boxes). Do we convert the string/number to a biginteger.js (or use the BigInt option in the lib) before calling `JSONBigInt.stringify()` on it?

oskin1: @rooooooooob I guess `number` is not really a safe option for dApps to work with tokens. As many of them have 8-9 decimals `MaxSafeNumber` often gets violated. And what JS does in that case is implicitly lowering the precision which is the worst thing in regards of working with tokens. Many dApp devs can simply forget to always check for MaxSafeNumber overflow, so I'd propose to fix the standard and step away from `number | string`.

oskin1: The best option is to replace it with `bigint | string`, I guess.

oskin1: Or, if `bigint | string` requires too heavy refactoring we can stick to `string` only.

oskin1: We will add string support to `sigma-rust` JSON decoders on monday.

packages/yoroi-extension/chrome/extension/ergo-connector/types.js
zuzunker: @rooooooooob , I had to add parsing after all, yes, cuz I have been missing that built-in JSON stringification is also used to send messages thru the socket. Now these messages are being biging-stringified before sending and then bigint-parsed on the other side. Inside Yoroi itself `json-bigint` is configured to use native bigint wherever parsing is used, so we can make the typecheck for `bigint`.

@oskin1, when receiving messages back from Yoroi I have used a browserified version of the `json-bigint` without the native setting being set, so right now it will give out `bignumber.js` bigints, independent of what have been sent in (the PR is in a draft mode for now). There's a decision we need to make where we either will always return `bignumber.js` or always return native, or do some of it by default and then  we add some kind of a setting on the `ergo` object where the dev could do `ergo.setUsingNativeBigIntForParsing(true)`.

zuzunker:
The issues is closing as sigma-rust will be updated to pass all numeric values as strings always, to bypass unsafe JS numbers completely.

PR #2242, opened 2021-07-19
Create Cardano-example in the Dapp-connector

neuodev:
- [Task on Asana](https://app.asana.com/0/1200458692294040/1200587769143229/f)
- First I faced some problems as the `cardano_request_read_acces` was not working. As we implement it in the wrong way so I fix it first (The fixation  in this repo in the `inject.js` file)
- Duplicate The example for `ergo` and rename it to `example-ergo` and `example-cardano`
- The example-ergo is the same as the prev example 
- the new one is `example-cardano` it only includes a button and when you click it will connect to yoroi and ask you to choose a wallet and if it accepted you will see an alert that you now have access.

### Notes
1. This repo doesn't include testing for the get_blance function for Cardano as the repo isn't accepted yet
2. when you exclude the `cardano_request_read_access` you will see a pop from the yoro-ergo-connector extension but it only displays the ergo wallets!!!!. So we need to configure it to display both `ergo` and `cardano` depends on the type of the connection


zuzunker (commented):
/check

yushih (approved):

zuzunker (commented):
/check

PR #2244, opened 2021-07-19
dapp-connector inject some fixes

zuzunker:
Removed alert. Fixed weak comparisons. Fixed cardano connection request.

### DIFF
--- a/packages/yoroi-ergo-connector/src/inject.js
+++ b/packages/yoroi-ergo-connector/src/inject.js
@@ -15,7 +15,7 @@ window.addEventListener("message", function(event) {
 function ergo_request_read_access() {
     return new Promise(function(resolve, reject) {
         window.postMessage({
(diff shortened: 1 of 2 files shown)


packages/yoroi-ergo-connector/src/inject.js
```diff
@@ -32,7 +32,7 @@ function ergo_check_read_access() {
 function cardano_request_read_access() {
     return new Promise(function(resolve, reject) {
         window.postMessage({
-            type: "yoroi_connect_request/cardano",
+            type: "connector_connect_request/cardano",
```
rooooooooob, line 35: It's probably not too important but `window.postMessage` is shared amongst all wallets trying to implement this so we might not want too generic of a message to send, maybe we should have both have "yoroi" somewhere in the name to be safer?

PR #2245, opened 2021-07-19
ergo-lib updated to 0.14.0-alpha.

oskin1:


zuzunker (approved):

rooooooooob (commented):
We should update the flow types for any API changes made from sigma-rust 0.13.0 to 0.14.0. This includes the JSON as we have have that flow-typed within `packages/yoroi-extension/flow-typed/npm/ergo-lib-wasm-browser_vx.x.x.js`. Are there other API changes within 0.14.0?

PR #2246, opened 2021-07-20
Ergo number Values to be encoded as strings

zuzunker:
Built from this contribution with the new sigma-rust version: https://github.com/Emurgo/yoroi-frontend/pull/2245

Ever converting any Ergo utxo-value or an asset-amount to a JS number causes silent loss of value, cuz Ergo values are i64 and JS numbers are i53.

This PR changes the internals of Yoroi and the Dapp-connector API to always exchange utxo-values and asset-amounts as strings. This means Yoroi will expect to receive string values through the dapp-connector and will return string values though the dapp-connector API (e.g. utxos for addresses will have values as strings and asset amounts as strings). Dapp developers must update to send strings and expect strings and are encouraged to use libraries like `BigNumber.js` or native BigInts to work with potentially huge values safely.


zuzunker (commented):
/check

packages/yoroi-ergo-connector/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "yoroi-ergo-connector",
-  "version": "0.4.0",
+  "version": "0.4.2",
```
rooooooooob, line 3: Why are we updating this? Is it to make up for the change from the last PR with the logging changes?

rooooooooob (approved):
I'd be extra careful about not accidentally impacting non-connector Ergo functionality (tx sending within extension UI, syncing, etc). I think it should be fine though if the flow types agree and those have been tested.

packages/yoroi-ergo-connector/package.json
zuzunker: @rooooooooob , yeah, there were a separate PR where `0.4.1` have been released, but the PR never got merged. So I have release `0.4.2` from this PR now.

zuzunker (commented):
/check

packages/yoroi-extension/app/api/ergo/lib/state-fetch/remoteFetcher.js
```diff
@@ -1,45 +1,120 @@
...
+      },
+      transformResponse: resp => {
+        debug('RSPRAW', resp);
+        return JSONBigInt.parse(resp);
+      },
+      transformRequest: data => {
+        if (!data) {
+          return data;
+        }
+        const res = JSONBigInt.stringify(data);
```
yushih, line 93: Since you pass the request as a string to axios, axios defaults the "content-type" header to "application/x-www-form-urlencoded", instead of "application/json" as desired. I find that adding this line at the global scope fixes it:
```
axios.defaults.headers.post['Content-Type'] = 'application/json';
```

packages/yoroi-extension/app/api/ergo/lib/state-fetch/remoteFetcher.js
```diff
@@ -60,218 +135,151 @@ export class RemoteFetcher implements IFetcher {
...
-          throw new GetTxHistoryForAddressesApiError();
-        });
+          return new GetTxHistoryForAddressesApiError();
+        },
+        responseMapper: (data: HistoryResponse) => {
+          for (const datum of data) {
+            datum.outputs.forEach(o => fixUtxoToStringValues(o));
+            // TODO: the inputs fix might potentially be removed
+            // https://github.com/ergoplatform/explorer-backend/issues/92
+            datum.inputs.forEach(i => { i.assets = i.assets ?? [] })
```
yushih: Just to confirm we don't need to apply `fixUtxoToStringValues` to `inputs` ?

yushih (commented):
Not to say that something is wrong, but things can be simplified a little if we enforce that between Yoroi-front and Yoroi-ergo-backend, numbers are always represented as strings. This way there is no need for `fixUtxoToStringValues` and Yoroi-frontend doesn't need `json-bigint`. Only when interfacing with the ergo explorer, Yoroi-ergo-backend employs `json-bigint`. Maybe this can be a future refactoring goal.

packages/yoroi-extension/app/api/ergo/lib/state-fetch/remoteFetcher.js
zuzunker: Good point! I have missed that returned inputs contain values and assets as well, thought they should be only references to utxos.

Adding new commit for this.

packages/yoroi-extension/app/api/ergo/lib/state-fetch/remoteFetcher.js
zuzunker: That's a good point! But I am kinda anxious changing the axios defaults as I am not sure if it will not accidentally affect some other parts of the system that might incidentally or intentionally rely on the axios type detecting, or something like that.

I have found that we can set the content type for that specific request every time in the same request-mapper tho, so fixed in https://github.com/Emurgo/yoroi-frontend/pull/2246/commits/64b72daf5aec865594343584cd9c1b2093c4ec07

zuzunker:
> but things can be simplified a little if we enforce that between Yoroi-front and Yoroi-ergo-backend, numbers are always represented as strings. This way there is no need for `fixUtxoToStringValues` and Yoroi-frontend doesn't need `json-bigint`. Only when interfacing with the ergo explorer, Yoroi-ergo-backend employs `json-bigint`. Maybe this can be a future refactoring goal.

That could be a good improvement in the future, that's true. For this time the main goal was to try and fix the number-handling on the backend, while keeping it compatible with the current versions of Yoroi in production. In the future when we will have the new Yoroi versions in production be compatible with the strings - we can update the backend to also return strings and it should not break Yoroi then at all, as parsing with the json-bigint will not affect strings in any way, and then `fixUtxoToStringValues` will just attempt to convert strings to strings and do nothing. And then we can delete the bigint related code in the next version.

zuzunker (commented):
/check

packages/yoroi-extension/app/api/ergo/lib/state-fetch/remoteFetcher.js
yushih: make sense!

yushih (approved):

Issue #2231, opened 2021-07-13
Yoroi Nightly crash on loading wallet


hodlonaut:
MacOS 11.4, Chrome 91.0.4427

Log file attached
[yoroiNightlyCrash.log](https://github.com/Emurgo/yoroi-frontend/files/6812552/yoroiNightlyCrash.log)

zuzunker:
Thank you, @hodlonaut! Was this with the nightly version `4.5.7002`?

hodlonaut:
> Thank you, @hodlonaut! Was this with the nightly version `4.5.7002`?

Yep that's right, 4.5.7002.

yushih:
Hi, @hodlonaut ! Thank you very much for reporting the issue. Do you mean the Chrome tab has crashed? And the log file is the Chrome debug log (https://support.google.com/chrome/a/answer/6271282)?

hodlonaut:
Yes, the extension crashed (over and over and over, presumably can't cope with my testnet wallet) and I was presented with an option to reload it and not much in the way of 'see details of the crash', so I launched Chrome browser with debug log command line parameter and grabbed the information from that file that might or might not be useful.

yushih:
@hodlonaut  sorry for what has happened and thank you very much for reporting the problem. However, the Chrome debug log didn't provide much insight into what Yoroi was doing that could cause Chrome to crash. I'm no expert in debugging Chrome but I'll keep an eye on the issue. Currently, I hope a newer version of Chrome would fix this. If you need to operate your wallet now, could you please try Yoroi for Firefox?

hodlonaut:
Hi, what's the link for the Yoroi Nightly download/installation page for Firefox?

yushih:
We only publish nightly for Chrome.

hodlonaut:
So given my issue occurred on testnet, I cannot 'please try Yoroi for Firefox' suggestion of yours :)
You're welcome to provide me with an email address that I can send the testnet mnemonic of this wallet to, so you can hopefully reproduce the crash in-house and hopefully fix?

yushih:
Could you please encrypt your testnet mnemonic with my public key and post the cipher here? Please run this in cli:
```
[...]

PR #2262, opened 2021-07-30
Attempt to fix a crash scenario by optimizing memory usage

yushih:
I believe #2231 is due to OOM. By profiling memory allocation with Chrome dev tools, I found the hot spot to be:
https://github.com/Emurgo/yoroi-frontend/blob/develop/packages/yoroi-extension/app/api/ada/lib/storage/database/transactionModels/multipart/api/read.js#L209
There are over 800 txs in this case and each time the map functions for `request.txs.map` runs, the `tokens.map` expression consumes about 9MB of memory, eventually draining memory. But this is a constant expression over the loop and its return value is a read-only object (guaranteed by its type definition). So I lift the expression out of the map loop.

In my env, this fixes the crash of the wallet in question.


zuzunker:
Damn great catch, @yushih !

zuzunker (approved):

PR #2287, opened 2021-08-02
Fix Firefox logo in README

SebastienGllmt:
The logos used for the various projects in the README are taken from Twitter (was kind of a lazy solution to getting the browser icons). It means whenever they update their icon on Twitter (because for some reasons browsers update their logos every few months), it breaks the link in the README

I noticed the image was broken and it's an easy fix so I just made the PR 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 | Firefox | Chrome | Edge |
 |---|----|----|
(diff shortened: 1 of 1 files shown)


zuzunker (approved):
Thanks, Seba!

PR #2288, opened 2021-08-02
Changing dependabot to a specific whitelist of important stuff

zuzunker:
Dependabot is spanny as hell with its PRs for minor libraries. All we care about is updated for important libs or security warnings, that's it. So trying to change the dependabot config to a whitelist.


wolverineks:
how will this affect keeping the dependencies up to date?

wolverineks:
whats the current status of `npm outdated`?

wolverineks:
package.json seems...decent

although...
dropbox is 6 major versions behind
crypto-random-string is 1 major behind
flow is 14 versions behind

zuzunker:
> package.json seems...decent
> 
> although...
> dropbox is 6 major versions behind
> crypto-random-string is 1 major behind
> flow is 14 versions behind

@wolverineks, Yeah, for the dropbox specifically dependabot created a PR with a bump already and I didn't close it, marked as "in-investigation", we need to look into this one, weird that we have such an old version for it. Need to look into the reasoning and when it was first added.

Flow is a good candiate for an update, but needs to be a separate PR, might require some fixes to go along with it.

For `crypto-random-string` tho - a good question might be is whether we need this change or not, really. if we are using a specific version of the library and it works and suffices our needs and is not involved in any security-concerned functionality - I'd say we don't care that much if it gets new major versions or not. It's worth checking what was actually changed on their side, but they might have just added some new functionality or changed their API or the implementation, which is of no concern to us really. I don't think there needs to be this default mode of keeping absolutely all used libraries up to the latest version, as long as we are using a version that works and gives us what we need - we should be satisfied with it.

The problem with the dependabot is that for any major-version library update I'd rather have it done as a human input - e.g. someone checked the release notes from the library, made sure we need it, updated the version and then resolved any differences and made any additional updated within the same branch. Which would make sense for whenever we know of some security concern or some really important critical libraries we are using.

While dependabot just opens 99 PRs for a load of shit that we don't care about at all. A random unnecessary update of a JS library might cause way more surprising problems than it would solve. For all the version we are using right now - at least we know everything is working more or less as expected.

The idea with the whitelist was to maybe have it more as a notification mechanism where an open PR would notify us that we need to assign someone to work on an update more directly. But it's not perfect as well. Ideally I'd cancel dependabot compeltely and make some kind of a monitoring that would run the `outdated` and just ping us with a single list of what new version are available, whenever something new happens. We might work on that later. For now it's just a way to clean the meaningless clutter of PRs.

PR #2289, opened 2021-08-03
Fix numberOfDecimals set to NaN

SebastienGllmt:
`parseEIP0004Data` returns `null | string`

`parseInt(null, 10)` gives `NaN`

This means that incorrectly constructed decimal fields is causing Yoroi to set NaN to the number of decimal places which causes issues down the line.

On a side note: I noticed the `parseEIP0004Data` function is just a better implementation of the code I wrote a while ago to implement EIP-4 for testing tokens so I just unified these two as well


zuzunker:
Thanks, Seba!

zuzunker (approved):

zuzunker (commented):
/check

PR #2290, opened 2021-08-03
Update Send all options 

neuodev:


neuodev:
@yushih  There is a problem with the e2e test. even if I select the dropdown but it doesn't open

packages/yoroi-extension/features/support/webdriver.js
```diff
@@ -247,6 +247,13 @@ function CustomWorld(cmdInput: WorldInput) {
       });
     }, index);
   };
+
+  this.clickElementByQuery = async (query) => {
+    await this.driver.executeScript(
+      `document.querySelector('${query}').click()`,
+    );
+  };
```
neuodev, line 255: Util we can use it whenever the normal `this.click` doesn't work

zuzunker (commented):
/check

zuzunker (approved):
Works great!

PR #2300, opened 2021-08-11
Remove toggleSendAll

neuodev:
In the previous  implementation we had a checkbox it made sense to use `toggleSendAll` now we have a dropdown so we are replacing this now with `updateSendAllStatus`
- Remove `ToggleSendAll`  and replace it with `updateSendAllStatus` which takes a boolean
- Cover the case when you click on an option twice ( ex: send all.. ) doesn't update the amount or update send all


packages/yoroi-extension/app/components/wallet/send/WalletSendForm.js
```diff
@@ -484,23 +484,22 @@ export default class WalletSendForm extends Component<Props> {
             options={sendAmountOptions}
             {...form.$('selectedAmount').bind()}
             onChange={value => {
-              this.form.$('selectedAmount').value = value
-              if(value === CUSTOM_AMOUNT && this.props.shouldSendAll) {
-                this.props.toggleSendAll();
+              if(value === CUSTOM_AMOUNT) {
```
yushih: space between `if` and `(`

packages/yoroi-extension/app/components/wallet/send/WalletSendForm.js
```diff
@@ -484,23 +484,22 @@ export default class WalletSendForm extends Component<Props> {
...
+                this.props.updateSendAllStatus(false);
+                this.form.$('amount').clear()
```
yushih: semicolon

packages/yoroi-extension/app/components/wallet/send/WalletSendForm.js
```diff
@@ -484,23 +484,22 @@ export default class WalletSendForm extends Component<Props> {
...
-                // if we toggle shouldSendAll from true -> false
-                // we need to re-enable the field
-                // and set it to whatever value was used for the sendAll value
-                this.props.updateAmount(new BigNumber(
-                  formattedAmountToNaturalUnits(
-                    this.form.$('amount').value,
-                    this.getNumDecimals(),
-                  )
-                ));
+                this.props.updateSendAllStatus(true)
```
yushih: semicolon

packages/yoroi-extension/app/components/wallet/send/WalletSendForm.js
```diff
@@ -484,23 +484,22 @@ export default class WalletSendForm extends Component<Props> {
...
+                // Will only update the amount to the max when we switch from custom amount to send all
```
yushih: No. The original behavior is to update the amount to the (auto-calculated) max when switching *from send all to custom amount*. Please resume the original behavior.

yushih (changes requested):

packages/yoroi-extension/app/components/wallet/send/WalletSendForm.js
```diff
@@ -484,23 +484,19 @@ export default class WalletSendForm extends Component<Props> {
             options={sendAmountOptions}
             {...form.$('selectedAmount').bind()}
             onChange={value => {
-              this.form.$('selectedAmount').value = value
-              if(value === CUSTOM_AMOUNT && this.props.shouldSendAll) {
-                this.props.toggleSendAll();
+              if (value === CUSTOM_AMOUNT) {
+                this.props.updateSendAllStatus(false);
+                this.form.$('amount').clear();
               } else if (value !== CUSTOM_AMOUNT) {
```
yushih: the `if` on line 490 is no-op

packages/yoroi-extension/app/components/wallet/send/WalletSendForm.js
```diff
@@ -484,23 +484,19 @@ export default class WalletSendForm extends Component<Props> {
...
-                this.props.toggleSendAll()
-              }
-              if (this.props.shouldSendAll) {
-                // if we toggle shouldSendAll from true -> false
-                // we need to re-enable the field
-                // and set it to whatever value was used for the sendAll value
+                this.props.updateSendAllStatus(true);
                 this.props.updateAmount(new BigNumber(
```
yushih, line 499: this function call `this.props.updateAmount()` is no-op

yushih (changes requested):
Note originally the behavior is: when switching from 'all tokens' to 'custom amount', the 'amount' input is retained.  With this PR, it is cleared.

But even if this is desired, there are some problems, please see comments to the code.

packages/yoroi-extension/app/components/wallet/send/WalletSendForm.js
```diff
@@ -484,23 +484,19 @@ export default class WalletSendForm extends Component<Props> {
             options={sendAmountOptions}
             {...form.$('selectedAmount').bind()}
             onChange={value => {
-              this.form.$('selectedAmount').value = value
-              if(value === CUSTOM_AMOUNT && this.props.shouldSendAll) {
-                this.props.toggleSendAll();
+              if (value === CUSTOM_AMOUNT) {
+                this.props.updateSendAllStatus(false);
+                this.form.$('amount').clear();
```
zuzunker: Should not clear the amount when selecting custom-amount

packages/yoroi-extension/app/components/wallet/send/WalletSendForm.js
zuzunker: @AhmedIbrahim336 , you can replace `else if (value !== CUSTOM_AMOUNT)` with just `else` here, because you have the checks like:

```
if (x === y) {

} else if (x !== y) {

}
```

So it's same as 

```
if (x === y) {

} else {

}
```

packages/yoroi-extension/app/components/wallet/send/WalletSendForm.js
zuzunker: @AhmedIbrahim336 , you have changed the logic so that when "custom amount" gets selected back - the amount pre-calculate for send-all is erased, this is why you removed the comment about preserving that value, but the function call to `.updateAmount` is not removed, even tho it lost it's sense.

BUT! The correct change should be to NOT erase the amount field, the logic was correct before that. Please return the comment back and handle this case correctly:

1. User has "custom amount" set by default
2. User changes to "send all"
3. User input is replaced with the new amount automatically
4. User changes back to "custom amount"
5. The automatic value is preserved and just made editable

zuzunker (changes requested):
@AhmedIbrahim336 , check comments and ping me for a sync if needed

packages/yoroi-extension/app/components/wallet/send/WalletSendForm.js
neuodev: Okay

yushih (changes requested):
@AhmedIbrahim336 could you please add a guard to the `onChange` handler of the `selectedAmount` dropdown, so that if the handler is executed but the selected item is not changed (this happens when the user re-select the item that is already selected), it just returns without doing anything.

You can see currently if you re-select "send all", it says "not enough fund". But if you re-select it again, it becomes normal. This is weird.

neuodev:
> @AhmedIbrahim336 could you please add a guard to the `onChange` handler of the `selectedAmount` dropdown, so that if the handler is executed but the selected item is not changed (this happens when the user re-select the item that is already selected), it just returns without doing anything.
> 
> You can see currently if you re-select "send all", it says "not enough fund". But if you re-select it again, it becomes normal. This is weird.

I got you point i just added a new check if we are selecting the same option again just return and do nothing

yushih (approved):

yushih (approved):

PR #2301, opened 2021-08-11
Add tokens list

neuodev:
tokenlist

@Marta-Ryzhok  Feel free to request changes


Marta-Ryzhok:
@AhmedIbrahim336 
if there is 50 type of tokens it will be listed below the input?

neuodev:
> if there is 50 type of tokens it will be listed below the input?
@Marta-Ryzhok Feel free to suggest another option

packages/yoroi-extension/app/components/wallet/send/WalletSendForm.js
```diff
@@ -403,7 +408,13 @@ export default class WalletSendForm extends Component<Props> {
         })
       ]
     })()
-
+    const tokenListClasses = classnames([
+      styles.tokenList,
+      {
+        [styles.show]: this.props.shouldSendAll && 
+           this.form.$('selectedAmount').value === tokenId
```
yushih: Should this be `this.form.$('selectedToken').value === tokenId` ?

packages/yoroi-extension/app/components/wallet/send/WalletSendForm.js
```diff
@@ -510,6 +524,15 @@ export default class WalletSendForm extends Component<Props> {
             )}
           />
 
+          <div className={tokenListClasses}>
+            <h1>{intl.formatMessage(messages.willSendAll)}</h1>
+            { tokenOptions.map(token => (
+              <p key={token.id} className= "accordion-item__paragraph" >
```
yushih: What is the class `accordion-item__paragraph` for? Seems there's no style that applies to this class.

Also please fix the indentation.

yushih (changes requested):
I think there's a problem: if I select a token and select "send all", then I'm only sending this token so it should not show the token list, but it does:
tokenlist

Marta-Ryzhok:
> I think there's a problem: if I select a token and select "send all", then I'm only sending this token so it should not show the token list, but it does:
> tokenlist

Yeah you’re right, shouldn’t be a token list below

packages/yoroi-extension/app/components/wallet/send/WalletSendForm.js
neuodev: I replaced it 👍

yushih (changes requested):
In the last commit, 3 irrelevant `package-lock.json` files are checked in. Please remove the changes. Otherwise LGTM.

yushih:
@AhmedIbrahim336, please resolve the conflict.

yushih (changes requested):
@AhmedIbrahim336 , please resolve the conflict. And be careful not to break the behavior of the "send all" dropdown.

neuodev:
@yushih  Done

yushih (approved):

PR #2314, opened 2021-08-19
Extend the Cardano dApp connector with more methods

yushih:


rooooooooob:
@vsubhuman @yushih  Just a reminder that [the CIP](https://github.com/cardano-foundation/CIPs/pull/88) is still in progress. Last month there was a change to the way the injected code is structured as it's now in a namespaced `cardano.yoroi` `wallet` object instead to allow multiple wallets supporting the same spec. There's been a couple other changes too now that we're in the review stage of the CIP.

zuzunker (approved):

PR #2321, opened 2021-08-25
Allow overriding the SSL key and cert used to create the Webpack server

pedromtcosta:
## Abstract
This proposal aims to provide a way for developers to easily override the SSL key and certificate used in the creation of the Webpack server. (method exported from `webpack-httpolyglot-server`)

## Motivation
In my system (PureOS, Linux distro based on Debian), building the extensions with `npm run dev:build` or `npm run dev:stable` throws the following exception:
`error:140AB18F:SSL routines:SSL_CTX_use_certificate:ee key too small`. This PR adds a way for other people facing this error to circumvent this problem.

## Background
Inspecting the code from `webpack-httpolyglot-server`, I found out it tries getting the key and certificates from it's own package location (`./node_modules/webpack-httpolyglot-server/ssl`) and that we can also send a specific key and certificate on the last argument from the method.

I am not sure why the keys contained in the package's directory doesn't work in my system, but I assume it should be because this package was developed using a different OS than mine.

## Proposal
When running on `debug`, check if there's a file named `sslOverrides.js` at the `yoroi-extension/scripts` folder. If the file exists, load it and pass the `key` and `cert` defined there to the `serverOpts` argument from `webpack-httpolyglot-server`

### DIFF
--- a/packages/yoroi-extension/.gitignore
+++ b/packages/yoroi-extension/.gitignore
@@ -40,3 +40,6 @@ Yoroi Shelley Testnet
 
 # Local Netlify folder
 .netlify
+
+# SSL overrides for DEV
+scripts/sslOverrides.js
--- a/packages/yoroi-extension/scripts/build.js
+++ b/packages/yoroi-extension/scripts/build.js
@@ -31,6 +31,9 @@ export function buildDev(env: string) {
   const webpackDevMiddleware = require('webpack-dev-middleware');
   const webpackHotMiddleware = require('webpack-hot-middleware');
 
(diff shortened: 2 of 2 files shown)


yushih (approved):
LGTM. But I wonder why we need the dev server to serve https in the first place.

zuzunker (approved):

zuzunker:
> But I wonder why we need the dev server to serve https in the first place.

Good question. We might need to review it later to see if we can simplify things for the dev env specifically. But not high priority tho

PR #2322, opened 2021-08-27
Send All (ADA, ERG) But Keep Tokens

neuodev:
On the send page adding the new option to send all (ADA, Ergo ) But keep tokens

### DIFF
--- a/packages/yoroi-extension/app/api/ada/index.js
+++ b/packages/yoroi-extension/app/api/ada/index.js
@@ -986,8 +986,17 @@ export default class AdaApi {
       let unsignedTxResponse;
       const trxMetadata =
         request.metadata !== undefined ? createMetadata(request.metadata): undefined;
-
-      if (hasSendAllDefault(request.tokens)) {
+      console.log(request.tokens[0])
+      if(request.tokens[0].shouldSendAllKeepTokens){
+        const receiver = request.receivers[0];
+        unsignedTxResponse = shelleySendAllUnsignedTx(
+          receiver,
+          request.utxos,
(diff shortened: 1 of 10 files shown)


packages/yoroi-extension/app/api/ada/index.js
```diff
@@ -986,7 +986,7 @@ export default class AdaApi {
       let unsignedTxResponse;
       const trxMetadata =
         request.metadata !== undefined ? createMetadata(request.metadata): undefined;
-
+      console.log(request.tokens[0])
```
neuodev, line 989: Don't give attention to `logs` this is just to track the current values.
Still on development will remove it once we are done

packages/yoroi-extension/app/api/common/index.js
```diff
@@ -324,7 +324,7 @@ export function builtSendTokenList(
         new BigNumber(0)
       );
       amount.add({
-        amount: relatedUtxoSum,
+        amount: relatedUtxoSum.minus(1500000),
         identifier: token.token.Identifier,
         networkId: token.token.NetworkId,
       });
```
neuodev, line 330: Here if we don't specify amount `amount == null ` this means that we will send it all so we are summing all the Utxos in the variable `relatedUtxoSum`. 
### The problem 
- I am getting an error that `NotEnoughMoneyToSend` 
- it goes away when I add some safe margin for example if I want to send 1000 ADA the sum of ada from Utxo will be 1000ADA this will throw an error but if I subtract 2ADA The error goes away and I can send the 998ADA

PR #2363, opened 2021-09-02
[nomerge] Chinese Traditional translation update

Tadashi1024:
Hi teams,
I have updated the Chinese Traditional translation for better wording and grammar.


zuzunker:
Hi, @Tadashi1024! Thank you for this update! Unfortunately we cannot get it merged directly as we are using a side-service Crowdin for translations so any files except default english are getting updated only from there, but I will make sure we put these changes into Crowdin and then they will be merged in in another PR before the next release! 🙌

PR #2370, opened 2021-09-04
Setting up material UI

paulclindo:
Upgraded  `react`, `react-dom` to v17 as `Material UI` v5 needs them as peer dependencies. 
- Remove unused react imports since now we don't need it to use JSX and there is a new JSX transform
- Set up babel, eslint, and flow


zuzunker (approved):

zuzunker (commented):
/check

PR #2371, opened 2021-09-06
[dapp-connector] Display token info even if it's not available

### LINKED ISSUE #2397: indefinite loader when issuing tokens on Ergo using the dapp connector
I am using Yoroi-nighty 4.7.200 and dapp connector 0.7.0.
When trying to issue tokens, an indefinite loader appears and I get the following error in the console:
```
Failed to load resource: the server responded with a status of 400 ()
ergo.bundle.js:1 RemoteFetcher::getAssetInfo error: {
  "name": "SyntaxError",
  "message": "Unexpected ''",
  "at": 1,
  "text": ""
}
error @ ergo.bundle.js:1
ergo.bundle.js:1 batchedFetcher::batchAssetInfo error: {
  "message": "api.errors.checkAdressesInUseApiError: undefined",
  "name": "E",
  "stack": "E: api.errors.checkAdressesInUseApiError: undefined
    at Object.errorFactory (chrome-extension://poonlenmfdfbjfeeballhiibknlknepo/js/618.bundle.js:1:88337)
    at chrome-extension://poonlenmfdfbjfeeballhiibknlknepo/js/618.bundle.js:1:89006
    at async Promise.all (index 0)
    at async chrome-extension://poonlenmfdfbjfeeballhiibknlknepo/js/618.bundle.js:1:87127
    at async fe (chrome-extension://poonlenmfdfbjfeeballhiibknlknepo/js/ergo.bundle.js:1:133723)
    at async N (chrome-extension://poonlenmfdfbjfeeballhiibknlknepo/js/618.bundle.js:1:45372)
    at async he (chrome-extension://poonlenmfdfbjfeeballhiibknlknepo/js/ergo.bundle.js:1:134338)
    at async Ee._getTxAssets (chrome-extension://poonlenmfdfbjfeeballhiibknlknepo/js/ergo.bundle.js:1:186229)
    at async chrome-extension://poonlenmfdfbjfeeballhiibknlknepo/js/ergo.bundle.js:1:192555
    at async h.trigger (chrome-extension://poonlenmfdfbjfeeballhiibknlknepo/js/544.bundle.js:1:1271576)",
  "id": "api.errors.checkAdressesInUseApiError",
  "defaultMessage": "!!!Error received from server while checking used addresses.",
  "values": {}
}
error @ ergo.bundle.js:1
618.bundle.js:1 rolling back Lovefield query for
Token
EncryptionMeta with error {"id":"api.errors.checkAdressesInUseApiError","defaultMessage":"!!!Error received from server while checking used addresses.","values":{}}
N @ 618.bundle.js:1
```
ind
 
cc @rooooooooob @vsubhuman

neuodev:
Screen Shot 2021-09-06 at 12 05 13 PM

task on [Asana](https://app.asana.com/0/1200458692294040/1200899147254165/f)


packages/yoroi-extension/app/ergo-connector/stores/ConnectorStore.js
```diff
@@ -297,12 +297,16 @@ export default class ConnectorStore extends Store<StoresMap, ActionsMap> {
...
+    try {
+      await addErgoAssets({
+        db: selectedWallet.getDb(),
+        tokenIdentifiers,
+        getAssetInfo: stateFetcher.getAssetInfo,
+        network: selectedWallet.getParent().getNetworkInfo(),
+      });
+    } catch (error) {
+      console.error('Token info unavailable', error.message)
+    }
```
zuzunker, line 316: Let's not wrap the entire call to the storage with a catch related to unavailable tokens, because there might be a lot of other reasons for a db-call to fail than specifically non-available tokens, so that error message will be very confusing in the logs later.

The state-fetcher function is being passed right there, so you can wrap it into a safety check like that:
```
try {
  await addErgoAssets({
    db: selectedWallet.getDb(),
    tokenIdentifiers,
    getAssetInfo: async (req) => {
        try {
            return await stateFetcher.getAssetInfo(req);
        } catch (e) {
            console.error('Aseet info request failed', e);
            return {};
        }
    },
    network: selectedWallet.getParent().getNetworkInfo(),
  });
} catch (error) {
  console.error('Failed to add ergo assets!', error);
}
```

zuzunker (changes requested):

packages/yoroi-extension/app/ergo-connector/components/signin/SignTxPage.js
```diff
@@ -126,20 +126,41 @@ class SignTxPage extends Component<Props> {
...
     if (mintedToken != null) {
       return mintedToken;
     }
-    return this.props.getTokenInfo(tokenEntry);
+
+    try {
+      return this.props.getTokenInfo(tokenEntry);
+    } catch (error) {
+      return null
+    }
```
zuzunker: @AhmedIbrahim336 , also, this seems like too wide of an assumption. There might be various reasons for an error and here we are silently ignoring it.

if you check what is being passed as the `getTokenInfo` prop, you will see:
```
genLookupOrFail(this.generated.stores.tokenInfoStore.tokenInfo)
```

Which just wraps a simple map to explicitly throw an error in case the key is not found. So instead of catching that error - you can replace the wrapper with something like `genLookupOrNull` and not throw the error in the first place.

zuzunker (changes requested):

zuzunker:
closes https://github.com/Emurgo/yoroi-frontend/issues/2397

arobsn (changes requested):
I agree with @vsubhuman on be more "surgical" on these points, so nothing to add for now.

zuzunker (approved):

zuzunker (commented):
/check

PR #2373, opened 2021-09-06
Adapt to Trezor streamed transaction signing

yushih:
Upstream change: https://github.com/vacuumlabs/connect/pull/13 .

Previously, the `cardanoSignTransaction` API of `trezor-connect` returned the serialized tx, so Yoroi extension needed only to submit it. Now it only returns witnesses and the Catalyst signature (if applicable), so Yoroi extension needs to build the tx by itself, just as working with Ledger.

Manual test result:
|scenario|test result|comments|
|-|-|-|
|sending ADA|pass||
|delegate stake|pass||
|withdraw rewards|<del>fail</del>pass|<del>likely due to https://github.com/Emurgo/yoroi-graphql-migration-backend/pull/255, will re-check</del>|
|catalyst registration|<del>fail</del>pass|<del>likely due to minimum fund requirement, will double-check on the testnet backend</del>|

<del>There is a mock Trezor implementation used by e2e tests that needs to be updated to align with the new `trezor-connect` API. Currently, I'm using only commit 967e43f to pass flow and lint checks until this is done. </del>


zuzunker (approved):

zuzunker (commented):
/check

PR #2374, opened 2021-09-06
[DO NOT MERGE] Tmp patch to make the Trezor update PR work

yushih:
Before `connect.trezor.io`, which does the heavy lifting of signing txs, is updated to the streamed Cardano tx signing API, in order to test https://github.com/Emurgo/yoroi-frontend/pull/2373, this patch must be applied. Also, a local trezor connect needs to be on localhost:8088 using code on this branch: https://github.com/vacuumlabs/connect/pull/13.  See https://wiki.trezor.io/Developers_guide:Running_Trezor_Connect_on_localhost.

### DIFF
--- a/packages/yoroi-extension/.eslintignore
+++ b/packages/yoroi-extension/.eslintignore
@@ -1,3 +1,4 @@
 node_modules
 /chrome/content-scripts
 flow-typed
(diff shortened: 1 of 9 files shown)


yushih:
not longer needed

PR #2376, opened 2021-09-07
Function for counting token and NFT types

pedromtcosta:
# Abstract
This PR intends to add a method at `packages/yoroi-extension/app/api/ada/lib/storage/models/utils.js` to count token and NFT types from addresses UTXOs

# Motivation
We are going to display the token and NFT counts for each wallet together with the balance. The addition of this method is the first step towards that.

# Background
It is important to understand that despite having the information from the tokens in the UTXOs, we don't have a way to distinguish tokens which are NFTs from the ones that aren't by just using UTXO data. Because of that a new endpoint was added on the backend which returns metadata from a mint transaction of a token, given its name and policy ID. It is standard that a metadata is attached when NFTs are minted and that it is wrapped inside the `key` `721`.

***References:***
**[Minting NFTs](https://developers.cardano.org/docs/native-tokens/minting-nfts/)**
**[How we minted the NFTAs, and why we went for this standard](https://developers.cardano.org/blog/nft-minting-standard/)**

For more info, see the [PR for adding this new endpoint](https://github.com/Emurgo/yoroi-graphql-migration-backend/pull/254)

Also, see the [task in Asana](https://app.asana.com/0/1200842454080452/1200880500671157).

# Implementation
The implementation is composed of the following parts:

### Add new function on `RemoteFetcher`
As described in the background, there's a new endpoint on the backend so add a new function named `getMultiAssetMintMetadata` to `RemoteFetcher` and `IFetcher` to send requests to it. We also add the new `*Func` type `MultiAssetMintMetadataFunc` to hide the implementation behind a type, as it is done with the other  requests.

### Insert flag to determine if a token is a NFT in the DB

Here we add a new field called `IsNFT` to the `TokenRow` which is inserted into the DB. Then, at the `genCardanoAssetMap` function we make a request to the new endpoint using `MultiAssetMintMetadataFunc` to get the metadata from tokens passed as parameter to `genCardanoAssetMap` and check if we have any metadata with `key === '721'` when mapping the tokens. If we do, we set the `IsNFT` flag to `true`.

We also make a change to the `upsert` method of `ModifyToken` so it updates the token row in case the `IsNFT` flag is changed.

### Add `getTokenCountForAddresses` and `getTokenCountForUtxos` function

Finally, we add the functions which perform the counting. `getTokenCountForAddresses` is just a wrapper for which calls `getTokenCountForUtxos` for the addresses, and `getTokenCountForUtxos` simply counts the token and NFT types based on the `IsNFT` flag and return a object with the format `{nftTypes: number, tokenTypes: number}`.

### Pass along the new `IsNFT` flag and `MultiAssetMintMetadataFunc` where needed

Lastly we need to update references which are creating `TokenRow` objects to include the `IsNFT` flag as `false` and update the tests which use `getTokenCountForAddresses` to pass a mock implementation of `MultiAssetMintMetadataFunc` as parameter.


packages/yoroi-extension/app/api/ada/lib/state-fetch/remoteFetcher.js
```diff
@@ -427,4 +429,25 @@ export class RemoteFetcher implements IFetcher {
         throw new GetCatalystRoundInfoApiError();
       });
   }
+
+  getMultiAssetMintMetadata: MultiAssetMintMetadataRequest
+    => Promise<MultiAssetMintMetadataResponse> = async (body) => {
+      const { BackendService } = body.network.Backend;
+      if (BackendService == null) throw new Error(`${nameof(this.getCatalystRoundInfo)} missing backend url`);
```
neuodev: Please update the name of the function from `this.getCatalystRoundInfo` to `this.getMultiAssetMintMetadata`

packages/yoroi-extension/app/api/ada/lib/state-fetch/remoteFetcher.js
```diff
@@ -427,4 +429,25 @@ export class RemoteFetcher implements IFetcher {
...
+        `${BackendService}/${prefix}multiAsset/metadata`,
+        {
+          method: 'post',
+          data: {
+            assets: body.assets
+          }
+        }
+      ).then(response => response.data)
+      .catch((error) => {
+        Logger.error(`${nameof(RemoteFetcher)}::${nameof(this.getCatalystRoundInfo)} error: ` + stringifyError(error));
```
neuodev: Update this as well

packages/yoroi-extension/app/api/ada/lib/state-fetch/remoteFetcher.js
pedromtcosta: Done!

packages/yoroi-extension/app/api/ada/lib/state-fetch/remoteFetcher.js
pedromtcosta: Done!

zuzunker (approved):

packages/yoroi-extension/app/api/ada/lib/state-fetch/remoteFetcher.js
```diff
@@ -427,4 +429,25 @@ export class RemoteFetcher implements IFetcher {
         throw new GetCatalystRoundInfoApiError();
       });
   }
+
+  getMultiAssetMintMetadata: MultiAssetMintMetadataRequest
+    => Promise<MultiAssetMintMetadataResponse> = async (body) => {
+      const { BackendService } = body.network.Backend;
+      if (BackendService == null) throw new Error(`${nameof(this.getMultiAssetMintMetadata)} missing backend url`);
+      const isMainnet = /Mainnet/g.test(body.network.NetworkName)
+      const prefix = isMainnet ? '' : 'api/'
```
yushih: Did you follow one of the earlier PRs to generate the URL? I think it was wrong and was changed. Regardless of whether it is mainnet, there is the `/api` component in the URL.

packages/yoroi-extension/app/api/ada/lib/storage/bridge/updateTransactions.js
```diff
@@ -1959,6 +2003,8 @@ export async function genCardanoAssetMap(
     tokenInfoResponse = {};
   }
 
+  const metadata = await getTokenMintMetadata(tokenIds, getMultiAssetMetadata, network);
```
yushih, line 2009: Sometimes this function might be called with the first parameter `tokenIds` = `[]`. In this case, the server-side returns a 500 error with `assets should not be empty`. I think we can optimize away this round trip to the server by making it a no-op?

yushih (changes requested):
Generally LGTM and nice documentation! But there are some subtle problems:

1.
Although the logic is 100% correct if Yoroi is run from a clean state, actually when the user upgrades to a new version with code in the PR, their indexeddb already contain `Token` entries without this field. So I think it's more appropriate to set the type of the field to `IsNFT?: boolean` in `tables.js`. And in `updateTransactions.js`, when the token row doesn't have this row, request the mint info endpoint and update the token row (maybe changing L1990 suffices).

2.
I find that every time Yoroi starts, 5 default token entries are inserted.

3.
The inserted token rows don't have the `IsNFT` field.

I didn't get to the bottom of causes 2 and 3, but it seems to me that putting `IsNFT` inside `Metadata` works.

packages/yoroi-extension/app/api/ada/lib/state-fetch/remoteFetcher.js
pedromtcosta: Fixed in the latest push

packages/yoroi-extension/app/api/ada/lib/storage/bridge/updateTransactions.js
pedromtcosta: Fixed in the latest push

pedromtcosta:
> Generally LGTM and nice documentation! But there are some subtle problems:
> 
> 
> Although the logic is 100% correct if Yoroi is run from a clean state, actually when the user upgrades to a new version with code in the PR, their indexeddb already contain `Token` entries without this field. So I think it's more appropriate to set the type of the field to `IsNFT?: boolean` in `tables.js`. And in `updateTransactions.js`, when the token row doesn't have this row, request the mint info endpoint and update the token row (maybe changing L1990 suffices).
> 
> 
> I find that every time Yoroi starts, 5 default token entries are inserted.
> 
> 
> The inserted token rows don't have the `IsNFT` field.
> 
> I didn't get to the bottom of causes 2 and 3, but it seems to me that putting `IsNFT` inside `Metadata` works.

@yushih thank you for the very detailed review! I was able to find out why these problems happen and the PR should be good now. To elaborate on each point you raised:

1. changing this line wasn't needed actually. I changed `IsNFT` to be optional as suggested and my tokens which previously didn't had this field at all now have it set properly;
2. This was actually a deeper bug which manifested just now. In theory any changes to the metadata of the default assets would cause them to be re-inserted instead of updated. See [this commit](https://github.com/Emurgo/yoroi-frontend/pull/2376/commits/7b86f2c457fa052e718e7587893bda5b339a438c);
3. This was something I overlooked. We also have to call `schemaBuilder.addColumn('IsNFT', Type.BOOLEAN)` so the schema gets created properly;

Bellow I am leaving how I tested the above to make sure it is working:
- clear all data from Token table;
- build and run current version from`develop` branch;
- look into current values from `Token` table: only default assets are present (ADA, Ergo, TADA). My native assets and NFTs are not on the table yet;
- resync wallet;
- look into current values from `Token` table: my native tokens and NFTs are there now (without the `IsNFT` flag, as this is the `develop` branch). So now we have 9 records in total, 5 default assets and 4 native assets;
- switch to `chore/token-and-nft-count-method` branch;
- build and run current version from `chore/token-and-nft-count-method` branch;
- right after the extension starts, look into the `Token` table: we still have the same 9 records, but the 5 default assets records get the new `IsNFT` flag with the value set to false;
- enter my wallet which holds some native tokens;
- look into the `Token` table again: now I see my native tokens have the `IsNFT` flag correctly set (so NFTs have the flag as `true` and non-NFTs have it set to `false`);

packages/yoroi-extension/app/api/ada/lib/storage/database/primitives/tables.js
```diff
@@ -832,7 +832,7 @@ export const populatePrimitivesDb = (schemaBuilder: lf$schema$Builder) => {
     .addColumn(TokenSchema.properties.TokenId, Type.INTEGER)
     .addColumn(TokenSchema.properties.NetworkId, Type.INTEGER)
     .addColumn(TokenSchema.properties.IsDefault, Type.BOOLEAN)
-    .addColumn(TokenSchema.properties.IsNFT, Type.BOOLEAN)
+    .addColumn('IsNFT', Type.BOOLEAN)
```
yushih, line 835: I guess this change is not meant to be there?

packages/yoroi-extension/app/api/ada/lib/storage/database/primitives/api/write.js
```diff
@@ -692,8 +692,18 @@ export class ModifyToken {
           toAdd.push(row);
         } else if (JSON.stringify(item.Metadata) !== JSON.stringify(row.Metadata)
         || item.IsNFT !== row.IsNFT) {
-          // we want to update the row if the metadata was updated
-          toAdd.push(row);
+          // we want to update the row if the metadata or IsNFT flag was updated
```
yushih, line 695: question: `row` is the original entry, correct? How could it not have the `TokenId` field?

yushih (approved):
Thanks for the detailed explanation! LGTM. See below comments for some minor issues.

packages/yoroi-extension/app/api/ada/lib/storage/database/primitives/tables.js
pedromtcosta: It was, otherwise we get this flow error:  "Cannot call `schemaBuilder.createTable(...).addColumn(...).addColumn(...).addColumn(...).addColumn` with `TokenSchema.properties.IsNFT` bound to `name` because  undefined [1] is incompatible with  string [2].Flow(incompatible-call)" now that `IsNFT` is optional

packages/yoroi-extension/app/api/ada/lib/storage/database/primitives/api/write.js
pedromtcosta: Actually not. `item` is the original entry. `row` comes from iterating `deduplicatedRows`, and `deduplicatedRows` is created from the `rows` function argument. At `app/api/ada/lib/storage/database/index.js` we have `populateAssetDefaults` calling `upsert` with the `defaultAssets` coming from `app/api/ada/lib/storage/database/prepackaged/networks.js`. These default assets are defined without `TokenId`

packages/yoroi-extension/app/api/ada/lib/storage/database/primitives/tables.js
yushih: Aha, make sense. Thanks!

packages/yoroi-extension/app/api/ada/lib/storage/database/primitives/api/write.js
yushih: Thanks for explaining!

yushih (approved):

PR #2378, opened 2021-09-09
Add missing row on Lovefield

SebastienGllmt:
Although in IndexedDB we were saving the network name, I forgot to add this column to Lovefield. This was causing a subtle issue where if for some reason Lovefield cache was invalidated, Yoroi UI would lose access to the network name causing an empty string to appear in the wallet list where the network name is supposed to be


zuzunker (approved):
Thanks, Seba!

PR #2389, opened 2021-09-14
Txs with invalid scripts are marked as Failed

yushih:
[task in asana](https://app.asana.com/0/1200528935481340/1200691762858109)

Corresponding backend change: https://github.com/Emurgo/yoroi-graphql-migration-backend/pull/259

For backward compatibility, if the added `valid_transaction` is not present in the response of tx history query, the tx is treated as valid.


packages/yoroi-extension/app/api/ada/lib/storage/bridge/updateTransactions.js
```diff
@@ -2261,6 +2261,9 @@ export function networkTxHeaderToDb(
         Extra: {
           Fee: tx.fee,
           Metadata: tx.metadata,
+          // for backward compatiblity, if the field is not present, we take the tx as valid
+          IsValid: !Object.prototype.hasOwnProperty.call(tx, 'valid_transaction') ||
+            tx.valid_transaction,
```
zuzunker: Think this might be replaced just with:
```
tx.valid_transaction ?? true;
```

packages/yoroi-extension/app/components/wallet/transactions/Transaction.js
```diff
@@ -567,6 +571,9 @@ export default class Transaction extends Component<Props, State> {
     const { intl } = this.context;
     const isFailedTransaction = state < 0;
     const isPendingTransaction = state === TxStatusCodes.PENDING;
+    const isValidTransaction = (data instanceof CardanoShelleyTransaction) ?
+      data.isValid :
+      true;
```
zuzunker, line 584: Might be something like:
```
const isShelleyTx = ...;
const isValidTransaction =
    !isShelleyTx || data.isValid;
```

zuzunker (approved):

packages/yoroi-extension/app/components/wallet/transactions/Transaction.js
zuzunker: not critical at whatsoever, just thought while reading

packages/yoroi-extension/app/api/ada/lib/storage/bridge/updateTransactions.js
zuzunker: @yushih , I ended up adding this single change because it feels more secure to rather treat any nullish/undefined state as `true`, than checking specifically for whether the field is present or not. E.g. any kind of an accidental rework on the backend that might set this value to `undefined` or `null` when flag is not present is in the realm of possibilities.

zuzunker (commented):
/check

zuzunker (commented):
/check

zuzunker (commented):
/check

zuzunker (commented):
/check

PR #2395, opened 2021-09-16
Replace from RP to MUI components - 1

paulclindo:
List of components: 
- [x] Added logic to handle themes in MUI
- [x] Button 

Note: In Dialog component, RP Buttons are not replaced yet since it would make bigger changes on other components and testing. So that will be worked on a future PR for Dialog component


zuzunker:
@paulclindo , resolve conflict plz

zuzunker (approved):

PR #2400, opened 2021-09-19
Coin selection improve

zuzunker:
1. Updated to the new version of the serialization-lib that requires new parameters for the max-tx-size and max-value-size and handles change with many native assets securely
2. Added displaying the tx size in the tx-send confirmation dialog when the size is close to the limits
3. Inserted input prioritisation into the coin-selection for ADA txs so that pure inputs or inputs with required assets are always prioritized and randomized and dirty inputs with unrelated assets are sorted by highest spendable value


zuzunker (commented):
/check

zuzunker (commented):
/check

zuzunker (commented):
/check

PR #2401, opened 2021-09-20
Assets deposit value

zuzunker:
Built on top of: https://github.com/Emurgo/yoroi-frontend/pull/2400

Implemented background request that along with the balance recalculates sum of minimum ADA values for all existing dirty UTxOs (if any) and added displaying this information next to the wallet balance so it can be visible for users that there's a minimum unspendable amount in their wallet as token deposit.


zuzunker (commented):
/check

zuzunker (commented):
/check

PR #2408, opened 2021-09-22
Fix e2e test crash when starting

yushih:
The cause of the problem is that in a recent change, `packages/yoroi-extension/chrome/extension/ergo-connector/utils.js` becomes imported by the e2e test mock server (running in node.js) but itself imports `packages/yoroi-extension/app/utils/logging.js`, which is meant to be processed by webpack and crashes when running in node.js.

The solution is not to directly import the logger into  `packages/yoroi-extension/chrome/extension/ergo-connector/utils.js` but to use runtime dependency injection.


zuzunker (approved):
Nice! Thank you!

PR #2418, opened 2021-09-27
Remove Zero Decimals From Yoroi 

neuodev:
Complete task information available on [asana](https://app.asana.com/0/1200458692294040/1200897469103929/f)

### Look how we are displaying the balance right now ( 0 ADA ) Instead of (0.00000..... ADA)
Screen Shot 2021-09-27 at 10 55 16 AM

Screen Shot 2021-09-27 at 10 57 33 AM


packages/yoroi-extension/app/components/topbar/NavWalletDetails.js
```diff
@@ -185,11 +185,12 @@ export default class NavWalletDetails extends Component<Props> {
         shiftedAmount,
         tokenInfo.Metadata.numberOfDecimals,
       );
-
+      // const displayAmount = shiftedAmount.toFormat(tokenInfo.Metadata.numberOfDecimals)
```
yushih: please don't leave commented-out lines in the code

packages/yoroi-extension/app/components/topbar/NavWalletDetails.js
```diff
@@ -185,11 +185,12 @@ export default class NavWalletDetails extends Component<Props> {
...
       balanceDisplay = (
         <>
+         {/* {displayAmount} */}
```
yushih: please remove this line

packages/yoroi-extension/app/components/topbar/NavWalletDetails.js
```diff
@@ -185,11 +185,12 @@ export default class NavWalletDetails extends Component<Props> {
...
           {beforeDecimalRewards}
-          <span className={styles.afterDecimal}>{afterDecimalRewards}</span>
+          <span>{afterDecimalRewards}</span>
```
yushih: please remove `.afterDecimal` from the scss file

yushih: Also, since we are displaying the parts before and after the decimal point with the same font now, can we not split the amount? It is confusing.

packages/yoroi-extension/app/utils/formatters.js
```diff
@@ -1,15 +1,38 @@
 // @flow
 import BigNumber from 'bignumber.js';
 
+export const removeZeros: (string => string) = (amount: string): string => {
```
yushih: Please don't export this function, unless it is used by other modules.

packages/yoroi-extension/app/utils/formatters.js
```diff
@@ -1,15 +1,38 @@
...
   amount: BigNumber,
   decimalPlaces: number,
 ): [string, string] {
   const valString = amount.toFormat(decimalPlaces);
   const startIndex = valString.length - decimalPlaces;
-  return [valString.substring(0, startIndex), valString.substring(startIndex)];
+  let beforeDecimal = valString.substring(0, startIndex)
+  const afterDecimal = removeZeros(valString.substring(startIndex))
+  // Remove the dots if no decimals
+  if (Number(afterDecimal) === 0) {
```
yushih: why not test if `afterDecimal === ''`?

packages/yoroi-extension/app/utils/formatters.js
```diff
@@ -1,15 +1,38 @@
...
+    beforeDecimal = beforeDecimal.slice(0, beforeDecimal.length - 1)
+  }
+  return [beforeDecimal, afterDecimal]
```
yushih, line 20: Note when `amount` is `new BigNumber('1')` and `decimalPlaces` is `0`, this function outputs `['', '']`, but it should output `['1', '']`. Please fix and add a test file `formatters.test.js` to test it.

packages/yoroi-extension/app/utils/formatters.js
yushih: The purpose of this function is to remove trailing zeros? Then you can simpily do:
`return amount.replace(/0+$/, '')`

packages/yoroi-extension/app/utils/formatters.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
+export const amountWithoutZeros: (string => string) = (amount: string): string => {
```
yushih: `return amount.replace(/0+$/, '')..replace(/\.$/, '')` would do

yushih (changes requested):

packages/yoroi-extension/app/components/topbar/NavWalletDetails.js
neuodev: I got your point. The reason why we are still splitting the amount is the `splitAmount` function has the logic to remove the zeros. Also, In the dashboard page we are still displaying the decimal points in smaller fonts

packages/yoroi-extension/app/utils/formatters.js
neuodev: Awesome 👏

packages/yoroi-extension/app/utils/formatters.js
neuodev: Because
```js
Number('') // 0 
// Also 
Number(' ') // 0 
```
Any way I just update it to 

```js
if (!afterDecimal) {
 // 
} 
```

packages/yoroi-extension/app/utils/formatters.js
neuodev: I just run this test case and it passed 
```js 
test('Should return only integer of no decimals', () => {
    const amount = new BigNumber('1')
    const [before, after] = splitAmount(amount, 5)
    expect(before).toBe('1')
    expect(after).toBe('')
})
```

yushih (changes requested):
`splitAmount` is wrong when `decimalPlaces` is `0` (which could happen for some tokens)
This test fails:
```
test('splitAmount with 0 decimal place', () => {
    const amount = new BigNumber('1')
    const [before, after] = splitAmount(amount, 0)
    expect(before).toBe('1')
    expect(after).toBe('')
})
```

packages/yoroi-extension/app/utils/formatters.js
yushih: Please note this test doesn't cover the case where `decimalPlaces` is `0`.

packages/yoroi-extension/app/utils/formatters.js
yushih: Make sense.

yushih (approved):

yushih (approved):

PR #2455, opened 2021-10-14
Pack same tokens together, if available

yushih:
[the task](https://app.asana.com/0/1200528935481340/1201060593215866/f)


packages/yoroi-extension/app/api/ada/transactions/shelley/transactions.js
```diff
@@ -695,7 +697,30 @@ function _newAdaUnsignedTxFromUtxo(
...
+                false,
+                { networkId: protocolParams.networkId }
+              );
+              usedUtxos.push(packCandidate);
+
+              packed = true;
+              break;
+            }
+          }
+          if (oneExtraInput && !packed) {
```
zuzunker, line 726: packing an additional token input should not affect the "extra-input" logic, because the branch with `oneExtraInput = false` will also pack the additional one for token - so this breaks the logic.

```suggestion
          if (oneExtraInput) {
```

yushih: But when we *do* pack an input, there is no need for an extra input, because the packed input serves as the "extra" one.

zuzunker: Ah, because if we found an additional input where all assets are already present in the tx - we definitely have enough ADA purified from that utxo to cover getting the change. Good point, I missed that.

zuzunker (approved):

zuzunker (approved):

PR #2460, opened 2021-10-18
add an initial example to test minting NFTs

twwu123:
Example to mint a native token on Cardano. To mint an NFT, all we need to do is to add metadata to the transaction and change to policy script slightly to include a slot to timeout (the minting policy scripts are simply native scripts).


a-hosani (approved):

zuzunker:
closing in favour of the separate example dapp project

PR #2462, opened 2021-10-20
Add new option to export transaction id as a new column

### LINKED ISSUE #2453: Export transaction CSV with transaction id
Hello,

I have noticed that csv transactions exported to the csv file for tax purposes do not include transaction id. Is there any plan to add it ? Or may be there are any reason not to add it ?

neuodev:
Screen Shot 2021-10-21 at 10 16 35 AM

## Currently 
Users can expert their transactions history (including the rewards) without transactions ids being included 

## Expected Behaviour 
Users should have the option to include the transaction ids as a new column

## CSV Example
Screen Shot 2021-10-21 at 10 23 54 AM

## To Do 
- [X] Test for ergo and cardano  
- [ ] Test on Hardware wallet @yushih  (Can you please check it 🙌)

## Notes:
1. We are skipping transaction ids related to reward as we currently don't store them.


zuzunker:
Testing with hardware wallets now

zuzunker (approved):

PR #2470, opened 2021-10-26
Add Polish language files

matalbec:
This PR adds files for Polish language locale. No translation is provided yet but as I understand this step is needed for translation to show up on https://crowdin.com/project/yoroi-app

I have prepared an initial translation here https://github.com/matalbec/yoroi-frontend/blob/polish_translation_test/packages/yoroi-extension/app/i18n/locales/pl-PL.json which I am currently testing to see if I got the context right. I will use it to propose translations on crowdin.

I have tried to exclude Polish language from production build but I was unable to verify whether it really is excluded as I am unable to build production on Windows 10(I am still not sure if that is just an issue with my setup). I have also changed the way in which languages in translation are handled(see first commit).


zuzunker:
HI, @matalbec! Thank you for the contribution! The only thing is that you need to remove changes for all translation files except the `en-US.json` - it will be used as default for all other languages when they will be missing that key but also those other language files should never be edited manually, only through Crowdin, so this new value will appear automatically in all languages next time we get update from Crowdin

matalbec:
Hi, for some reason I had this idea that the process is to first add files in English which would later on appear in Crowdin for translation. After double reading the README for translations I am ot sure where I got that idea from I will revert all of the changes to language files except for en-US.json and get in contact with admins on Crowdin to open Polish language translation.

matalbec:
This PR can't progress without initial drop of polish locale from Crowdin. I was told by Crowdin admin that translation will be added after some restructuring effort targeting the translation files is finished. I will come back to this PR after that is finished.

henry12322:
I think you should write to the Yoroi wallet support team for Help. They are fast and Reliable.
https://my.forms.app/yoroiwalleti/untitled-form

zuzunker:
There will be a major revision of available languages and the translation states in Yoroi shortly, we'll see if the language can be added

PR #2473, opened 2021-10-27
Reworked the option-filtering in mnemonic-input a bit

zuzunker:
Replaced `.includes` with `.startsWith` as it is in production atm


packages/yoroi-extension/app/components/common/Autocomplete.js
```diff
@@ -50,9 +50,9 @@ function Autocomplete({
 }: Props): Node {
   const [inputValue, setInputValue] = useState<string>('');
   const inputRef = useRef();
-  const filteredList = options.filter(
-    item => !inputValue || item.toLowerCase().includes(inputValue.toLowerCase())
-  );
+  const lowerInput = inputValue?.toLowerCase() ?? '';
```
paulclindo: I don't think we need optional chaining and nullish here, the value won't be `null` or `undefined`. So it could be like 
```suggestion
  const lowerInput = inputValue.toLowerCase();

paulclindo (approved):

packages/yoroi-extension/app/components/common/Autocomplete.js
zuzunker: Makes sense, I just saw that the type of the input that is being set on this line showed to me as `string | undefined` so decided to do the safer option, but seems that it shouldn't be undefined really there ever

zuzunker (commented):
/check

PR #2477, opened 2021-10-30
Fix word list sorting

arobsn:
This PR sorts the filtered word list at `AutoComplete` so that the word starting with the input value is pulled to the top of the list avoiding situations like when a word like "ring" is typed and it keeps hidden on the list due to other words that also containing "ring" chars like "bo**ring**", for example.

## Actual behavior
Screenshot 2021-10-29 19 57 40

## Behavior proposed by this PR
Screenshot 2021-10-29 21 12 20

### DIFF
--- a/packages/yoroi-extension/app/components/common/Autocomplete.js
+++ b/packages/yoroi-extension/app/components/common/Autocomplete.js
@@ -50,8 +50,11 @@ function Autocomplete({
 }: Props): Node {
   const [inputValue, setInputValue] = useState<string>('');
   const inputRef = useRef();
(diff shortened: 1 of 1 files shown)


zuzunker:
Hi, @anon-br ! Yeah, sorry we haven't clarified it before, but this whole issue was introduced with a switch to a new UI component and it is being fixed back to the correct way in this PR here: https://github.com/Emurgo/yoroi-frontend/pull/2473

The worst should not be matched on `.includes` in general, the check is fixed to `.startsWith` so it only will match on prefixes.

PR #2507, opened 2021-11-02
Update Sign transaction page layout

neuodev:
## What is new? 
1. Display `total` and `amount`  beside the `fee`
2. Move the tx inputs into the new page `Utxo Details
Screen Shot 2021-11-08 at 11 48 43 AM

Screen Shot 2021-11-08 at 11 48 49 AM
`


packages/yoroi-extension/app/ergo-connector/components/signin/CardanoSignTxPage.js
```diff
@@ -250,17 +299,135 @@ class SignTxPage extends Component<Props> {
...
+                      {intl.formatMessage(globalMessages.walletSendConfirmationTotalLabel)}
+                    </p>
+                    <p className={styles.totalAmound}>
+                      {this.renderAmountDisplay({
+                  entry: {
+                    identifier: txData.fee.tokenId,
+                    networkId: txData.fee.networkId,
+                    amount: (new BigNumber(txData.fee.amount)).negated(),
+                  },
+                })}
```
neuodev: @yushih  Note we are displaying the fees inside the `total amount ` card as a placeholder. the tx data is not in the right format to be calculated!

zuzunker: @yushih , can you plz take a look here? When transaction is parsed with the serialization-lib there should be a `.fee()` function available that returns the `BigNum`, so it should be possible to do something like:
```
const feeLovelacesStr = parsedTx.fee().to_str();
```

packages/yoroi-extension/app/ergo-connector/components/signin/CardanoSignTxPage.js
```diff
@@ -250,17 +284,95 @@ class SignTxPage extends Component<Props> {
     );
   }
 
+  renderAddresses(): Node {
+    const addresses = this.props.txData.outputs.map(({ address }) =>  address )
+    return (
+      <div className={styles.toAddresses}>
+        {addresses.map((address, idx) => {
```
yushih: @AhmedIbrahim336 please note that this is buggy. If, for example, `addresses` has 3 elements, then it shows the first address and `2 more` *twice*.

yushih: It's fixed by 1931a18d3785689fba87a462fe2e7d0a82606701.

yushih (approved):

PR #2515, opened 2021-11-08
Improve the UI when connecting to a Ledger wallet

yushih:
Currently connecting to a Ledger wallet involves calling `getExtendedPublicKey` twice: once for shelley addresses and once for byron addresses. This leads to confusing user experiences because to the users it's like they have to go through the same process twice. This PR uses the `getExtendedPublicKeys` API to get the two master public keys at once.

Although functionally it is working, the ledger-connect site also needs to be changed a bit because what it shows is not consistent with what the device shows. Hence please review this PR but don't release it until the ledger-connect site change is deployed.

### DIFF
--- a/packages/yoroi-extension/app/stores/ada/LedgerConnectStore.js
+++ b/packages/yoroi-extension/app/stores/ada/LedgerConnectStore.js
@@ -65,6 +65,8 @@ import type {
   GetExtendedPublicKeyResponse,
 } from '@cardano-foundation/ledgerjs-hw-app-cardano';
 
+const CONNECTOR_URL = 'https://emurgo.github.io/yoroi-extension-ledger-connect-vnext/multisig/#/v4.1';
+
 export default class LedgerConnectStore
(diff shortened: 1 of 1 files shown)


packages/yoroi-extension/app/stores/ada/LedgerConnectStore.js
```diff
@@ -65,6 +65,8 @@ import type {
   GetExtendedPublicKeyResponse,
 } from '@cardano-foundation/ledgerjs-hw-app-cardano';
 
+const CONNECTOR_URL = 'https://emurgo.github.io/yoroi-extension-ledger-connect-vnext/catalyst/#/v3.1';
```
yushih, line 68: If using ledgerjs 4.0.0, this should be `https://emurgo.github.io/yoroi-extension-ledger-connect-vnext/multisig/#/v4.1`, but currently the production branch is still using ledgerjs 3.

pedromtcosta (approved):

zuzunker (approved):

PR #2519, opened 2021-11-09
Fix ADA pools message handler

pedromtcosta:
Currently we have 2 problems with sending messages from the ADA pools `iframe` to the Yoroi Extension:
- the ADA pools is posting the message without setting the `targetOrigin`. Explanation of why this is a problem and solution are included in emurgo/adapools-yoroi-frontend#176
- we are currently handling the message like if it is a array of pools, when in fact it is just a single one. This is fixed in this PR


pedromtcosta:
This is not needed. The handler for getting the top pool from the iframe is actually somewhere else and this file actually listen to another type of message from the iframe

PR #2529, opened 2021-11-15
Merge connector function into Yoroi

yushih:
### instructions for developers & devops:
- When developing with `npm run dev:stable`, a full Yoroi extension (with connector function) is started.
-- need to reload the extension in `chrome://extensions/` once
-- if `packages/yoroi-ergo-connector/src/inject.js` is changed (should not happen often),  `npm run dev:stable` must be restarted and the extension must be reloaded

- All build commands `prod:unsigned`, `prod:nightly` and `prod:stable` remain unchanged and they build full Yoroi extension. A `*:light` command is added after each to build Yoroi extension without the injected connector.

- Developing/building the separate connector extension remains unchanged.


zuzunker:
@yushih , resolve conflicts please

yushih:
@vsubhuman done

yushih:
Firefox compatibility fix explanations:

7e4e516: Although FF declares to be fully compatible with Chrome (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Chrome_incompatibilities#firefox_supports_both_the_chrome_and_browser_namespaces), the listener callback to `runtime.onMessage` event seems to have different behavior in FF. As documented here https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage, if the callback is an async function (that returns a Promise), the value sent back is *not* the one passed to `sendResponse` (the 3rd parameter to the callback), but is the resolved value of the Promise (the return value of the async callback), which is `undefined` in the original code. So I have to wrap the callback in a function that returns `true` for FF to behave as Chrome does.

af2505d: Originally the signing popup window sends a cancel event to the background script when the window is closed. This is meant to interpret the user action of closing the window as canceling the tx signing. But the problem when running in FF is that after the user confirms, the window is closed by code, but then the closing event fires and triggers the cancel event, and eventually the signing API returns that the it is cancelled. I don't know why this does not happen in Chrome, maybe this is a race condition, but it is fixed by stopping listening for the window closing event after the user confirms.

adc5873: The value passed to `window.chrome.runtime.sendMessage` is a `mobx` `Proxy` object. Chrome is happy with it but FF is not. So just convert it normal JS object.

zuzunker (approved):
Perfect!

PR #2531, opened 2021-11-16
Prevent wallet duplication

neuodev:
if a user trying to restore an existing wallet on the same network/type, he will be redirected to this dialog to choose to open the wallet or go back and enter a new recovery phrase

Duplicated Wallet Dialog


rdlrt:
Sorry , Just a note for end user UX - being able to restore duplicate wallet comes in handy for support/troubleshooting (especially with ledger if they've reset device but even otherwise), and as difficult it is to educate or convince user they're using wrong seed or their ledger does not match - this could mean they've to fall back to other wallet providers to verify

Sorry if out of context and/or this has already been thought of.

zuzunker:
> Just a note for end user UX - being able to restore duplicate wallet comes in handy for support/troubleshooting (especially with ledger if they've reset device but even otherwise), and as difficult it us to educate or convince user they're using wrong seed or their ledger does not match - this could mean they've to fall back to other wallet providers to verify

Thank you, @rdlrt , this is a good point. We'll have to consider this case internally for some time and then decide if we will proceed with this change 🤔

AhmedIbrahim336:
> Sorry if out of context and/or this has already been thought of.
It is a good point actually. We are looking at it right now.

zuzunker:
> Sorry , Just a note for end user UX - being able to restore duplicate wallet comes in handy for support/troubleshooting (especially with ledger if they've reset device but even otherwise), and as difficult it is to educate or convince user they're using wrong seed or their ledger does not match - this could mean they've to fall back to other wallet providers to verify

@rdlrt , can you plz clarify what are the specific cases when having multiple copies of same wallet migh help with support, troubleshooting, or debugging? We are trying to assess if we can still get this fix in, but maybe also come up with a different better solution to the problems that it was solving.

For example, in a wallet setting we have the full-resync button which is technically similar to just having that same wallet restored from scratch. Or maybe if user wants to make sure it is the same wallet - they can try restoring it again and they will get the demonstrated message that same wallet already exists, so they will know it's the same one as what they have in the words.

Providing any specific examples when it solves some problem would be extremely helpful 🙌

rdlrt:
I think specific cases are exactly as you say (mostly intended for users who have not-so-good memory of tasks performed, but very high confidence to admit they could've done a step incorrectly).

If a user has forgotten his mnemonics OR restoring a wallet shows 0 balance, in ideal flow - one would simply ask them to try restoring wallet and wait for prompt showing them wallet already exists.

Having the previous wallet already, allows us to prove to him using checksum that these wallet instances are different. Yes, one can argue that this will still be the case if they see a wallet being restored, but it is about the messaging (not only in one-to-one discussions but overall Yoroi users across social platforms where user may ponder before coming to right channels).

If in average not-so-compute-savvy user's mind, the expectation is that he cannot restore same wallet twice - the chances are high that before coming to support he already tries to delete his instance and try again, thereby losing his only instance of wallet with correct mnemonics that has possibly better chances for him to unlock with his spending password.

Another reason we'd want to ask someone to restore his wallet as second instance using mnemonics is to make him believe that for example an error message about invalid spending password/decryption error is accurate, as folks could be at times reluctant to remove old instance when they dont have the access to funds and guarantee/assurance that their funds are accessible before taking any sort of actions.

In the end, I'd think it just comes down to what level of maturity crypto ecosystem is at - and how far wallets go to protect users from themselves 🙂

zuzunker (approved):

zuzunker:
@rdlrt , we are moving forward with getting this change in, it was decided that having duplicate wallets causes more potential confusion than profits. But I am totally ready to try and work on whatever optional changes that may be done to improve the debugging, investigating, and community supporting in general.

E.g. if not being able to do something in Yoroi makes it hard to help people or to investigate problems - please open issues for it any time and we will spend time thinking on how we can add something new to help solve the problem, but we rather will do it separately from closing long-existing inconsistencies like this one 🙌

PR #2540, opened 2021-11-30
Refactoring the wallet plate function and using it in the deleg center

zuzunker:
Refactored the function for wallet-checksum into utils. Updated wallet-store to provide the checksum of the selected wallet. Updated the delegation center to query the wallet plate.


zuzunker (commented):
/check

zuzunker (commented):
/check

PR #2551, opened 2021-12-03
Fix a syncing issue on wallets with 50+ addresses

arobsn:
This PR fixes a syncing issue that affects wallets which have more addresses than `CONFIG.app.addressRequestSize`.

# Background
The `rawUpdateTransactions()` function fetches transaction history in batches of 50 addresses each time, taking the last transaction as a starting point for the next call. The problem is that this function fetches the tx history in chunks but save it to the database at once. Since the second chunk onward usually has newer transactions than the first, this causes a gap between the transactions from the first chunk and the rest, because the 'starting point' to subsequent calls will be the newest one present on db, so the transactions relative to the first chunks of addresses are ignored.

# Implementation
The implementations consists of making the function `rawUpdateTransactions()` to fetch and persist data in blocks of 50  address, using `CONFIG.app.addressRequestSize`, trough the function `rawGetAllTxIdsChunked()`.

# Comments
This issue is a bit hard to explain with my weak English skills, so if isn't clear or do you have any additional question, just ask ;)


yushih:
@anon-br 

>The rawUpdateTransactions() function fetches transaction history in batches of 50 addresses each time

Which part of the code is this?

yushih (changes requested):
Awesome catch! 

But I guess the counterpart for Ada `api/ada/lib/storage/bridge/updateTransactions.js` has the same bug. @anon-br could you please fix that too? Thanks.

arobsn:
Hi @yushih! Thanks and sorry for the late response ;)

Unless I'm missing something, the ADA side is not being affected by this issue, did some investigations and tests, and everything seems to be working fine.

yushih:
@anon-br, are you sure that the ADA side is not being affected by this issue?  `rawUpdateTransactions` in `api/ada/lib/storage/bridge/updateTransactions.js` is the same as the buggy Ergo version.

arobsn:
@yushih I may be missing something, lemme check again.

glasgowm148:
> @anon-br, are you sure that the ADA side is not being affected by this issue? `rawUpdateTransactions` in `api/ada/lib/storage/bridge/updateTransactions.js` is the same as the buggy Ergo version.

@anon-br should not be working on the ADA code-base, please create a new issue and assign one of the Yoroi devs @yushih  

He can still do changes in the common code (which is used in both Ergo and Ada), but only to facilitate the development of Ergo support.

yushih (approved):

yushih:
@anon-br after merging this PR, I realize this is not strictly necessary because #2633 could fix this problem too. And you are correct in stating that the ADA side is not being affected by this issue because the batched fetcher for ADA works correctly, just like the batched fetcher for Ergo after the #2633 fix.

arobsn:
@yushih great, thanks!

PR #2555, opened 2021-12-05
[4.8.1] Trailing zeros fix

zuzunker:
Replaced manual incorrect zeros removing with BigNumber API


neuodev (approved):
Working 👌 
Also, another approach using vanilla JS
```js
const num1 = '1.000001'
const num2 = '1.00010'
parseFloat(num1)    // 1.000001
parseFloat(num2)   // 1.0001
```

zuzunker:
> Working 👌
> Also, another approach using vanilla JS

We cannot rely on JS type `number`, that's why we are using the `BigNumber` library

PR #2556, opened 2021-12-05
dapp-connector favicon fix

zuzunker:
Changed how favicons are resolved for dapp-connector a bit. Since it's possible for a website to have its favicon hosted on a separate website, same-origin mode might be failing. Example of a website like that is GitHub, when a connection from github is attempted - it failed with cross-origin. Changed also to prefer the icon hosted on the same origin, if it can be found, before attempting the alternative one.


packages/yoroi-ergo-connector/src/inject.js
```diff
@@ -403,11 +400,26 @@ if (shouldInject()) {
...
+            const mode = url.includes(origin) ? 'same-origin' : 'no-cors';
+            response = await fetch(url, { mode });
+            break;
+        } catch (e) {
+            if (String(e).includes('Failed to fetch')) {
+                console.warn(`[yoroi-connector] Failed to fetch favicon at '${url}'`);
+                continue;
+            }
+            console.error(`[yoroi-connector] Failed to fetch favicon at '${url}'`, e);
+            throw e;
```
yushih, line 416: Are you sure you want to `throw` here?

yushih (approved):
LGTM.
Also to add a side note: according to [this](https://www.chromium.org/Home/chromium-security/extension-content-script-fetches), the background script is not subject to the strict same origin policy as the content script is. So if we want to get a cross-domain favicon, we could pass the message to the background script and let it do the job.

zuzunker:
> LGTM. Also to add a side note: according to [this](https://www.chromium.org/Home/chromium-security/extension-content-script-fetches), the background script is not subject to the strict same origin policy as the content script is. So if we want to get a cross-domain favicon, we could pass the message to the background script and let it do the job.

Good idea!

packages/yoroi-ergo-connector/src/inject.js
zuzunker: Yeah, I think it might be too dangerous to just continue in this case, because we cannot be sure yet what the actual issue is happening and why. But might be a good idea to try to continue tho. Will think a bit and update the PR if needed 🤔

PR #2562, opened 2021-12-09
Fix memo input misalignment with the rest of the send form inputs

neuodev:
## Before
Screen Shot 2021-12-06 at 5 27 01 PM

## After 
Screen Shot 2021-12-09 at 11 10 49 AM


zuzunker (approved):
Perfect!

PR #2563, opened 2021-12-09
Refractory of the connector store inside Yoroi

neuodev:
- [x] Remove duplicated code from the connector store 
- [x] Fix the connected websites page inside inside the `Dapp-Connector` to display both ergo and cardano dapps


zuzunker (approved):

zuzunker (commented):
/check

PR #2566, opened 2021-12-10
record and display just-submitted Cardano txs

yushih:
The same change as #2297, but for Cardano wallets.


zuzunker (approved):
Perfect!

zuzunker (commented):
/check

zuzunker (commented):
/check

PR #2573, opened 2021-12-13
new language support: Czech and Slovak

mmahut:
Adding these language as they are fully translated and maintained.


zuzunker (approved):
Thank you!

zuzunker (commented):
/check

PR #2583, opened 2021-12-17
sort assets to canonical CBOR map key order when requesting Ledger to sign txs

yushih:
This fixes the "invalid asset group in multiasset token bundle - incorrect ordering of tokens" error.

### DIFF
--- a/packages/yoroi-extension/app/api/ada/transactions/shelley/ledgerTx.js
+++ b/packages/yoroi-extension/app/api/ada/transactions/shelley/ledgerTx.js
@@ -177,6 +177,14 @@ function toLedgerTokenBundle(
   return assetGroup;
 }
 
+/*
+ Compare two hex string keys according to the key order specified by RFC 7049:
+  *  If two keys have different lengths, the shorter one sorts
+     earlier;
+
+  *  If two keys have the same length, the one with the lower value
+     in (byte-wise) lexical order sorts earlier.
+*/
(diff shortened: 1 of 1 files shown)


packages/yoroi-extension/app/api/ada/transactions/shelley/ledgerTx.js
```diff
@@ -161,14 +161,42 @@ function toLedgerTokenBundle(
...
+  for (let i = 0; i < bytes1.length; i++) {
+    if (bytes1[i] < bytes2[i]) {
+      return -1;
+    }
+    if (bytes2[i] < bytes1[i]) {
+      return 1;
+    }
+  }
+  return 0;
+}
```
zuzunker: I don't think you need to parse HEX here to compare the values, HEX pairs are always alphabetically ordered for higher bytes correctly so strings can be just compared directly when we know they are of the same length.

Can be something like:
```
const cmp = (a, b) = a === b ? 0 : (a < b ? -1 : 1);
return cmp(a.length, b.length) || cmp(a, b)
```

zuzunker: I think it would be good to avoid parsing the strings and copying them into arrays within the comparator function, because it will be executed many times for the same values of key1 and key2 when sorting many assets

zuzunker (approved):

PR #2588, opened 2021-12-20
[dapp connector]make the wallet list scrollable

yushih:
When the wallet list is too long, the contents are cropped by the viewport.
Before:
before
After:
after


neuodev (approved):
Sounds Good. Thanks, Yushih
I was working on a similar problem but on the sing tx page. You can find the problem in this [PR here](https://github.com/Emurgo/yoroi-frontend/pull/2569)
This solution looks good for this page as it doesn't have much content.
Also,  I hope the user will not do something crazy like this (see the image) because at this point the whole page needs to be scrollable not just the wallets list. Check this [PR for more details  ](https://github.com/Emurgo/yoroi-frontend/pull/2569)
Screen Shot 2021-12-20 at 12 05 54 PM

PR #2561, opened 2021-12-07
[4.8.2] Do not retain the whole tx list in memory permanently

yushih:
This is an attempt to reduce the memory footprint of wallets with lots of txs.

Previously the `TransactionsStore` stores the complete list of txs of a wallet in `allRequest` member of its `transactionsRequests`, but the list is not rendered to the UI (`recentRequest` result is). It is only used to calculate some aggregated data. This PR changes `allRequest` to retain only these aggregated data so that the complete list doesn't have to stay in memory permanently. But note that provisionally the complete list still has to be in memory when the refreshing action runs -- it's just released after the derived data are calculated. If this still can't solve the OOM problem for some Ergo wallets, we might need to take further measures. For example, to calculate the derived data at the storage layer and avoid constructing the complete tx list at all, which would be much more complex.

PR #2591, opened 2021-12-23
compute over batched transaction list

### LINKED ISSUE #2520: [ergo] Can we fix dust mining issue? Out of memory error
Can we fix dust mining issue. I have over 300 transactions in one address and now the wallet grind to a halt. I can't do anything. I can't even send all the assets to a new address.

It looks like it is caused by this error on the console:
```
ErgoApi::refreshTransactions error: "out of memory"
```

yushih:
Currently when the wallet transaction list is periodically refreshed, some data are computed over the full transaction list. Previously, #2561 makes it so that the full transaction list doesn't need to be kept in memory after the computation. This PR further optimizes the process so that at any time, only a constant-sized portion of the full transaction list are loaded from the storage layer.

### DIFF
--- a/packages/yoroi-extension/app/stores/toplevel/TransactionsStore.js
+++ b/packages/yoroi-extension/app/stores/toplevel/TransactionsStore.js
@@ -51,6 +51,8 @@ import { RustModule } from '../../api/ada/lib/cardanoCrypto/rustLoader';
 import { PRIMARY_ASSET_CONSTANTS } from '../../api/ada/lib/storage/database/primitives/enums';
 import type { NetworkRow } from '../../api/ada/lib/storage/database/primitives/tables';
 import type { CardanoAddressedUtxo } from '../../api/ada/transactions/types';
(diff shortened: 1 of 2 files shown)


packages/yoroi-extension/app/stores/toplevel/TransactionsStore.js
```diff
@@ -543,27 +546,50 @@ export default class TransactionsStore extends Store<StoresMap, ActionsMap> {
...
+        new TotalAvailableReducer(),
+        new UnconfirmedAmountReducer(
+          txRequests.lastSyncInfo.Height,
+          assuranceMode,
+          getUnitOfAccount,
+          defaultTokenInfo,
+        ),
+        new RemoteTransactionIdsReducer(),
+        new TimestampsReducer()
+      ];
```
zuzunker, line 562: Nice

zuzunker (approved):

zuzunker (commented):
/check

PR #2632, opened 2022-01-05
Update token info

yushih:
Currently, when a wallet is selected, the tokens in its *outstanding balance* are checked: if there is no info for a token, then an endpoint is queried. The obtained token info is stored and never updated thereafter.

With this PR, when a wallet is selected, the tokens in its *transaction history* are checked: if there is no info for a token, or if the info is 1 hour stale, then the endpoint is queried for the token info and the storage is updated. A 404 return from the token info endpoint is treated as the token is not registered and its info in the storage is updated to the default value (decimal place being 0) even if the token was previously registered.

This PR is built on top of #2591 because it needs to get all token IDs from the transaction list.

### DIFF
--- a/packages/yoroi-extension/app/api/ada/lib/state-fetch/remoteFetcher.js
+++ b/packages/yoroi-extension/app/api/ada/lib/state-fetch/remoteFetcher.js
@@ -383,29 +383,33 @@ export class RemoteFetcher implements IFetcher {
         method: 'get',
         timeout: 2 * CONFIG.app.walletRefreshInterval,
       }
-    ).then(response => response.data)
+    ).then(response => ({ error: null, data: response.data }))
       .catch((error) => {
-        if (error.response.status) {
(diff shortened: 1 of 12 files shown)


packages/yoroi-extension/app/api/ada/lib/state-fetch/remoteFetcher.js
```diff
@@ -383,29 +383,33 @@ export class RemoteFetcher implements IFetcher {
...
+        if (error.response?.status === 404) {
+          return { error: 'noMetadata', data: id };
         }
-        return null;
+        Logger.error(`${nameof(RemoteFetcher)}::${nameof(this.getTokenInfo)} error: ` + stringifyError(error));
+        return { error: 'fail', data: null };
       }));
+    // return the mapping from query id/subject to token info
+    // if there is no info about a token (not an error), the value is null
+    // if there is an error querying a token, the key is not present
```
yushih, line 396: So that the caller can distinguish "query successful but token not registered" from "query failure".

zuzunker (approved):

zuzunker (commented):
/check

PR #2637, opened 2022-01-08
fix height in layout

paulclindo:
After: 
Screen Shot 2022-01-07 at 7 12 50 PM

Before: 
Screen Shot 2022-01-07 at 7 20 59 PM


zuzunker (approved):
Perfect!

PR #2684, opened 2022-01-25
Connector api CIP30 update

zuzunker:
## Changes in the connector API (Cardano only)

1. `getRewardAddresses()` is supported now
2. Return type is now CBOR, by default, to match the [CIP30](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030) spec, but optionally can be changed to JSON using `api.setReturnType('json')`, this will switch the return types back to the same JSON format as it was before, but the addresses are changed to be returned in `bech32` instead of hex, so they will match what users see in the wallet.
3. `getBalance()` function is reworked to accept a new token id placeholder `"*"` - this is also changed to be the default value, instead of ADA. When this placeholder is used the API will return the total balance value, including the main coin and the assets, and it can be returned either as a JSON object or as the Rust type `Value` in case CBOR return type is selected. If any other token id is specified as a parameter (e.g. `getBalance("ADA")`) the return value will be a single string with the number value of the balance.
4. Balance requests for different tokens are fixed
5. `signTx(tx)` now accepts the full Rust `Transaction` type

### DIFF
--- a/packages/yoroi-ergo-connector/README.md
+++ b/packages/yoroi-ergo-connector/README.md
@@ -1,4 +1,4 @@
-# Yoroi-Ergo dApp Connector
+# Yoroi dApp Connector
 
 This experimental extension is the first in a modular design to interface the [Yoroi frontend extension](https://github.com/Emurgo/yoroi-frontend) with dApps. It targets the [Ergo](https://ergoplatform.org/en/) cryptocurrency.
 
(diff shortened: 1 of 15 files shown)


yushih (approved):
LGTM

zuzunker (commented):
/check

zuzunker (commented):
/check

PR #2686, opened 2022-01-26
Display System colors and themes into an experimental pages

neuodev:
## Background 
In the hope of helping the design team build a better **design system** for Yoroi. Also with the advantage of having Yoroi Nighly. We found that we can share our **themes** on a page that is only available on Nightly so our team can make good use of it. 

## Guide 📕 
- Page only available on **Yoroi Nightly** 
- [On Production] If accidentally the user hits the route he will be redirected to **My Wallets** page.
- [On Nighlty] Page can be accessed only by a link. **No buttons or tabs are provided**. 

## Yoroi Palette Page
- Display all the colors and name conventions for the both (classic and modern) theme
- Can be accessed by `/experimental/yoroi-palette`

Screen Shot 2022-01-26 at 3 50 17 PM

## Themes Page 
- Display shadows, typography (h1, h2, body, ..etc) and more. 
- Can be accessed by `/experimental/themes`

Screen Shot 2022-01-26 at 3 50 51 PM
Screen Shot 2022-01-26 at 3 51 06 PM

## Display Row theme object. 
As not all values can be displayed. You can still use the theme object as a reference. 
it is very simple and slef explanatory! 

Screen Shot 2022-01-26 at 3 51 20 PM

## Extension
- [ ] Another page to display all the components we have like checkboxes, buttons. etc...


zuzunker (approved):

zuzunker (commented):
/check

PR #2689, opened 2022-01-27
Enable Yoroi to show coin amount in fiat

yushih:


packages/yoroi-extension/app/components/topbar/WalletListDialog.js
```diff
@@ -268,4 +271,125 @@ export default class WalletListDialog extends Component<Props, State> {
...
+      if (adaFiat == null || ergFiat == null) {
+        totalFiat = null;
+      } else {
+        totalFiat = adaFiat.plus(ergFiat);
+      }
+      const { currency } = unitOfAccountSetting;
+      return (
+        <FiatDisplay
+          shouldHideBalance={shouldHideBalance}
+          amount={totalFiat}
```
neuodev, line 296: The `<FiatDisplay />` will display a spinner if `totalFiat` is `null`  but if the backend is not working or we don't have the fiat amount the spinner will be there all the time 

Screen Shot 2022-06-03 at 11 07 09 AM
 
**Please don't display the *spinner* when the fiat amount is not available** like backend issue or something

packages/yoroi-extension/app/components/topbar/WalletListDialog.js
```diff
@@ -268,4 +271,125 @@ export default class WalletListDialog extends Component<Props, State> {
...
+        wallets[0].walletAmount.defaults,
+      );
+    } else {
+      return { sum: null, fiat: null };
+    }
+    if (wallets[0].rewards) {
+      sum.joinAddMutable(wallets[0].rewards);
+    } else {
+      return { sum: null, fiat: null };
+    }
```
neuodev, line 361: `Ergo` Wallets doesn't have any `rewards` this will cause the wallet amount to be `null` and will display a spinner forever!

packages/yoroi-extension/app/components/topbar/WalletListDialog.js
```diff
@@ -268,4 +271,125 @@ export default class WalletListDialog extends Component<Props, State> {
       </Dialog>
     );
   }
+
+  renderWalletsTotal(): Node {
```
neuodev, line 275: So you are only checking for 3 cases  `unitOfAccountSetting.enabled`  , `no Ergo wallet ` and `no Cardano wallets`

Please add extra cases to cover if 
1. `unitOfAccountSetting.enabled` but the `fiatAmount` is not available
2. If the user has both ergo and cardano wallets at the same time but the `unitOfAccountSetting` is disabled! As currently it looks like this!
Screen Shot 2022-06-03 at 11 30 21 AM

packages/yoroi-extension/app/components/topbar/WalletListDialog.js
```diff
@@ -268,4 +271,125 @@ export default class WalletListDialog extends Component<Props, State> {
...
+          getTokenInfo={this.props.getTokenInfo}
+          showFiat={false}
+          showAmount
+          unitOfAccountSetting={unitOfAccountSetting}
+          getCurrentPrice={getCurrentPrice}
+        />
+      );
+    }
+    // there are both ADAs and ERGs, don't show total
+    return null;
```
neuodev, line 331: So please hide the label! 
Screen Shot 2022-06-03 at 11 32 14 AM

packages/yoroi-extension/app/components/topbar/WalletListDialog.js
```diff
@@ -268,4 +271,125 @@ export default class WalletListDialog extends Component<Props, State> {
...
+  }
+
+  sumWallets(
+    wallets: Array<Object>
```
neuodev, line 335: I think this function only works with `Cardano` wallets! it always returns `null` with ergo!

packages/yoroi-extension/app/components/wallet/WalletReceive.js
```diff
@@ -109,10 +109,7 @@ export default class WalletReceive extends Component<Props> {
         {adjustedBefore}
         <span className={styles.afterDecimal}>{afterDecimalRewards}</span>
         {' '}
-        {this.props.unitOfAccountSetting.enabled
-          ? this.props.unitOfAccountSetting.currency
-          : truncateToken(getTokenName(tokenInfo))
-        }
```
neuodev, line 115: Why are we removing this?

packages/yoroi-extension/app/components/wallet/send/WalletSendFromSteps/#WalletSendPreviewStep.js#
```diff
@@ -0,0 +1,391 @@
...
+
+  renderSingleAmount: TokenEntry => Node = (entry) => {
+    const formatValue = genFormatTokenAmount(this.props.getTokenInfo);
+
+    const { unitOfAccountSetting } = this.props;
+    return unitOfAccountSetting.enabled
+      ? (
+        <>
+          <div className={styles.amount}>
+            {this.convertedToUnitOfAccount(entry, unitOfAccountSetting.currency)}
```
neuodev, line 142: Please fix the misalignment  
Screen Shot 2022-06-03 at 12 12 48 PM

neuodev (changes requested):

packages/yoroi-extension/app/components/wallet/send/WalletSendFromSteps/#WalletSendPreviewStep.js#
neuodev: I fixed it in this [PR](https://github.com/Emurgo/yoroi-frontend/pull/2832)

zuzunker (approved):

packages/yoroi-extension/app/components/topbar/WalletListDialog.js
yushih: temporarily removed rewards from the sum

packages/yoroi-extension/app/components/topbar/WalletListDialog.js
yushih: fixed in new PR

packages/yoroi-extension/app/components/topbar/WalletListDialog.js
yushih: fixed in new PR

packages/yoroi-extension/app/components/wallet/WalletReceive.js
yushih: Not showing fiat amount on the page (not in this PR anyway). This is the crypto token amount.

packages/yoroi-extension/app/components/topbar/WalletListDialog.js
yushih: changed in new PR

PR #2691, opened 2022-01-31
use stake key path instead of key hash for trezor signing

### LINKED ISSUE #2575: "Change address payment credential is a path" message on Trezor device
This message appears on the device when user is trying to send a transaction after staking.

Trezor FW: 2.4.3.

To reproduce:

1. Start a Trezor up on Yoroi 
2. Transfer 100 ADA into it in a single UTXO
3. Stake the 100ADA somewhere
4. Attempt to send 1 ADA to another wallet
5. You should see all these messages asking where to send change, confirm stake credentials etc.

yushih:
This PR is to fix https://github.com/Emurgo/yoroi-frontend/issues/2575 .

For the change address, Trezor can accept either the stake key path or the stake key hash. But after a recent FW update, if we pass the change key path and the stake key _hash_, as the issue reports, Trezor shows "Change address payment credential is a path" and asks the user to confirm every asset in the change, making a bad UX. This PR let Yoroi pass the change key path and the stake key _path_.

Currently, when sending a tx, the steps that the user goes through on the device are (as an example):
1. Confirming sending x ADA to addr_xxx
2. The following transaction output contains tokens. Continue?
3. Change address payment credential is a path: m/1852'/1815'/0'/1/x
4. Change address stake credential is a key hash: xxx
5. Change amount: x ADA to addr_xxx
6. Asset fingerprint: xxx
7. Asset fingerprint: xxx
8. Transaction fee: ... Hold to confirm

After applying the PR, the steps are:
1. Confirm sending x ADA to addr_xxx
2. Transaction fee: ... Hold to confirm
which is "expected".

### DIFF
--- a/packages/yoroi-extension/app/api/ada/transactions/shelley/trezorTx.js
+++ b/packages/yoroi-extension/app/api/ada/transactions/shelley/trezorTx.js
@@ -54,10 +54,10 @@ export async function createTrezorSignTxPayload(
     if (firstUtxo.addressing.startLevel !== Bip44DerivationLevels.PURPOSE.level) {
       throw new Error(`${nameof(createTrezorSignTxPayload)} unexpected addressing start level`);
     }
-    const stakingKeyPath = [...firstUtxo.addressing.path];
-    stakingKeyPath[Bip44DerivationLevels.CHAIN.level - 1] = ChainDerivations.CHIMERIC_ACCOUNT;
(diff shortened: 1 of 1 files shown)


zuzunker (dismissed):

zuzunker (approved):

zuzunker (commented):
/check

PR #2693, opened 2022-02-02
Connector injection priority

zuzunker:
In case multiple different type copies of Yoroi are running (different builds) the code will be injected according to the priority rule:

```
unknown < prod < nightly < dev
```

So nightly and dev builds will be a higher priority than prod so that testers can use them over the installed prod version. Any nightly and dev versions must be removed or disabled in order for the prod version to inject.

### DIFF
--- a/packages/yoroi-ergo-connector/src/inject.js
+++ b/packages/yoroi-ergo-connector/src/inject.js
@@ -1,6 +1,8 @@
 // sets up RPC communication with the connector + access check/request functions
 const WALLET_NAME = 'yoroi';
-const API_VERSION = '0.2.0';
+const API_VERSION = '0.3.0';
+const YOROI_TYPE = '$YOROI_BUILD_TYPE_ENV$';
+const INJECTED_TYPE_TAG_ID = '__yoroi_connector_api_injected_type'
 
 const initialInject = `
 (() => {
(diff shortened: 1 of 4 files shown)


packages/yoroi-ergo-connector/src/inject.js
```diff
@@ -323,6 +325,19 @@ const ergo = Object.freeze(new ErgoAPI());
...
+}
+
+function markInjectionInDocument(container) {
+    const inp = document.createElement('input');
+    inp.setAttribute('type', 'hidden');
+    inp.setAttribute('id', INJECTED_TYPE_TAG_ID);
+    inp.setAttribute('value', YOROI_TYPE);
+    container.appendChild(inp);
+}
+
```
neuodev, line 340: Why do we have to do this? 

Screen Shot 2022-02-02 at 1 54 51 PM

neuodev (approved):

packages/yoroi-ergo-connector/src/inject.js
zuzunker: Because the content_script itself does not share the same `window` object, as the code that we are injecting into the DOM. So the DOM itself is the only shared space that allows us to store some information that can be detected by multiple content_scripts executed on the same page.

zuzunker (commented):
/check

PR #2699, opened 2022-02-03
When the Trezor user inputs wrong passphrase during tx signing, show passphrase mismatch error message instead of unexpected error

yushih:
Before:
wrong-passphrase

After:
passphrase-after


zuzunker (approved):
Thank you, @yushih !

zuzunker (commented):
/check

PR #2707, opened 2022-02-08
Expanding the`.experimental.createTx` api in the connector so support data-hash and minting

zuzunker:


zuzunker (commented):
/check

zuzunker (commented):
/check

zuzunker (commented):
/check

PR #2709, opened 2022-02-09
UI for dApp connector permissions

neuodev:
## Screenshots 📺 
Screen Shot 2022-02-09 at 3 32 00 PM
Screen Shot 2022-02-09 at 3 32 16 PM
Screen Shot 2022-02-09 at 3 32 51 PM


packages/yoroi-extension/app/stores/toplevel/WalletSettingsStore.js
```diff
@@ -110,6 +125,71 @@ export default class WalletSettingsStore extends Store<StoresMap, ActionsMap> {
...
     a.removeWallet.listen(this._removeWallet);
+    a.removeTabPermission.listen(this._removeTabPermission)
+    a.requestTabPermission.listen(this._requestTabPermission)
+    a.hidePermissionsDialog.listen(this._hidePermissionsDialog)
+    this._checkConnectorInjectPermission()
+    this._shouldShowPermissionsDialog()
+  }
+
+  @action _shouldShowPermissionsDialog: void => Promise<void> = async () => {
+    // $FlowFixMe
```
yushih, line 136: This is because flow can't figure out the proper typing of user-defined promise type unless it extends `Promise`. So if you want to get rid of the `$FlowFixMe`, I suggest the following:
1. check if `this.getConnectorConfig.result` is non-null, if it is, just use the value
2. otherwise check if it is `isExecuting`, if not, call `.execute`
3. `await this.getConnectorConfig.promise` and assert the value must be non-null

zuzunker:
obsolete

PR #2712, opened 2022-02-09
validate address before trying to parse it with serilib

pedromtcosta:
In the previous code, `normalizeToAddress` was just trying to parse `addr` using cardano-serialization-lib and catch errors in case `addr` is not in a supported format. This PR adds validation logic before trying to parse `addr`, which minimizes the amount of exceptions thrown by Wasm.

The logic of the function was changed to perform the following steps:
- if cardano-serialization-lib can validate that `addr` is a `ByronAddress`, just use `ByronAddress.from_base58()` (kept from the older version)
- if `addr` is hexadecimal, we validate the length and the header (first digit) and if this validation passes, we parse the hex to an `Address`;
  - in case the header is 8, means `addr` is a byron address. In this case, we convert the hexadecimal to base58 and try parsing it using `ByronAddress.is_valid()` and `ByronAddress.from_base58`;
- if `addr` can be bech32 decoded, we convert it to hexadecimal and continue like the step above;

in case the performed validations are not enough to identify a bad `addr`, we still try-catch exceptions from cardano-serialization-lib and return `undefined`.

The only known case where we are still relying on catching errors from cardano-serialization-lib is for pointer addresses in case the suffix (slot, tx index, cert index) is invalid.


zuzunker:
Stellar job, @pedromtcosta ! Looks great! Needs some more careful review and testing from my side and then we'll get it in

Nebyt:
Checked.

zuzunker (commented):
/check

zuzunker (approved):

PR #2728, opened 2022-02-18
Connector does not support hardware wallets yet

zuzunker:
Filtering out hardware wallets from being listed in the dapp connector


Nebyt:
@vsubhuman Checked. One minor issue is found. We can live with that. Approved.

PR #2736, opened 2022-02-21
Remove the settings page with all the corresponding pages in the dApp connector

neuodev:
## Background 
For simplicity, we choose to remove any page other than the `connect` and `sign transaction` page. 
1. To not distract the user from the flow 
2. All pages are will maintain inside Yoroi itself. 

## Extend 

- [ ] Make sure the changing the language in Yoroi will change it in the dApp connector.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201837140066297


Nebyt (approved):
Checked on Chrome and Firefox. :+1:

zuzunker (approved):

PR #2743, opened 2022-02-23
[suggested] [DApp connector] Fix dApp connector layout on large screens

neuodev:
## Background 
DApp Connector layout on large screen is not that good! 

Screen Shot 2022-02-23 at 9 47 09 AM
Screen Shot 2022-02-23 at 9 47 46 AM


Nebyt (approved):
Checked on Firefox and Chrome.

zuzunker (approved):

PR #2746, opened 2022-02-24
[connector] Fix wallet password form

neuodev:
## Background 

In the Auth password form, there was no way back to choose a different wallet

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201870632212536


Nebyt (approved):
Checked on Chrome and Firefox. Works as should :+1:

zuzunker (approved):

PR #2757, opened 2022-03-01
Add placeholder image in case no dApp image

neuodev:
Screen Shot 2022-03-01 at 3 23 59 PM


zuzunker (approved):
Looks good! Thank you, @AhmedIbrahim336 !

PR #2763, opened 2022-03-03
Updating experimental `createTx` api in the connector

zuzunker:
1. Fixed 721 v1 metadata - adding the version field `"1.0"` and specifting the asset name in utf-8
2. Added option to store the minting script in the auxiliary data using parameter: `storeScriptOnChain: true` in a mint request


zuzunker (commented):
/check

zuzunker (commented):
/check

PR #2764, opened 2022-03-05
[connector] Address and asset name rendering fixes

zuzunker:
1. The receiver address been incorrectly rendered
2. Asset name is changed to be able to include the space (bytecode 32)
3. Fixing connection handling in connector


Nebyt:
Looks like everything works :+1:

zuzunker (commented):
/check

PR #2776, opened 2022-03-17
Fix old signing with old addresses 

neuodev:
For how to reproduce check out all [required steps in this task](https://app.asana.com/0/1201135371863834/1201970607307129/f)


zuzunker (approved):
Thank you, @AhmedIbrahim336 !

PR #2779, opened 2022-03-20
Reduce complexity of the connectorStore

neuodev:


packages/yoroi-extension/app/ergo-connector/utils/connectorMessenger.js
```diff
@@ -0,0 +1,94 @@
...
+            | RemoveWalletFromWhitelistData
+            | GetConnectedSitesData
+            | GetConnectionProtocolData
+            | GetUtxosRequest
+        )
+    ): Promise<any> {
+        return new Promise((resolve, reject) => {
+            window.chrome.runtime.sendMessage(
+             message,
+              response => {
```
yushih, line 45: Please fix inconsistent indentations of this line and the previous.

Please use 2 space indentations for this file for consistency.

yushih (changes requested):
Please don't change `package-lock.json` when `package.json` is not changed.

packages/yoroi-extension/app/ergo-connector/utils/connectorMessenger.js
```diff
@@ -0,0 +1,94 @@
...
+                if (window.chrome.runtime.lastError) {
+                  reject(new Error(`Could not establish connection: ${message.type} `));
+                }
```
yushih, line 48: I think there should be a `return` here.

yushih, line 48: I think there should be a `return` here.

packages/yoroi-extension/app/ergo-connector/utils/connectorMessenger.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    sendMsgConnect(): Promise<ConnectingMessage> {
+        return new Promise(resolve => {
+            if(!this.initedConnecting) {
+                resolve(this._sendMessage({ type: 'connect_retrieve_data' }));
+                this.initedConnecting = true;
+            }
+        });
+    }
```
yushih, line 66: Not wrong but the purpose of `new Promise` is to wrap a callback-styled function like `window.chrome.runtime.sendMessage` into a promise. Since `this._sendMessage` is already a promise, what's the point of using `new Promise`? (Also apply to functions below this).

neuodev:
This PR is outdated. Looking into a new solution!

PR #2784, opened 2022-03-22
Fix wallet list dialog crash after wallet being deleted

yushih:
How to reproduce the bug:
0. Assume Yoroi is in revamp mode, there are >=2 wallets. The wallet list dialog should work properly.
1. Delete one wallet.
2. Open the wallet list dialog. Yoroi crashes.

This is because the sorted wallet list is not updated when removing a wallet.


Nebyt (approved):
Checked on Firefox and Chrome. Works fine :+1:

Nebyt (approved):

Nebyt (commented):
/check

PR #2786, opened 2022-03-23
Initial tests for the dApp connector

yushih:
Only test:
1. requesting access (without authentication)
2. getting the balance

The "dApp" under test is a blank web page that does nothing. The test driver uses the `executeAsyncScript` method to execute code to call the connector APIs inside the context of the dApp as if it is the dApp calling the injected connector APIs.


Nebyt (approved):

Nebyt (commented):
/check

Nebyt (commented):
/check

Nebyt (commented):
/check

PR #2803, opened 2022-03-31
Denis/yoext 94

Nebyt:
Moved the mock dApp page methods to the separate file.

### DIFF
--- a/packages/yoroi-extension/features/connector-cardano.feature
+++ b/packages/yoroi-extension/features/connector-cardano.feature
@@ -10,7 +10,7 @@ Feature: dApp connector data signing
     Then I should see the connector popup
     And I select the only wallet named shelley-simple-15 with 5.5 balance
     Then The popup window should be closed
-    And The access request should succeed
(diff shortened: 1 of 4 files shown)


packages/yoroi-extension/features/connector-cardano.feature
```diff
@@ -10,7 +10,7 @@ Feature: dApp connector data signing
     Then I should see the connector popup
     And I select the only wallet named shelley-simple-15 with 5.5 balance
     Then The popup window should be closed
-    And The access request should succeed
+    And The access request should be succeed
```
yushih: Just nitpicking: "succeed" is a verb so it should be "succeed" or "be successful".

yushih (approved):

packages/yoroi-extension/features/connector-cardano.feature
Nebyt: Got it. I will change it back.

Nebyt (commented):
/check

PR #2791, opened 2022-03-28
Wallet send page revamp part 2

neuodev:
This is part include 
- Add NFT dialog 
- Add token dialog 
- Display tokens that are included in the transaction 

# Note
This part includes only the UI elements. No functionality yet! 

# Screenshots 📺 
Screen Shot 2022-03-28 at 10 39 23 AM

Screen Shot 2022-03-28 at 10 39 42 AM

Screen Shot 2022-03-28 at 10 39 31 AM

Screen Shot 2022-03-28 at 10 39 04 AM

PR #2812, opened 2022-04-04
Read receiver address from QR code using webcam 

neuodev:
This PR should be reviewed and merged after this one https://github.com/Emurgo/yoroi-frontend/pull/2791


package.json
```diff
@@ -16,5 +16,6 @@
   "devDependencies": {
     "husky": "4.3.8",
     "lerna": "^4.0.0"
-  }
+  },
+  "dependencies": {}
```
yushih: @AhmedIbrahim336 is this necessary?

neuodev: It happened because I installed a package in the wrong directory! I just removed it

packages/yoroi-extension/app/components/wallet/send/WalletSendFormRevamp.js
```diff
@@ -513,21 +513,38 @@ export default class WalletSendForm extends Component<Props, State> {
...
+              {this.renderCurrentStep(currentStep)}
+            </div>
           </div>
         </div>
-      </div>
+      </>
     );
   }
 
+  renderDialog(): Node {
```
yushih, line 534: For consistency  I think it's better to move rendering of the `QRScannerDialog` into `WalletSendPage.renderDialog`, the same way as `AddNFTDialog` and `AddTokenDialog` are rendered.

Also, as an application of the law of demeter of software design, it's better not to pass generic methods like `openDialog`, `isOpen` and `closeDialog` to this component. Instead, pass callback like `openAddNFTDialog` and `openQRScannerDialog`.

yushih (approved):

yushih:
@AhmedIbrahim336 fix flow, please. Or fix flow in the last PR.

packages/yoroi-extension/app/components/wallet/send/WalletSendFormRevamp.js
neuodev: The reason why I added the `QRScannerDialog` inside `WalletSendPage` is that I need access to the `receiverField` so I can update its state and display the address

yushih: ok

PR #2817, opened 2022-04-11
connector resyncs before submitting tx

yushih:
This is a temporary measure but it is needed for the current implementation. I will refactor the whole submitted transaction stuff to eliminate the requirement for resyncing when submitting tx.


neuodev (approved):
Why are we disabling DB resync?

yushih:
> Why are we disabling DB resync?

It takes time. So we should do DB sync only when necessary.

PR #2832, opened 2022-04-25
Enable multi-asset sending in the revamp

neuodev:
This PR should be reviewed and merged after this one https://github.com/Emurgo/yoroi-frontend/pull/2791

Also if possible after this one as well https://github.com/Emurgo/yoroi-frontend/pull/2812


yushih:
Issue 1:
When I input an amount that exceeds the available amount, the error message should be specific.
max

yushih:
Issue 2: 
When I choose a second token, "calculating fee..." is shown before I start to input amount.
calc

Also, consider re-calculate the min amount only after the token & nft dialog closes and the chosen tokens/nfts actually change.

yushih:
Issue 3:
Please check with Dima. The max button is not highlighted after being clicked.
max-btn

yushih:
Issue 4:
See the screenshot: the minimum required amount is not 1 ADA because I've chosen a token. But the error message shows 1 ADA.
1ada

yushih:
Issue 5:
The "minimum required" error message is sometimes shown above the form (in issue 4), sometimes below:
min-ada-2

yushih:
Issue 6:
The "Calculating fee..." message is also shown in different places:
calc2

yushih:
Issue 7:
Why the min ADA amount is sometimes shown below the amount input but not as an error message? For consistency can we always show the min amount as an error message below the form *only when* current amount < min amount? Please consult with Sergio and Dima.

min2

neuodev:
> Issue 1: When I input an amount that exceeds the available amount, the error message should be specific. max

We already have the same generic error in the old send page as well 
Screen Shot 2022-04-29 at 2 21 20 PM

yushih:
Issues 1 - 6 are fixed. 
Issue 7 remains: as soon as the focus is on the input, the "min ADA" row disappears.

yushih:
Issue 8:
there are 10000 wBTC tokens
enough1
but when I want to send 1, it says not enough
enough2

yushih:
Issue 9:
Sometimes min ADA is 0 but it shouldn't be.
minada

yushih:
Issue 10: 
Sometimes I can add more tokens than there is.
There is only 1 of this token:
overadd1
But I can add 2 without any error:
overadd2

yushih:
Possible issue 11:
I think during test, I once see that there is no token chosen, like this:
inconsis2
But when I open the dialog, there are tokens chosen.

I couldn't repeat this, maybe I was wrong, but please double check if this could happen.

neuodev:
> Issues 1 - 6 are fixed. Issue 7 remains: as soon as the focus is on the input, the "min ADA" row disappears.

As I discussed with Dima that we should display it when the user didn't enter any ADA but if the user entered any ADA there is no point in showing it!

neuodev:
> Issue 8: there are 10000 wBTC tokens enough1 but when I want to send 1, it says not enough enough2

Weird! I am able to enter the amount normal! 
Screen Shot 2022-05-13 at 2 35 14 PM

Screen Shot 2022-05-13 at 2 37 40 PM

neuodev:
> Issue 9: Sometimes min ADA is 0 but it shouldn't be. minada

I tried it and I can see the min ADA. How can I reproduce it?

neuodev:
> Issue 10: Sometimes I can add more tokens than there is. There is only 1 of this token: overadd1 But I can add 2 without any error: overadd2

How can I reproduce it? 
Screen Shot 2022-05-13 at 2 41 12 PM

neuodev:
> Possible issue 11: I think during test, I once see that there is no token chosen, like this: inconsis2 But when I open the dialog, there are tokens chosen.
> 
> I couldn't repeat this, maybe I was wrong, but please double check if this could happen.

This might happen because you didn't enter any amount so it got removed from the tokens list

neuodev:
Also, I discussed with Dima that if there is any error or we don't have any value for Min-ADA then should display it as **0.0**

neuodev:
> But when I open the dialog, there are tokens chosen.
I couldn't repeat this, maybe I was wrong, but please double check if this could happen.

I updated it. Should not happen again!

zuzunker (approved):

PR #2879, opened 2022-06-08
remove max-width of wallet name box

yushih:
The hardcoded max-width causes undesirable line wrap:
lb1
After removing:
lb2


zuzunker (approved):
Thank you, @yushih !

PR #2883, opened 2022-06-09
remove fiat conversion on pending tx amount

yushih:
Should've been part of https://github.com/Emurgo/yoroi-frontend/pull/2877 but I overlooked it. I guess we don't need to release a new update just for this because the pending tx amounts appear relatively rarely.


Nebyt:
Fixed :+1:

Nebyt:
Checked again) :+1:  Fixed

PR #2896, opened 2022-06-24
Merge ledger handling code

yushih:
Many changed files! But actually not a very complicated PR.
- the original https://github.com/Emurgo/yoroi-extension-ledger-connect-vnext source dir is copied to `/ledger`.
- some build script and webpack configs are changed so that the code in `/ledger` is compiled into a new page `ledger.html`
- when interacting with Ledger, instead of popping up `https://emurgo.github.io/yoroi-extension-ledger-connect-vnext/ledgerjs5/#/v5`, now Yoroi pops up `chrome-extension://<extension ID>/ledger.html`.
- instead of using the library `@emurgo/ledger-connect-handler` to communicate with the above page, the updated `LedgerConnect` class in `app/utils/hwConnectHandler.js` does the job.
- minimum changes are made to Yoroi extension itself and the original ledger-connect (now in `/ledger`) to adapt to the change.

After this PR is released:
- https://github.com/Emurgo/yoroi-extension-ledger-connect-handler can be retired
- https://github.com/Emurgo/yoroi-extension-ledger-connect and https://github.com/Emurgo/yoroi-extension-ledger-connect-vnext needs to be kept around for a while, because users of older Yoroi version still needs the gh-pages web site of these repos.


Nebyt:
@vsubhuman The PR is checked. It works in general only with several UI issues. Please check Jira.

zuzunker (approved):

PR #2841, opened 2022-05-02
[4.13.100] Optimize initialization speed

yushih:
Please review commit https://github.com/Emurgo/yoroi-frontend/commit/c5d2601ca02ff80565a78dc95284c549c74351c0 individual if it helps.

 [ee7d093](https://github.com/Emurgo/yoroi-frontend/pull/2841/commits/ee7d093a2072ad7d731883b891978881b5d2f1a9) makes perceived loading speed faster by loading wallet data after the main UI starts to render.

  https://github.com/Emurgo/yoroi-frontend/pull/2841/commits/80f85766a1a019b28ec1dd789c914d76b1e590d8 loads SVGs on-demand (some of them are huge but rarely used). But because of it, all the SVG import statements need to be rewritten: c5d2601ca02ff80565a78dc95284c549c74351c0 is just mechanical changing of all import syntax for SVGs. Also, flow type of SVG imports must be fixed by bd83cd2e36726bc3637c8273629e42e37b4a9b2c.

Other commits should be straightforward by the commit messages.

PR #2897, opened 2022-06-27
bug fix: language change is not immediately effective

yushih:
Language change is not immediately effective. It requires refreshing the page.

The bug is introduced by https://github.com/Emurgo/yoroi-frontend/pull/2841.


zuzunker (approved):
Thank you, @yushih !

PR #2898, opened 2022-06-27
Bug fix: balance shown in "all wallets" dialog is always loading

yushih:
The bug is because the balances of wallets shown on the "all wallets" dialog include rewards, but currently, rewards are not stored locally, nor refreshed until the wallet is open. The fix is to initiate a delegation data request at the start.

Also fixed a bug where there are both TADA and ADA wallets.


Nebyt:
Checked. :+1:

zuzunker (approved):

PR #2906, opened 2022-07-04
Max Sendable ADA

neuodev:
## Background 
Implementation of the Max sendable ADA that users can send. Based on [Max sendable ADA (Locked assets deposit)](https://emurgo.atlassian.net/wiki/spaces/CARDANO/pages/729940008/Max+sendable+ADA+Locked+assets+deposit)

## Behaviour
1. If the wallet has enough ADA the max amount will be calculated and then fill the form with it. The user should be able to edit the form.

2. If the user has not enough ADA the form will be filled with `0.0` and the user will see that the `Minimum required is 1 ADA` error.

3. the `MAX` button is not shown for ergo wallets. 

## Todo 
- [x] Proper error message 
- [x] Flow


zuzunker (approved):

zuzunker (commented):
/check

zuzunker (approved):

Nebyt:
Checked. All found issues are fixed.

zuzunker (commented):
/check

PR #2918, opened 2022-07-19
improve typing for NFT metadata

yushih:
NFT metadata is defined by https://cips.cardano.org/cips/cip25/ . The lifecycle of the metadata is:

minting tx -> blockchain -> db-sync -> Yoroi backend db -> /multiAsset/metadata endpoint -> Yoroi extension storage -> Yoroi extension UI 

Currently, there are some problems with the handling of NFT metadata:
1. The type definition in Yoroi-frontend is not strong enough, `any` types are used.
2. The `image` and `description` fields are assumed to be strings, but according to the spec they can be arrays of strings.
3. Most importantly, the values of the metadata are not checked. They are just cast from `any` to `CardanoAssetMintMetadata`, which flow is happy to accept but without guaranteeing the actual value conforms to the type.

This PR addresses these problems but correctly strong-typing the NFT metadata and adding run-time checks.

### DIFF
--- a/packages/yoroi-extension/app/api/ada/lib/storage/database/primitives/tables.js
+++ b/packages/yoroi-extension/app/api/ada/lib/storage/database/primitives/tables.js
@@ -470,6 +470,18 @@ export const AddressMappingSchema: {|
   }
 };
 
+export type NFTMetadata = {|
+  name: string,
+  image: string | Array<string>,
+  mediaType?: ?string,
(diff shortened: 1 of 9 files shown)


zuzunker (commented):
/check

zuzunker (commented):
/check

zuzunker (approved):

PR #2920, opened 2022-07-20
paginate request to mint metadata

yushih:
The backend has a request size limit:
https://github.com/Emurgo/yoroi-graphql-migration-backend/blob/master/src/services/multiAssetTxMint.ts#L62
So the front-end must paginate the request


zuzunker (approved):
Thank you, @yushih !

zuzunker (commented):
/check

zuzunker (commented):
/check

PR #2942, opened 2022-07-28
Sidebar Footer Flex

leebickmtu:
Combine FAQ and Toggle buttons into wrapping flexbox. Makes for cleaner look.

Results in "FAQ" and sidebar toggle button being aligned in a single row when sidebar is expanded. And wraps them so they are stacked when sidebar is collapsed.


neuodev:
Screen Shot 2022-08-08 at 5 01 17 PM

@leebickmtu  The sidebar is too wide and the elements are not aligned! We might need to think of something else.

zuzunker:
@leebickmtu , seems like when the sidebar us collapsed, the elements also get alined into a single row, which does not look that good. If you can change it to align **only** in the expanded state - it would be great and we'd gladly accept the change

leebickmtu:
@vsubhuman , sorry about that. Give it a try now.

zuzunker (approved):
Looks perfect! Thank you, @leebickmtu 👍

PR #2947, opened 2022-08-03
fix fund message

pedromtcosta:
IOHK now has a `fund_name` in the response at https://servicing-station.vit.iohk.io/api/v0/fund?id=1.
We should display this as the fund name instead of the ID.


zuzunker (approved):
Thank you, @pedromtcosta !

zuzunker (commented):
/check

PR #2955, opened 2022-08-09
Loading Zendesk in an Iframe

neuodev:
The service should work the same way as before but now it is isolated in an Iframe for more security control and migration requirements to manifest V3. 
The actually Zendesk integration is no longer in Yoroi it is happening [here](https://github.com/Emurgo/yoroi-support). 
for more info check the [README.md](https://github.com/Emurgo/yoroi-support/blob/main/README.md)


yushih:
Good work @AhmedIbrahim336 ! I have merged this into https://github.com/Emurgo/yoroi-frontend/pull/2965 and it works under manifest v3.

neuodev:
Nice! thanks @yushih

PR #2957, opened 2022-08-10
Not allowing tx-building when token amount is zero

zuzunker:


zuzunker (commented):
/check

zuzunker (commented):
/check

packages/yoroi-extension/app/stores/toplevel/TransactionBuilderStore.js
```diff
@@ -365,10 +365,17 @@ export default class TransactionBuilderStore extends Store<StoresMap, ActionsMap
         time: this.stores.serverConnectionStore.serverTime ?? new Date(),
       }).slot);
 
+      const tokens = this._genTokenList();
+      const hasZeroTokenAmount = tokens.some(({ amount }) =>
+        amount != null && new BigNumber(amount).isLessThanOrEqualTo(0));
+      if (hasZeroTokenAmount) {
+        return;
+      }
+
```
neuodev: I think this logic will better be added to the `_canCompute` function which is be used [here](https://github.com/Emurgo/yoroi-frontend/blob/e8e70b1b5cba01581862d85e5915e0569923020c/packages/yoroi-extension/app/stores/toplevel/TransactionBuilderStore.js#L352)
and declared [here](https://github.com/Emurgo/yoroi-frontend/blob/e8e70b1b5cba01581862d85e5915e0569923020c/packages/yoroi-extension/app/stores/toplevel/TransactionBuilderStore.js#L269). 
It is already doing the same thing. but need to include the new check

Currently, it is doing this 
```js
  _canCompute(): boolean {
    if (this.plannedTxInfoMap.length === 0) return false;
    for (const token of this.plannedTxInfoMap) {
      // we only care about the value in non-send all case
      if (
        !token.shouldSendAll && !token.amount
      ) {
        return false;
      }
      if (this.receiver == null) {
        return false;
      }
    }

    return true;
  }

```
We can combine these functions into  
```js 
  _canCompute(): boolean {
    if (this.plannedTxInfoMap.length === 0 || this.receiver === null ) return false;
    return this.plannedTxInfoMap.some((
      { amount, shouldSendAll }) => shouldSendAll || ( amount != null && new BigNumber(amount).isLessThanOrEqualTo(0) === false 
    ))
``` 

Let me know If you want me to push the fix above!

zuzunker: @AhmedIbrahim336 , good point! Thank you

zuzunker (commented):
/check

neuodev (approved):

PR #2958, opened 2022-08-11
Pat/staking center fixes

nistadev:
**Before:**
Message is not closed
[UNDELEGATE] button is disabled

image

**Now:**
Message is closed
[UNDELEGATE] button is hidden

image


neuodev:
@paulclindo  there is still some issue with the fix please check the comment on the [task](https://emurgo.atlassian.net/browse/YOEXT-55?focusedCommentId=12471)

packages/yoroi-extension/app/components/wallet/staking/dashboard-revamp/StakingTabs.js
```diff
@@ -160,16 +160,19 @@ const StyledTab = styled(Tab)({
...
-    <StyledBox>
-      <InfoIconSVG />
-      <Typography variant="body2" color="var(--yoroi-palette-gray-600)" marginLeft="8px">
-        {message}
-      </Typography>
-      <IconButton>
-        <CloseIcon />
-      </IconButton>
-    </StyledBox>
+    show && (
```
neuodev, line 165: We can't rely on the **local state** to manage this behavior as if the user navigated to the save page or switched to another tab and then got back to the same tab he will still be able to see it. The user should see it once for this we will use `localStorage` to permanently store the state.

nistadev: Got it! Since this is a temporary message and in the new revamp version will not be displayed, it is ok to do it just with the local state so we don't save unnecessary state to the `localStorage`?

zuzunker: Thank you for the catch, @AhmedIbrahim336 ! 

@nistadev , it's ok to leave it as this for now, we'll create a separate task to rework if needed later, after the final version of the design is created.

zuzunker (approved):

zuzunker (commented):
/check

PR #2963, opened 2022-08-12
Replace overlay value for overflow css prop

nistadev:
**Context**
This fixes a bug where firefox users are not able to send assets due to layout shift and not able to scroll down as seen in this video:
https://user-images.githubusercontent.com/22194171/184313659-9704a878-bc7e-4cbd-a9ce-7fe28fe749f9.mp4

The cause was this css prop `overflow: overlay` which is only compatible with [76.36%](https://caniuse.com/css-overflow-overlay) of browsers and is a deprecated prop. Refer to the table above from MDN too:
image

The property `overflow: auto` works aswell but the difference is that takes a little space on the layout for the scroll. Until `overlay` is supported, we still have to use `auto` as there is no fallback for this.

> The overlay value of the overflow CSS property is a non-standard value to make scrollbars appear on top of content rather than take up space. This value is deprecated and related functionality being standardized as [the scrollbar-gutter property](https://caniuse.com/mdn-css_properties_scrollbar-gutter).

I replaced all occurrences of the value to avoid other layout bugs caused by this incompatible prop.


neuodev:
Nice! That's great 💯  @nistadev

neuodev (approved):
LGTM 👍

zuzunker (approved):
Thank you, @nistadev !

zuzunker (commented):
/check

PR #2931, opened 2022-07-25
remove mutex on connector handler

ozgrakkurt:

PR #2968, opened 2022-08-17
Mutex removing cherrypick

zuzunker:
Cherrypick of the previous same change from https://github.com/Emurgo/yoroi-frontend/pull/2931

The original change got rolled back in https://github.com/Emurgo/yoroi-frontend/pull/2961


yushih:
I guess we can close this and leave the mutex there for the time being.

PR #2989, opened 2022-09-05
persist catalyst round info for better UX

yushih:


zuzunker (approved):
Nice! Thank you, @yushih !

PR #2997, opened 2022-09-21
fix hints page for ledger multi-asset sending

yushih:
This PR is made only now because the illustrations haven't been available until recently.
This fixes the hint page for Ledger wallet multi-asset sending, which is part of the revamp.


zuzunker (approved):
Thank you, @yushih !

PR #3002, opened 2022-09-26
Apply scoped modern theme classes to the revamp

neuodev:
## Background 
Because originally we had two themes (Classic, and Modern). We have some styles  scoped to only one of them but not for the other. For example, the classic might has a background image that is different from the modern. This was controlled in the `SASS`  files using the `:global` selector.

```sass
:global(.YoroiModern) { /* do something */ }
:global(.YoroiClassic) { /* do something else * / }
```

Adding the styles to be like this **makes the `Revamp` has no styles in some places**. Because the `Revamp` is similar to the `Modern` more than the `Classic` I genalized the styles of the `Modern` theme to be applied also on the `Revamp`. 

```sass
:global(.YoroiModern), :global(.YoroiRevamp) { /* do something */ }
:global(.YoroiClassic) { /* do something else * / }
```

Generally, this is good to have. The problem tho that this might break the layout somewhere in the extension. So it is better that @Nebyt  review it while he is doing some **full regression testing (for a new release or something)** and make sure the layout does look as we expected. The reason we need to do this as we have these scoped styles **across the whole extension (42 files)**.


zuzunker (approved):
Thank you, @AhmedIbrahim336 !

PR #3014, opened 2022-10-13
Fix `get_collateral_utxos` response for empty wallets

neuodev:
Background [here](https://emurgo.atlassian.net/browse/YOEXT-452)


zuzunker (approved):

zuzunker (approved):

zuzunker:
This is actually an incorrect solution to the problem described in the linked ticket. The problem in the ticket happens only within the example dapp because it's incorrectly handling the connector response. The connector response of an error is correct in a case of an empty wallet, as stated in the CIP30 spec:

> This shall return a list of one or more UTXOs (unspent transaction outputs) controlled by the wallet that are required to reach AT LEAST the combined ADA value target specified in amount AND the best suitable to be used as collateral inputs for transactions with plutus script inputs (pure ADA-only utxos). If this cannot be attained, an error message with an explanation of the blocking problem shall be returned. NOTE: wallets are free to return utxos that add up to a greater total ADA value than requested in the amount parameter, but wallets must never return any result where utxos would sum up to a smaller total ADA value, instead in a case like that an error message must be returned.

This change is not needed

zuzunker:
Replaced with https://github.com/Emurgo/yoroi-frontend/pull/3047

PR #3019, opened 2022-10-18
Reduce code duplication in the `connectorStore`

neuodev:
Jira: [Connector Store refractory](https://emurgo.atlassian.net/browse/YOEXT-67)


packages/yoroi-extension/app/ergo-connector/stores/ConnectorStore.js
```diff
@@ -79,60 +80,45 @@ import type { IGetAllUtxosResponse } from '../../api/ada/lib/storage/models/Publ
...
-          }
-
-          resolve(response);
-          initedConnecting = true;
+    window.chrome.runtime.sendMessage(
+      message,
+      response => {
+        if (window.chrome.runtime.lastError) {
+          // eslint-disable-next-line prefer-promise-reject-errors
+          reject(`Could not establish connection: ${JSON.stringify(typeof message === 'object' ? message : '')}`);
```
yushih: should `return` after `reject` because only either one of `reject` or `resolve` should be called.

packages/yoroi-extension/app/stores/DappConnectorStore.js
```diff
@@ -12,29 +12,19 @@ import type {
...
-            // eslint-disable-next-line prefer-promise-reject-errors
-            reject('Could not establish connection: connect_retrieve_data ');
-          }
 
-          resolve(response);
-          initedConnecting = true;
-        }
-      );
-  });
+async function sendMsgConnect(): Promise<?ConnectingMessage> {
```
yushih: The `_getConnectingMsg` function in this file is not used at all and should be removed. And since it is the only caller of `sendMsgConnect`,  `sendMsgConnect` and `initedConnecting` should be removed too.

yushih (changes requested):
`DappConnectorStore` should be in `stores/toplevel` like the others.

packages/yoroi-extension/app/stores/DappConnectorStore.js
neuodev: Done

neuodev:
> `DappConnectorStore` should be in `stores/toplevel` like the others.

Nice catch! I moved it

zuzunker (approved):

zuzunker (commented):
/check

zuzunker (commented):
/check

zuzunker (commented):
/check

PR #3046, opened 2022-11-01
Denis/support test improvement

Nebyt:
Improved working with emails


Nebyt (commented):
/regression-check

Nebyt (commented):
/regression-check

PR #3049, opened 2022-11-02
remove duplicated functions and fix typos

nistadev:
**Description**
Remove duplicated functions and fix typos to run cardano example.

PR #3050, opened 2022-11-02
Fix Cardano example crashes on start

neuodev:


neuodev:
Already fixed in #3049

PR #3056, opened 2022-11-16
Clean up the connector

yushih:
Some dead code are removed:
1. The signing request type `tx-create-req/cardano` and its handler function is removed. This request type was there when the `createTx` connector API signs the created tx.

2. The `getResponse` and `getSigningMsg` actions are removed. They are not used at all.

3. The `getConnectorWhitelist` action is removed from `ConnectorStore` and `ConnectorContainer`. It's unnecessary for the `ConnectorContainer` to trigger the action in `ConnectorStore` because the latter calls `this._getConnectorWhitelist()` in `setup()`. The action is only used by `DappConnectorStore` and `ConnectedWebsitesContainer`.


neuodev (approved):

zuzunker (approved):
Thank you, @yushih !

Issue #3057, opened 2022-11-27
Not all NFT images are shown in the gallery


philipphenkel:
While most NFT images are shown, some are not. See attached image.

I assume the frontend does not handle array-based image URLs well. The console log indicates that at least.

The following URL array is valid and recommend for CIDv1. The frontend seems to pick only the first element from the array and thus image loading fails.

```
         image: [
            "ipfs://",
            "bafybeidreetdv5a4b7sc3iufvadrolhylkezlfcfkqaicvydicrlxmbwn4"
         ],

```

[asset1ul2l6u869u4jcxpjrc0mdvcfkgejw6c0f69ldv](https://cardanoscan.io/token/e7d5fd70fa2f2b2c18321e1fb6b309b233276b0f) is an example of an unsupported NFT and works properly on [pool.pm](https://pool.pm/asset1ul2l6u869u4jcxpjrc0mdvcfkgejw6c0f69ldv)

![Screenshot 2022-11-27 12 38 51](https://user-images.githubusercontent.com/828416/204133371-12fdd6b4-ad30-4eb4-b724-e64c5159b7d6.png)

zuzunker:
This is fixed in 4.18, which is in production now

PR #3058, opened 2022-11-27
Support array-based IPFS image URLs for NFTs

philipphenkel:
This PR enables support for array-based IPFS image URLs in NFT metadata, e.g.

```
     image: [
            "ipfs://",
            "bafybeidreetdv5a4b7sc3iufvadrolhylkezlfcfkqaicvydicrlxmbwn4"
         ],
```
Currently, only the first string of an array is taken into account. This results in an invalid image URL for the above case.

This PR proposes to merge all strings in the image array and use that as the URL.

Fix for #3057


zuzunker (approved):
Thank you, @philipphenkel !

PR #3091, opened 2023-01-06
Connector mark foreign outputs

zuzunker:


packages/yoroi-extension/app/connector/stores/ConnectorStore.js
```diff
@@ -611,13 +611,21 @@ export default class ConnectorStore extends Store<StoresMap, ActionsMap> {
...
+    ])
+
     const outputs = [];
     for (let i = 0; i < txBody.outputs().len(); i++) {
       const output = txBody.outputs().get(i);
       const address = Buffer.from(output.address().to_bytes()).toString('hex');
       outputs.push(
         {
           address,
+          isForeign: !ownAddresses.has(address),
```
nistadev, line 628: Would need that on inputs if possible aswell! Overall looks great 👏🏽 🙌🏽

nistadev (approved):

packages/yoroi-extension/app/connector/stores/ConnectorStore.js
zuzunker: @nistadev, no need for this to be in the inputs, since there are two separate fields: `inputs` and `foreignInputs` )

packages/yoroi-extension/app/connector/stores/ConnectorStore.js
```diff
@@ -631,17 +639,10 @@ export default class ConnectorStore extends Store<StoresMap, ActionsMap> {
...
-      ...response.usedAddresses,
-      ...response.unusedAddresses,
-      response.changeAddress
-    ])
-
     const { amount, total } = await this._calculateAmountAndTotal(
       selectedWallet.publicDeriver,
       inputs,
-      outputs,
+      outputs.map(({ address, value }) => ({ address, value })),
```
neuodev: `_calculateAmountAndTotal` accepts `outputs` with the type below. So I needed to remove the `isForeign` flag so flow does not complain 

```ts
outputs : Array<{ address: string, value: MultiToken}>
```

neuodev (commented):
Perfect! Thanks, @vsubhuman

I pushed a new commit to fix the flow errors. 

I have one note about the `package-lock.json`. I can't see any new package installed! Can you please revert the changes in this file? Thanks in advance!

packages/yoroi-extension/app/connector/stores/ConnectorStore.js
zuzunker: Thank you, @AhmedIbrahim336 ! I have updated the code now a little, cuz we don't need to do the actual mapping (calculations) just to match the flow types. We can just accept an inexact type in the function if we mark it as read-only, see the change here: https://github.com/Emurgo/yoroi-frontend/pull/3091/commits/198900e27f37cd3917c2e0f3965a1a425f6e3b9e#diff-4fdbc4bd4e71aa6ffc075cb2acf855b5870659d9f1d40e25e1ce6b2cf76af58eR857

zuzunker (commented):
/check

PR #3099, opened 2023-01-24
Restructure code splitting — remove the quick flash white screen when switching pages 

neuodev:
## Background 

We implemented lazy loading (aka code splitting) back then when Yoroi was taking so long in the initial loading screen when opening the extension. 

## The Problem 

When switching between tabs/pages we have a quick white flash screen.

## Root Cause 

It happens because when switching to a new tab that you never visited before the **page is not actually there** what will happen is while loading this page we will show this blank white screen. This screen is shown because **we only have one `Suspense` with a fallback of `null` that wrapped the whole application ([here](https://github.com/Emurgo/yoroi-frontend/blob/develop/packages/yoroi-extension/app/Routes.js#L179))**

```tsx 
<Suspense fallback={null}>
  {/* all pages */}
</Suspense>
```

So whenever **page or subpage** is still loading we **fallback** to the blank white screen. 

## The Solution 

1. Instead of having one `Suspense` that wraps all pages we scope it into the **level of subpages** 
2. Don't lazy load some pages like `LoadingScreen`
3. For full pages we only move the `Suspense` to the page content not the whole page


packages/yoroi-extension/app/Routes.js
```diff
@@ -448,7 +427,9 @@ export function wrapAssets(
     <AssetsWrapper
       {...assetsProps}
     >
-      {children}
+      <Suspense fallback={null}>
```
nistadev, line 436: Maybe we can use some sort of loading placeholder? 

We have the .gif/svg of the logo, we can check it with the designers team.

nistadev (approved):

packages/yoroi-extension/app/Routes.js
neuodev: I thought of it but found out that the **blank white screen** doesn't actually take so long (a fraction of the second on my machine) so adding a loading page in the `fallback` might cause the user to wonder what was happing. Actually, you can try to do it yourself and you will not notice it is not that good! Try to put the `LoadingScreen` as the `fallback` for the `Suspense` and see if it looks good!

nistadev: Yeah, sounds good. Actually I was thinking of users with slow connection on 3g or other poor infrastructure. I think for those it will be better as they will know what's happening instead of a white screen. We won't have that problem because our connection is fast. Wdyt?

Edit: A skeleton would be better than a loader, but since we have to change some designs, maybe we can consider adding it in the future.

neuodev: Nice idea about the skeleton, I think it will make sense! we can add it in the future!

neuodev:
## Quick Note: 

The big diff in the PR `+1,204 −1,091` is from splitting both the `VotingPage.js` and `StakingPage.js` into two other pages. 
For example the `VotingPage.js` -> `VotingPage.js` to wrap the content and add the navbar and the sidebar and `VotingPageContent.js` includes the actual content/logic of the page 

And the same applies to the `StakingPage.js`! 

So the diff comes from just moving the stuff around but on actual adding or deletion was done!

Nebyt (approved):

zuzunker (commented):
/check

PR #3104, opened 2023-01-31
Pat/missing items tx tab

nistadev:
More info: [YOEXT-328](https://emurgo.atlassian.net/browse/YOEXT-328)

**Screenshots**
image
image
image

[YOEXT-328]: https://emurgo.atlassian.net/browse/YOEXT-328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/yoroi-extension/app/components/wallet/transactions/TransactionRevamp.js
```diff
@@ -916,3 +963,28 @@ export default class TransactionRevamp extends Component<Props, State> {
...
+      break;
+    }
+    case 'error': {
+      Icon = ErrorIcon;
+      break;
+    }
+  }
+
+  return <Box sx={{ width: 40, height: 40 }}>{Icon && <Icon />}</Box>;
+};
```
neuodev:
```suggestion
const icons = {
    send: SendIcon,
    receive: ReceiveIcon,
    reward: RewardIcon,
    error: ErrorIcon
};
const TypeIcon = ({ type = 'send' }) => {
  let Icon = icons[type];
  return <Box sx={{ width: 40, height: 40 }}>{Icon && <Icon />}</Box>;
};
```

packages/yoroi-extension/app/components/wallet/transactions/TransactionRevamp.js
```diff
@@ -97,6 +103,62 @@ export default class TransactionRevamp extends Component<Props, State> {
     this.setState(prevState => ({ isExpanded: !prevState.isExpanded }));
   };
 
+  getTxType(data: WalletTransaction): string | null {
```
neuodev: I noticed that the `getTxType` and the `getTxTypeMsg` have quite similar logic! they are almost the same :) 
and they are used in the same place. I suggest updating `getTxTypeMsg` to return both **messages the status**. So the end result should be one function called `getTxType` that returns both.

```ts
function getTxType(intl: $npm$ReactIntl$IntlFormat, currency: string, data: WalletTransaction): [string, string | null] 
```

neuodev (changes requested):
Nice Work! the new look is great! 
the code looks great, see my comments!
I have one note about code formatting. The PR has some changes that are just all about code formatting. Are you doing it on purpose?

packages/yoroi-extension/app/components/wallet/transactions/TransactionRevamp.js
nistadev: Checking this, thanks for the feedback! :)

neuodev (approved):
@nistadev Greate Work!! LGTM 🚀

Nebyt (approved):

PR #3107, opened 2023-02-06
use new CSL type to safely handle tx body when connector signs the tx

yushih:
1. When signing the tx, extract the tx body using the `FixedTransaction` type of CSL instead of the `cbor` library.
2. The `signTx` API may return only the witness or the full tx. The PR makes sure that in the latter case, the input tx body is preserved.


packages/yoroi-extension/chrome/extension/background.js
```diff
@@ -990,15 +990,26 @@ function handleInjectorConnect(port) {
...
+                try {
+                  fullTx = RustModule.WalletV4.FixedTransaction.new(
+                    bodyOrTxBytes,
+                    resp.ok.to_bytes(),
+                    true,
+                  );
+                } catch {
+                  fullTx = RustModule.WalletV4.FixedTransaction.from_bytes(bodyOrTxBytes);
+                }
+                fullTx.set_witness_set(resp.ok);
```
lisicky: Be careful, `Fixed Transaction.set_witness_set` accepts serialized witness set

yushih: good catch! Thanks @lisicky!

zuzunker (approved):

zuzunker (approved):

zuzunker (commented):
/check

PR #3115, opened 2023-02-10
temporarily revert wasm scoping code that breaks connector tx signing

yushih:


yushih:
Root cause fixed by https://github.com/Emurgo/yoroi-frontend/pull/3117

PR #3119, opened 2023-02-13
Fix Staking page crash when using non-reward wallet (Not Shelly)

neuodev:
## How to reproduce? 

Switch to the revamped theme and restore a non-reward wallet (not shelly, eg: paper wallet), and then navigate to the staking page. You should see the extension crashes with the error below 

<details> 
  <summary> Screenshot</summary>

Screen Shot 2023-02-13 at 12 59 22 PM

</details> 

## Expected Fix 

- Staking page will not be shown in the first place 
- While the user is on the **staking page** and he decided to switch to another non-reward wallet, the user will be redirected to the TXs page to prevent the extension from crashing.


Nebyt (approved):

zuzunker (approved):
Thank you, @AhmedIbrahim336 !

zuzunker (commented):
/check

PR #3134, opened 2023-03-06
Add the third step "Verify Recovery Phrase" from the "Create Wallet" flow

neuodev:
<details> 
  <summary>Preview</summary>

Screen Shot 2023-03-06 at 2 32 38 PM

Screen Shot 2023-03-08 at 1 58 22 PM

</details>


neuodev:
[autoscirpt] Jira Ticket: **[Create "Verify recovery phrase" step components](https://emurgo.atlassian.net/browse/YOEXT-537)**

[](__JIRA_COMMENT_FLAG)

packages/yoroi-extension/app/components/wallet/create-wallet/VerifyRecoveryPhraseStep.js
```diff
@@ -0,0 +1,239 @@
...
+      const copy = [...prev];
+      copy[idx] = word;
+      return copy;
+    });
+    setWrongWord(null);
+  }
+
+  function isWordAdded(word) {
+    return enteredRecoveryPhrase.some(w => w === word);
+  }
```
zuzunker: @AhmedIbrahim336 , recovery phrases can and will contain multiple instances of the same word , you cannot decide only by the word itself that a similar word should not be entered again. For example, this is a valid bip39 mnemonic:
```
day day day day day day day day day day day day day day day
```

Always keep this case in mind.

zuzunker (changes requested):

packages/yoroi-extension/app/utils/recoveryPhrase.js
```diff
@@ -0,0 +1,29 @@
...
+    word: string,
+    originalIdx: number,
+|};
+
+export function makeSortedPhrase(recoveryPhrase: Array<string>): SortedRecoveryPhraseEntry[] {
+  const sorted = recoveryPhrase.slice().sort();
+  const wordIndexes = new Set();
+
+  return sorted.map((sortedWord, sortedWordIdx) => {
+    const originalIdx = recoveryPhrase.findIndex((originalWord, idx) => {
```
vsubhuman, line 14: @AhmedIbrahim336 , same thing here, for a recovery phrase `'b b a'`, you would get sorted array `['a', 'b', 'b']` and then the `originalIdx` for both `b` would be `0`. Never assume you have unique words. You can fix this by preserving the original index before sorting and then sort by a specific condition, e.g.:

```
const sorted = recoveryPhrase
  .map((word, originalIndex) => { word, originalIndex })
  .sort((a, b) => a.word.localeCompare(b.word))
```

vsubhuman (changes requested):

packages/yoroi-extension/app/utils/recoveryPhrase.js
```diff
@@ -0,0 +1,29 @@
...
+      return sortedWord === originalWord && !wordIndexes.has(idx)
+    });
+
+    if (originalIdx === -1) throw new Error('Word not found in the original recovery phrase. Should never happen');
+
+    // Mark word index as watched to handle recovery phrase with duplicates
+    wordIndexes.add(originalIdx);
```
zuzunker: @AhmedIbrahim336 , you can avoid having to check for duplicates separately if you just store original indexes before sorting and then sort by a specific condition, might make it a bit more readable, e.g.:

```
const sorted = recoveryPhrase
  .map((word, originalIndex) => ({ word, originalIndex }))
  .sort((a, b) => a.word.localeCompare(b.word))
  .map((obj, id) => ({ ...obj, id }));
```

packages/yoroi-extension/app/components/wallet/create-wallet/VerifyRecoveryPhraseStep.js
```diff
@@ -0,0 +1,239 @@
...
+          justifyContent="center"
+          gap="8px"
+        >
+          {sortedRecoveryPhrase.map(({ word, id, originalIdx }) => {
+            return (
+              <button
+                type="button"
+                key={id}
+                className={classnames(styles.wordChip, {
+                  [styles.wordAdded]: isWordAdded(word),
```
zuzunker: Cannot use just the string word here, must check by a combination like 
```
`${word}-${originalIdx}`
```

packages/yoroi-extension/app/components/wallet/create-wallet/VerifyRecoveryPhraseStep.js
```diff
@@ -0,0 +1,239 @@
...
+                  [styles.wrongWord]: wrongWord === word,
```
zuzunker: Same thing , just the string word itself is not enough, there will be duplicate words, use combined keys like `${word}-${originalIdx}` to make them unique word identifiers, or even just store the original index of the word , e.g. 

```
[styles.wrongWord]: wrongWordOriginalIdx === originalIdx,
```

zuzunker (changes requested):

packages/yoroi-extension/app/components/wallet/create-wallet/VerifyRecoveryPhraseStep.js
```diff
@@ -0,0 +1,239 @@
...
+    const isInCorrectOrder = recoveryPhrase[nextWordIdx] === word;
+    if (!isInCorrectOrder) {
+      return setWrongWord(word);
+    }
+
+    setRecoveryPhrase(prev => {
+      const copy = [...prev];
+      copy[idx] = word;
+      return copy;
+    });
```
zuzunker, line 67: @AhmedIbrahim336 , in this function you receive parameter `idx`. Later in the code `originalidx` is passed into that parameter. By the logic you are using this parameter as "the index where to add the word", because you are assigning the new recovery phrase as `copy[idx] = word;`. BUT you are using `nextWordIdx` for validation, which is calculated as the first **empty** index of the entered recovery phrase: `enteredRecoveryPhrase.findIndex(w => w === null)`. This creates a weird behaviour case, for example

Imagine you have a recovery phrase `"b a b"`, the `recoveryPhrase` array is then `['b', 'a', 'b']`, in the sorted words the user will see options: 'a', 'b', and 'b', and the user needs to enter the first word of the phrase which is `b`. From the user perspective both words `b` are equally valid, but internally they have different `originalIdx` values: one is `0` and the other is `2`.

In your implementation if the user picks `b` with the `originalIdx = 2` it will pass the validation, because the `nextWordIdx` will be equal to `0` and `recoveryPhrase[0] === 'b'` will pass, but then the `setRecoveryPhrase` will actually assign the selected word to the index `2` instead of the index `0` because the `originalIdx` passed to `onAddWord` will be `2`, so the user tries to enter the first word and will instead get result `[null, null, b]`.

This must be changed. Don't enforce setting the words to the same index as they have been at originally. Logically equal words are interchangeable. The original index must only be used as a unique word identifier, so they are not compared just by text **internally**. I'd recommend to actually abandon storing `originalidx` and instead call it like:

```
internalWordId: `${word}-${originalIdx}`
```

And then only use this field for internal checks that require uniqueness, e.g. 
```
setWrongWord(selectedWord.internalWordId);
```

zuzunker (changes requested):

packages/yoroi-extension/app/components/wallet/create-wallet/VerifyRecoveryPhraseStep.js
neuodev: Noted! I fixed it

packages/yoroi-extension/app/components/wallet/create-wallet/VerifyRecoveryPhraseStep.js
neuodev: Yup, you are correct! I updated it. I am using the the `wrongWordOriginalIdx` now

neuodev:
Great review 🙌 

Here is a quick summary of what got changed 

- I removed the `originalIndex` to avoid confusion. 
- Using the `word & idx` to identify the wrong word (prev. was only the word)

<details><summary>Recovery phrase example with similar words</summary>
<p>

## Demo on "b a b" recovery phrase. 

Similar words must be ***exchangeable***. but sill be marked as entered. 

https://user-images.githubusercontent.com/72753578/231117048-d267b38c-ad23-420f-885e-063730d789e3.mov


Same example on a recover phrase with similar words. 

`day day day day day day day day day day day day day day day`

Note: any word you click will be entered at the correct index (top part) but still marked as entered (lower part). 

https://user-images.githubusercontent.com/72753578/231117068-6fc9a210-ed5c-4990-8c2e-2ea12e43dd5d.mov



</p>
</details>

packages/yoroi-extension/app/components/wallet/create-wallet/VerifyRecoveryPhraseStep.js
neuodev: - Checkout my comment https://github.com/Emurgo/yoroi-frontend/pull/3134#issuecomment-1502994619
- I fixed the confusion between the `idx`, `originalIdx`, and `nextWordIdx`


About the 

```js
setWrongWord(selectedWord.internalWordId);
```

Because the `wrongWord` is only considered with the `enteredRecoveryPhrase`, I only add the word `idx` not the `originalIdx` or the `internalWordId`

```js
 if (!isInCorrectOrder) return setWrongWordIdx(idx);
```

zuzunker:
> Recovery phrase example with similar words

Thank you, @neuodev ! The example videos are perfect! 🙌

zuzunker (approved):

neuodev:
@vsubhuman Flow & unit tests are fixed now. All look good!

zuzunker (commented):
/check

PR #3225, opened 2023-05-30
fix(fiat-price): convert date from milliseconds into seconds

neuodev:
It seems that due to a recent backend update https://github.com/Emurgo/yoroi-backend/pull/5 the timestamp is now in seconds not in milliseconds. 

This causes the front end to think that it **has outdated prices** so it doesn't show the prices to the users. 

This PR is making sure that we are comparing the same date unit (seconds).


neuodev:
[autoscirpt] Jira Ticket: **[Currency pairing doesn't work](https://emurgo.atlassian.net/browse/YOEXT-612)**

[](__JIRA_COMMENT_FLAG)

Nebyt (approved):

yushih:
Nothing in the backend is changed. Due to DevOps re-configuration, the endpoint is temporarily down. That is all.

PR #3233, opened 2023-06-14
feat: improve design system architecture and overall styling

nistadev:
**Description**
- The architecture for the theming was improved to support multiple themes and to reuse styles for overriding components. This is the base for the future improvements regarding styles and theming across all application.
- Removed gray backgrounds and layout paddings modified according to the new revamp specs (only affects revamp)
- **BONUS FEATURE:** as I was changing the architecture for the theming part, in parallel I added a new theme we were working on with designers, dark theme. Still WIP but it is a good PoC and keeps the structure clean and simple, we will only need to be updatind the UI components to use DS tokens instead of hardcoded values or css-in-js properties. You will find a button to toggle the color mode on Settings > Version. Note: not all components are updated to use theming in a proper way, this will be done in the near future.

**Screenshots**
Overall design improvement (background, paddings)
image

(disabled) Dark mode 
image


neuodev:
[autoscirpt] Jira Ticket: **[Add style variants for revamp components & remove grey background](https://emurgo.atlassian.net/browse/YOEXT-600)**

[](__JIRA_COMMENT_FLAG)

packages/yoroi-extension/app/styles/themes/revamp/dark-theme-base.js
```diff
@@ -0,0 +1,90 @@
...
+  main: '#FF1351',
+  '700': '#CF053A',
+  '600': '#E80742',
+  '500': '#FF1351',
+  '300': '#FBCBD7',
+  '100': '#FFF1F5',
+};
+
+export const darkTheme = {
+  name: 'revamp',
```
neuodev: I think the team name should be unique. 
```suggestion
  name: 'dark-revamp',
```

packages/yoroi-extension/app/styles/themes/revamp/light-theme-base.js
```diff
@@ -0,0 +1,90 @@
...
+      '600': '#6B7384',
+      '500': '#8A92A3',
+      '400': '#A7AFC0',
+      '300': '#C4CAD7',
+      '200': '#DCE0E9',
+      '100': '#EAEDF2',
+      '50': '#F0F3F5',
+      min: '#FFFFFF',
+      contrastText: '#FFF',
+    },
```
neuodev, line 60: The common theme already has [these colors](https://github.com/Emurgo/yoroi-frontend/blob/develop/packages/yoroi-extension/app/styles/themes/common-theme.js#LL20C1-L31C7) under the name `gray: {...}`. You can use it!

packages/yoroi-extension/app/styles/themes/revamp/dark-theme-base.js
nistadev: I used the approach found in the docs, they don't specify the theme name but the mode. I guess we can remove that since is not used or just rename it like you did. (https://mui.com/material-ui/customization/dark-mode/#dark-mode-with-a-custom-palette)

packages/yoroi-extension/app/styles/themes/revamp/light-theme-base.js
nistadev: I think we should name it exaclty as defined in the DS. Designers defined "grayscale" on the same level as "primary" and "secondary", I guess because it's incorrect to say gray to black. Anyways, imo, I think we should change the previous "gray" and use "grayscale" as per definition of the designers on the DS so we have a single language and we we look at the designs, it does not cause confusion.

nistadev: Checked and seems like they are not mapped correctly, so we just use the new approach as is better for the DS.

nistadev:
Fixed on the last commit 👇🏽
- ✅ Buttons “Receive“ and “Send“ are missing 
- ✅ Border color of the wallet component is different from the design 
- ✅ The line-height of Export button should be 22 not 19 as it is now
- ✅ Width of the left side-panel should be 72px.
- ✅ The color of the left side-panel should be linear-gradient.
- ✅ The icon for the Stake delegated is added to the design.
- ✅ The width of the central part should be fixed to 1,872px



Fixed (or will be) in other PR 👇🏽
- The copy address icon should have size 24x24 not 20x20 as it is now
- The line-height of table headers should be 22 not 24 as it is now
- The line-height of timestamp should be 22 not 24 as it is now
- The height of the subMenu tabs element should be 48 not 37 as it is now

neuodev:
- [ ] https://github.com/Emurgo/yoroi-frontend/pull/3254
- [ ] #3251 
- [ ] https://github.com/Emurgo/yoroi-frontend/pull/3253
- [x] https://github.com/Emurgo/yoroi-frontend/pull/3258

zuzunker (approved):

zuzunker (commented):
/check

PR #3245, opened 2023-06-29
enable vietnamese translation

zuzunker:
Using a unicode flag temporarily, while waiting for a proper svg element


zuzunker (commented):
/check

PR #3267, opened 2023-08-24
[fix/4.22.400] changed CardanoAPI to function as a singleton with no dependency on `this`

zuzunker:
Some users are encountering issues, when using certain dApps, that `CardanoAPI` functions are being called with undefined `this`, which is most probably caused by said dApps calling the functions as higher order and not using `.bind` to do it properly. Rather than telling users to go and report it to these dApps to use instances properly, we can just make an update on our side and change the api to function as a singleton internally, since logically there is no real reason to ever have more than a single instance of it anyway. The usage of `this` is removed in the API object entirely.

### DIFF
--- a/packages/yoroi-connector/src/inject.js
+++ b/packages/yoroi-connector/src/inject.js
@@ -164,18 +164,17 @@ class CardanoAuth {
 class CardanoAPI {
   
     constructor(auth, rpc) {
-      const self = this;
       function rpcWrapper(func, params) {
-        return rpc(func, params, self._returnType[0]);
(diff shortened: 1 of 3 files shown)


neuodev:
[autoscirpt] Jira Ticket: **[[External] User is not able to unlock liquidity on dApp](https://emurgo.atlassian.net/browse/YOEXT-687)**

[](__JIRA_COMMENT_FLAG)

Nebyt (dismissed):

zuzunker:
Release: https://github.com/Emurgo/yoroi-frontend/releases/tag/4.22.400

PR #3279, opened 2023-09-07
Connector `.getCollateral` parameters fix

zuzunker:
Both `.getCollateral` and the deprecated `.getCollateralUtxos` functions can now be called with one of these parameters:
1. A number as a `number`
2. A number as a `string`
3. An object `{ amount: string | number }` (proper CIP30)
4. A CBOR encoding of type `Value`
5. No parameter (in which case the default max allowed value of 5 ADA will be used)

### DIFF
--- a/packages/yoroi-connector/src/inject.js
+++ b/packages/yoroi-connector/src/inject.js
@@ -264,11 +264,15 @@ class CardanoAPI {
 
     // DEPRECATED
     getCollateralUtxos(requiredAmount) {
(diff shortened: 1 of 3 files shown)


neuodev:
[autoscirpt] Jira Ticket: **[Broken `getCollateralUtxos`](https://emurgo.atlassian.net/browse/YOEXT-535)**

[](__JIRA_COMMENT_FLAG)

PR #3288, opened 2023-09-14
change condition for more txs

yushih:
Original the Postgresql backend always returns 20 txs for the `txs/summaries` query. Now the neo4j backend returns an indeterminate ( but always < 20) number of txs. So the condition for determined whether there is possibly more txs must change accordining.


neuodev:
[autoscirpt] Jira Ticket: **[The button "Show more transactions" is one-time-use](https://emurgo.atlassian.net/browse/YOEXT-781)**

[](__JIRA_COMMENT_FLAG)

neuodev (approved):

zuzunker (approved):
Thank you, @yushih !

PR #3301, opened 2023-10-08
fix(send): wrong error appears when sending NFT

neuodev:
This patch updates the send form when the user is sending an NFT:

1. The amount input is disabled, and the amount is sent to one.
2. The "send all" option is chosen from the dropdown menu and the menu will be disabled

Please keep in mind that if the extension does not recognize the token as NFT, none of the changes will apply. above will work.

<details><summary>Preview</summary>
<p>

Screenshot 2023-10-08 at 12 25 53 PM

</p>
</details>


neuodev:
[autoscirpt] Jira Ticket: **[Wrong error appears. Send an Nft.](https://emurgo.atlassian.net/browse/YOEXT-135)**

[](__JIRA_COMMENT_FLAG)

zuzunker (approved):
Thank you, @neuodev!

PR #3307, opened 2023-10-10
fix failure to handle tx with pool registration certificate

yushih:
Originally the old backend returned an address, which is, strictly speaking, not the correct type. Now the new back correctly returns a key hash.


neuodev:
[autoscirpt] Jira Ticket: **[Wallets containing pool registration transactions can't sync history](https://emurgo.atlassian.net/browse/YOEXT-757)**

[](__JIRA_COMMENT_FLAG)

neuodev (approved):
/test

zuzunker (approved):

PR #3343, opened 2023-11-13
fix(ds): add dialog changes

nistadev:
**Screenshots**

image

image

image


neuodev:
[autoscirpt] Jira Ticket: **[UAT. General. Modals. Difference from the design.](https://emurgo.atlassian.net/browse/YOEXT-836)**

[](__JIRA_COMMENT_FLAG)

packages/yoroi-extension/app/components/widgets/Dialog.js
```diff
@@ -126,34 +136,55 @@ DialogFn.defaultProps = {
...
+    onClick={onClose}
+  >
+    {closeButton || <CrossIcon />}
+  </Box>
+);
+
 const ModalContainer = styled(Box)(({ theme }) => ({
   position: 'relative',
-  minWidth: 'var(--yoroi-comp-dialog-min-width-md)',
+  minWidth: theme.name === 'classic' ? 'var(--yoroi-comp-dialog-min-width-md)' : '612px',
```
neuodev: This will affect the modern dialogs as well

lonewold999: In general the modern dialogs follow the same styles as revam, at least in all this file, but we can try to add another check if that's an issue.

nistadev: In general the modern theme follows the revamp styles, at least in this file we can see it all around. If that's an issue we can try do do another check.

zuzunker (approved):

PR #3349, opened 2023-11-16
fix(ds): revamp modal dialog fixes

nistadev:
### Changes
Fixed a lot of styles and added scrollable borders

image

image

_Note: Since the scrollable content can be any element inside the modal content, I added a new prop to pass the class of the possible content that can be scrollable, the modal component checks at render if it is the case or not and adds the proper state & styles to the element. Below we have an example._

Add Token modal has a different scrollable element👇🏽
image

And AddNFT modal too👇🏽
image


neuodev:
[autoscirpt] Jira Ticket: **[UAT. General. Modals. Difference from the design.](https://emurgo.atlassian.net/browse/YOEXT-836)**

[](__JIRA_COMMENT_FLAG)

Nebyt (dismissed):

Nebyt (approved):

PR #3367, opened 2023-12-12
fix(opt-out-analytics): add height, padding and overflow to analytics page

nistadev:
**Description**
If a browser window height is around 560px then the Opt-Out Analytics page is not passable because both buttons are hidden.

**Screenshots**
image


neuodev:
[autoscirpt] Jira Ticket: **[The Opt-Out analytics first screen is not scrollable](https://emurgo.atlassian.net/browse/YOEXT-936)**

[](__JIRA_COMMENT_FLAG)

Nebyt (dismissed):

zuzunker (approved):
Thank you, @nistadev !

PR #3391, opened 2024-01-19
domain address resolver

zuzunker:
https://emurgo.atlassian.net/browse/YOEXT-910


neuodev:
[autoscript] ⚠️ `ruslan/resolver` doesn't full our branch naming convention.
Should follow `<tag>/<jira-ticket-id>/<short-name>`
<details>
<summary>Details</summary>

- `<tag>`: `fix`, `feat`, `chore`, `refactor`, `test`, ...
- `<jira-ticket-id>`: The issue ID of JIRA all capital letters
- `<short-name>`: A short and descriptive name for the branch
</details>

[](__JIRA_COMMENT_FLAG)

zuzunker (commented):
/check

packages/yoroi-extension/app/stores/ada/AdaAddressesStore.js
```diff
@@ -54,8 +58,102 @@ export async function filterMangledAddresses(request: {|
...
+
+  async resolveDomainAddress(resolve: string): Promise<?DomainResolverResponse> {
+    const { getCardanoAddresses } = this._domainResolverApi ?? {};
+    if (getCardanoAddresses == null || !isResolvableDomain(resolve)) {
+      return Promise.resolve(null);
+    }
+    const res = await getCardanoAddresses({ resolve });
+    let resultSuccess: ?DomainResolverResponse = null;
+    let resultForbidden: ?DomainResolverResponse = null;
+    let resultUnexpected: ?DomainResolverResponse = null;
```
nistadev: Why don't use just a single variable for the result?

packages/yoroi-extension/app/utils/formatters.js
```diff
@@ -87,10 +87,10 @@ export function truncateStakePool(addr: string): string {
   return truncateFormatter(addr, 18);
 }
 
-export function truncateAddress(addr: string): string {
+export function truncateAddress(addr: string, cutoff: number = 40): string {
```
nistadev: `truncateAddressShort` function is the same

nistadev (dismissed):
Looks great for me!

packages/yoroi-extension/app/utils/formatters.js
zuzunker: Good catch! Switched to use `truncateAddressShort` and rolled back that change

packages/yoroi-extension/app/stores/ada/AdaAddressesStore.js
zuzunker: @nistadev , the different result types are returned in a specific order, with multiple variables for the different types the order of the return is then clearly and obviously defined in the return itself, i.e.:
```
return resultSuccess ?? resultForbidden ?? resultUnexpected;
```

This was specifically chosen for the most readability. Replacing it with a single variable would mean the priority order would need to be defined in the IF checks, like `if (result == null || result.errorType === 'unexpected')`, etc. Which means the order logic gets spread across multiple unrelated lines (separate checks in each IF) and makes it MUCH harder to understand and follow.

After your feedback I have realised that `resultSuccess` is not needed to be stored, tho, because it is always highest priority, so I have made a change to short-cut on success and return it right away. But for the two error types there are still two variables used, for the reasons of readability.

zuzunker (commented):
/check

zuzunker (commented):
/check

zuzunker (commented):
/check

zuzunker (commented):
/check

PR #3461, opened 2024-03-17
unused exports

zuzunker:
1. Added `import/no-unused-modules` eslint rule as warn
2. Added `eslint-loud` script as opposite to default `eslint` which ignores warnings
3. Disabled max-len eslint warning (experience shows we don't care about enforcing that for now)
4. Added some files to be ignored using `/* eslint-disable  import/no-unused-modules */`
5. Removed some unused declarations
6. Removed `export` on some declarations that are only used locally
7. Moved few declarations if they been imported by some single other file to be in that file


neuodev:
[autoscript] ⚠️ `ruslan/unused-exports` doesn't full our branch naming convention.
Should follow `<tag>/<jira-ticket-id>/<short-name>`
<details>
<summary>Details</summary>

- `<tag>`: `fix`, `feat`, `chore`, `refactor`, `test`, ...
- `<jira-ticket-id>`: The issue ID of JIRA all capital letters
- `<short-name>`: A short and descriptive name for the branch
</details>

[](__JIRA_COMMENT_FLAG)

Nebyt:
The changes will be checked when testing the release.

yushih (approved):

packages/yoroi-extension/.eslintignore
```diff
@@ -1,3 +1,11 @@
...
+build
+chrome/content-scripts
+config
+dev
+docs
+flow
 flow-typed
+ledger
+node_modules
+translations
```
yushih, line 11: Some of the directories added are not in the parameters of the eslint command in package.json at all, so why bother to add them here?

zuzunker: That was an overkill when I locally tried to solve some issue with using `eslint .` , but it didn't work after all and I rolled it back, but left the ignores. Doesn't break anything so decided to leave it

zuzunker (commented):
/check

PR #3509, opened 2024-05-13
Better fix for broken `nameof` by babel

yushih:
After fixing the `@babel/helpers` version, `nameof` is restored.


zuzunker (approved):
Nice!

PR #3539, opened 2024-06-11
fix(swap): add limit price change action

nistadev:
**Description**
While we had selected "Limit order" and we try to edit the price, if we press the delete button or any number on keyboard, several zeros are added/removed and the value does not update the token quantities. This PR fixes it.

**Screenshot**
image


SorinC6 (approved):

zuzunker (approved):
Thank you, @nistadev !

PR #3547, opened 2024-06-14
use stake address instead of stake public key in cip36 metadata

yushih:


packages/yoroi-extension/app/api/ada/lib/cardanoCrypto/catalyst.js
```diff
@@ -43,7 +43,7 @@ export function generateRegistrationMetadata(
   const registrationData = RustModule.WalletV4.encode_json_str_to_metadatum(
     JSON.stringify({
       '1': [[prefix0x(votingPublicKey), 1]],
-      '2': prefix0x(stakingPublicKey),
+      '2': prefix0x(stakingAddress),
```
stackchain, line 46: @michaeljscript can you check on mobile?

michalszorad: @stackchain I'm having a look

michalszorad: We're using `stakingPublicKeyHex`

michalszorad: The implementation is in yoroi-lib of `generateRegistrationMetadata`. Can we reuse it in extension?

yushih:
no need for this. Should be stake pub key.

packages/yoroi-extension/app/api/ada/lib/cardanoCrypto/catalyst.js
yushih: False change. Sorry. Things are correct as is.

PR #3555, opened 2024-06-20
swap limit price reset to market fix

zuzunker:
removed state-updating logic from component render. Moved it to be an effect


emurgo-rnd:
[YOEXT-1205: swap failing state update](https://emurgo.atlassian.net/browse/YOEXT-1205)

nistadev (approved):

Nebyt (approved):

PR #3580, opened 2024-07-12
fix staking page crash when pool info query fails

yushih:


emurgo-rnd:
[YOEXT-1167: The app crashes on Staking page in case the 500 error for pool info](https://emurgo.atlassian.net/browse/YOEXT-1167)

Nebyt (dismissed):

zuzunker (dismissed):

PR #3601, opened 2024-08-12
refactor: change token names 

SorinC6:
⚠️ Huge PR alert

- as the design team decided to change the token names we are forced to have this huge PR in which we are changing: child and primary token names to reflect the design all over the application
- I've been delaying this but now we need to update it as there are noticeable differences when building the UI in Figma with what we have related to the token's name


emurgo-rnd:
[YOEXT-1271: Dark theme support for Wallet, Voting and Connector features](https://emurgo.atlassian.net/browse/YOEXT-1271)

SorinC6 (commented):
/check

zuzunker (dismissed):
Thank you, @SorinC6 !

SorinC6 (commented):
/check

PR #3603, opened 2024-08-19
Refactoring and fixing

yushih:
1. Refactor the content script message handler in the background to get rid of the deeply nested codes
2. Fix connected site app icon handling
3. Remove the `protocol` parameter in the background
4. Reduce log noise
5. Fix swap broken by re-architecturing


zuzunker:
> Remove the protocol parameter in the background

🎉

packages/yoroi-extension/chrome/extension/background/handlers/content/rpc.js
```diff
@@ -0,0 +1,731 @@
...
+    continuationData,
+    uid,
+  };
+  await setConnectedSite(tabId, connectedSite);
+  chrome.windows.create({
+    ...popupProps,
+    url: chrome.runtime.getURL(`/main_window_connector.html#/signin-transaction`),
+    left: (bounds.width + bounds.positionX) - popupProps.width,
+    top: bounds.positionY + 80,
+  });
```
zuzunker, line 92: I think `return` or `await` is missing. The function itself is async and returns a promise, but this last async call result is ignored so the function might `.then` before `windows.create` actually finished

packages/yoroi-extension/chrome/extension/background/handlers/content/rpc.js
```diff
@@ -0,0 +1,731 @@
...
+}
+
+declare function _MaybeWallet(cond: true): PublicDeriver<>;
+// eslint-disable-next-line no-redeclare
+declare function _MaybeWallet(cond: false): void;
```
vsubhuman, line 97: why redeclare tho?

packages/yoroi-extension/chrome/extension/background/handlers/content/rpc.js
```diff
@@ -0,0 +1,731 @@
...
+
+type Handler<NeedConnectedWallet, ParamT, ReturnT> = {|
+  needConnectedWallet: NeedConnectedWallet,
+  syncConnectedWallet: boolean,
+  handle: HandleFuncType<NeedConnectedWallet, ParamT, ReturnT>,
+|};
+
+class _TypeCarrier<ParamT, ReturnT> {
+  toBe<NeedConnectedWallet>(
+    needConnectedWallet: NeedConnectedWallet,
```
vsubhuman, line 133: I can't find a case when the `NeedConnectedWallet` generic type does not end-up being a boolean. I think it can be replaced with `boolean` in the parameters here similar to `syncConnectedWallet` and then also completely removed from the generics in the `Handler`, the `HandleFuncType`, and the `RequestType`.

Is it added only for readability? Maybe just something like this?
```
wallet: $Call<typeof _MaybeWallet, boolean /* NeedConnectedWallet */>,
```

packages/yoroi-extension/chrome/extension/background/handlers/content/rpc.js
yushih: This is OK because from the RPC handler function's perspective, popping up the window is the last action. The whole process will resume when the popup window sends further messages.

Actually, under mv3, the service worker may stop after this.

zuzunker (approved):

PR #3606, opened 2024-08-23
fix(swap): added custom hook for verified tokens

nistadev:
**Description**
When navigating between swap page tabs, if you selected a token to swap the app crashed because the request to the api fails. This adds a trycatch block to return a default value in case of error response from api.


emurgo-rnd:
[YOEXT-1312: [Swap] The app crashes in case error is received from a Muesliswap endpoint](https://emurgo.atlassian.net/browse/YOEXT-1312)

zuzunker:
Closed in favour of https://github.com/Emurgo/yoroi-frontend/pull/3649 because this one does not work correctly

PR #3615, opened 2024-09-02
post 5.3.100 pull

zuzunker:


SorinC6 (dismissed):

zuzunker (commented):
/check

packages/yoroi-extension/chrome/extension/background/handlers/utils.js
```diff
@@ -186,16 +182,18 @@ async function getWalletState(publicDeriver: PublicDeriver<>): Promise<WalletSta
     throw new Error('unexpected missing asGetBalance result');
   }
   const balance = await canGetBalance.getBalance();
+  const rewardBalance = await canGetBalance.getBalance();
```
yushih: @vsubhuman what's point of adding `rewardBalance`?

zuzunker: @yushih , good catch! I was initially planning to have that change added and used, but then changed my mind, removed the usage of it, and forgot the declaration. The plan is to make this change later as a separate PR to review and test separately. Ultimate goal is to simplify and encapsulate as much wallet-depending APIs to be a part of the wallet itself as possible. In the particular case, it's a very common pattern when both the utxo balance and the reward balance need to be summed together to have the "total" wallet balance. Remember that `canGetBalance.getBalance()` returns only the utxo balance.

zuzunker: I have removed this now

PR #3624, opened 2024-09-12
bundle and inject code for bring

yushih:


yushih:
Integrated into https://github.com/Emurgo/yoroi-frontend/pull/3720

PR #3707, opened 2024-10-21
updated message passing

yushih:


packages/yoroi-extension/app/api/thunk.js
```diff
@@ -404,9 +404,19 @@ const callbacks = Object.freeze({
...
+  Logger.debug('get message from background:', serializedMessage);
+
+  let message;
+  try {
+    message = JSON.parse(serializedMessage);
+  } catch {
+    return;
+  }
+  if (typeof message !== 'object') {
+    return;
```
zuzunker: Let's please add console.error logs before both of these returns, needs to be visible when anything goes wrong

zuzunker (changes requested):

packages/yoroi-extension/app/api/thunk.js
```diff
@@ -106,9 +106,12 @@ declare var chrome;
 
 // UI -> background queries:
 
-export function callBackground<T, R>(message: T): Promise<R> {
+export function callBackground<T, R>(message: {| type: string, request?: Object |}): Promise<R> {
   return new Promise((resolve, reject) => {
-    window.chrome.runtime.sendMessage(message, response => {
+    const serializedMessage = { type: message.type, request: JSON.stringify(message.request || null) };
```
zuzunker:
```suggestion
    const serializedMessage = { type: message.type, request: JSON.stringify(message.request ?? null) };
```

packages/yoroi-extension/app/api/thunk.js
```diff
@@ -404,9 +412,21 @@ const callbacks = Object.freeze({
...
+chrome.runtime.onMessage.addListener((serializedMessage, _sender, _sendResponse) => {
   //fixme: verify sender.id/origin
-  Logger.debug('get message from background:', JSON.stringify(message, null, 2));
+  Logger.debug('get message from background:', serializedMessage);
+
+  let message;
+  try {
+    message = JSON.parse(serializedMessage);
+  } catch {
+    Logger.error('underializable message');
```
vsubhuman, line 423:
```suggestion
    Logger.error('unserializable message');
```

zuzunker:
```suggestion
  } catch (e) {
    Logger.error('unserializable message: ' + serializedMessage + ' | Error: ' + stringifyError(e));
```

packages/yoroi-extension/app/api/thunk.js
```diff
@@ -404,9 +412,21 @@ const callbacks = Object.freeze({
...
+    return;
+  }
+  if (typeof message !== 'object') {
+    Logger.error('unrecognizable message');
```
zuzunker:
```suggestion
    Logger.error('unrecognizable message type: ' + (typeof message) + ' (expected object); Original message: ' + serializedMessage);
```

packages/yoroi-extension/chrome/extension/background/handlers/yoroi/index.js
```diff
@@ -107,6 +107,7 @@ export function getHandler(typeTag: string): ?Handler {
     return async (request, send, sendResponse) => {
       try {
         const result = await handler(request.request);
+        console.debug(`BACKGROUND [${typeTag}] sending result: `, JSON.stringify(result));
         sendResponse(result);
```
zuzunker, line 112: @yushih , every `chrome.runtime.onMessage.addListener` in the `thunk,js` now expects to receive `serializedMessage`.

Should this line `sendResponse(result)` change to always perform serialisation?

And then remove explicit in-place serialisation where it's performed for txs?
```
 // $FlowIgnore
return JSON.stringify(txs);
```

packages/yoroi-extension/app/api/thunk.js
yushih: The message has been logged on L417 and since the only thing in the `try` block is `JSON.parse`, I think the error can be easily inferred.

packages/yoroi-extension/app/api/thunk.js
yushih: The same argument as above.

packages/yoroi-extension/chrome/extension/background/handlers/yoroi/index.js
yushih: There are two different messages "channels" between thunk.js and the background:

1. API call, initialized by the front end by calling `callBackground`. Handled and replied by the background.
2. Notifications from the background to the UI. Received by `thunk.js` by calling `chrome.runtime.onMessage.addListener((serializedMessage, _sender, _sendResponse) => {` 

>  every chrome.runtime.onMessage.addListener in the thunk,js now expects to receive serializedMessage

this is 2

>  this line sendResponse(result)  

this is 1

> Should this line sendResponse(result) change to always perform serialisation?
And then remove explicit in-place serialisation where it's performed for txs?

We *can* but by now it's not necessary since everything looks OK now.

packages/yoroi-extension/app/api/thunk.js
zuzunker: The message logged on the L417 is in DEBUG, we need error logs as extended as possible for the case when users will be reporting issues. Very often we receive issue reports as only error messages from consoles or even their screenshots.

Not having full information about what exactly has happened in the error is completely developing-process-centric, thinking we can get into full debug mode at any moment, instead of thinking of cases when we gonna have minimal information possible from a user report with no clear way to reproduce what's failing.

Please always apply the simple rule of thumb:
1. DEBUG logs in the middle of **normal** flow are for tracing during development/investigation - they may contain minimal information needed. I.e. trace logs might be usable even for cases when errors are not happening in this particular case. Their main point is just to be able to track what is happening here during the normal operations.
2. ERROR logs are only used in branches that are NOT supposed to happen at all in a normal case , so when this branch happens - AS MUCH information as possible must be clearly outputted to help any future investigation

Understanding this difference may literally save us days of investigation in the future.

packages/yoroi-extension/chrome/extension/background/handlers/yoroi/index.js
zuzunker: Understood perfectly well now. Thank you!

Nebyt (dismissed):

zuzunker (dismissed):

Nebyt (approved):

PR #3760, opened 2024-11-27
fixed YOEXT-1464

Nebyt:
https://emurgo.atlassian.net/browse/YOEXT-1464

I made the condition to not show the "Verify on hardware wallet" button in case the current is the Address book. Because it is actually pointless, because the Address book contains not own addresses.

### DIFF
--- a/packages/yoroi-extension/app/containers/wallet/WalletReceivePage.js
+++ b/packages/yoroi-extension/app/containers/wallet/WalletReceivePage.js
@@ -45,7 +45,6 @@ import WalletReceiveRevamp from '../../components/wallet/WalletReceiveRevamp';
 import UnmangleTxDialogContainer from '../transfer/UnmangleTxDialogContainer';
 import StandardHeaderRevamp from '../../components/wallet/receive/StandardHeaderRevamp';
 import { maybe } from '../../coreUtils';
-import { ROUTES } from '../../routes-config';
 
 type Props = {|
   ...StoresAndActionsProps,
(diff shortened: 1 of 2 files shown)


packages/yoroi-extension/app/containers/wallet/WalletReceivePage.js
```diff
@@ -119,6 +120,7 @@ class WalletReceivePage extends Component<AllProps> {
     };
 
     const isHwWallet = publicDeriver.type !== 'mnemonic';
+    const isAddressBook = this.props.stores.router.location.pathname.startsWith(`${ROUTES.WALLETS.RECEIVE.ROOT}/addressBook/all`)
```
zuzunker: Please define the address book flag by the code already used on this same page:
```
addressTypeStore.meta.name.group === AddressGroupTypes.addressBook
```

You can then also replace multiple usages of this code expression by the use of your new flag `isAddressBook`

zuzunker (changes requested):

packages/yoroi-extension/app/containers/wallet/WalletReceivePage.js
Nebyt: Thank you! I didn't notice that. Will do shortly.

Nebyt: Fixed.

zuzunker (approved):

PR #3762, opened 2024-11-28
Network switch API & UI

yushih:
This is a backward-compatible change. 
Jira task: https://emurgo.atlassian.net/browse/YOEXT-1385


SorinC6 (commented):
@yushih there are 2 things that I noticed on the UI:
 - can you check the dark theme impl in the dialog as well, it seems like the text color is not correct
 - when switching networks all the app page is refreshing - is this intentional?   
 - I see that some of the components were done using Class components, maybe in the next iteration, or if we have time, we can change and use functional ones when building new UI parts from now on  

Screenshot 2025-01-10 at 13 23 44

zuzunker (approved):
Thank you, @yushih !

zuzunker (commented):
/check

PR #3804, opened 2025-01-13
refactor(dapp-center): prepare page structure

aatsindev:
**Description**
- [x] Prepare page structure


packages/yoroi-extension/app/UI/features/dapp-center/common/mockData.ts
```diff
@@ -0,0 +1,6 @@
+//TODO: ADD ALL THE MOCK DATA FOR RENDERING UI NEW
+const mockData = {
+
+};
```
SorinC6, line 4: this can be deleted?

SorinC6 (approved):
The main structure and layout looks good so far!

packages/yoroi-extension/app/UI/features/dapp-center/common/mockData.ts
aatsindev: Depends if later we need to use it for the connector, right now is only for keeping the structure, next step would be creating the components and logic.

zuzunker (commented):
/check

PR #3844, opened 2025-02-06
Correct connenction

Nebyt:
Bug: https://emurgo.atlassian.net/browse/YOEXT-1719


packages/yoroi-extension/chrome/content-scripts/inject.js
```diff
@@ -116,6 +116,8 @@ function listenToBackgroundServiceWorker() {
                             }
                         }, location.origin);
                     }
+                } else {
+                    connected = true;
```
zuzunker, line 120: I am not sure this is the correct place for this flag. Let me investigate this in greater details first. Marking to not be merged for now

zuzunker (changes requested):

PR #3923, opened 2025-03-31
the background sync should only sync wallets of the current network

yushih:


zuzunker (approved):
Thank you, @yushih !

zuzunker (commented):
/check

PR #3970, opened 2025-05-05
feat(nfts): migrate to TS

aatsindev:
**Description**
Migrate page to TS.

Task: https://emurgo.atlassian.net/browse/YOEXT-1523

Steps:
- [x] Extract components and folder structure
- [x] Refactor initial page and create use cases

**Screenshots**
image
image


packages/yoroi-extension/app/UI/components/Input/SearchInput.tsx
```diff
@@ -30,7 +30,6 @@ export const SearchInput = (props: InputProps) => {
   return (
     <StyledInput
       inputRef={ref}
-      disableUnderline
```
zuzunker, line 33: @aatsindev , why this change in a shared component? 🤔

aatsindev: Error in the console, so basically react doesn't read it anyway (theoretically). @SorinC6 created it, so he may have some context of why it was really there in the first time 🤔

packages/yoroi-extension/app/UI/features/nfts/common/helpers/index.ts
```diff
@@ -0,0 +1,6 @@
+export function imageExists(imageSrc: string, onload: () => void, onerror: () => void) {
```
zuzunker: Rename please into `imageLoadTest`

packages/yoroi-extension/app/UI/features/nfts/common/components/NftsHeader.tsx
```diff
@@ -0,0 +1,76 @@
...
+const messages = defineMessages({
+  searchNFTs: {
+    id: 'wallet.nftGallary.search',
+    defaultMessage: '!!!Search NFTs',
+  },
+  nftsCount: {
+    id: 'wallet.nftGallary.details.nftsCount',
+    defaultMessage: '!!!NFTs ({number})',
+  },
+});
```
SorinC6: For the strings maybe we can bring all of them in `common/hooks/useStrings` hook - to have all intl text usage for each feature in a single place

packages/yoroi-extension/app/UI/features/nfts/common/hooks/useNfts.ts
```diff
@@ -0,0 +1,42 @@
...
+          image: getImageFromTokenMetadata(policyId, hexName, token.info.Metadata),
+        };
+      });
+
+    setNftsList(nfts);
+    setLoading(false);
+  }, []);
+
+  return { nftsList, loading };
+};
```
SorinC6, line 42: Related to this hook, as each time we will need to get the nfts for a wallet will need to replicate and do this logic?  I was thinking on a more general solution in `createCurrentWalletInfo` as to already have this in context, or maybe we can create a general hook (for all features to use it if needed) and just use it in the feature context to get the wallet nfts and then just send it from the context 🤔

aatsindev: That's a good point, we could improve it further on the next iterations, as it would take much more workload as initially estimated and won't get through it. I'm thinking of this should be at Context level, the hook itself is good because it has the encapsulated logic and accesses the information from the context.

packages/yoroi-extension/app/UI/features/nfts/common/components/NftsHeader.tsx
zuzunker: Dunno if makes much sense. Same question with icons and everything else long term. Keeping everything in one place means loading everything even when user never visits some pages

aatsindev: I think he means using the strings inside a scoped hook @vsubhuman so that only loads the specific strings for the pages in the context.

SorinC6: hmm, still is weird to have all the strings spread along different pages in a feature but if is better we can do it like this as well

zuzunker: In that case it makes sense if it's a separate version of `useStrings` for each feature locally. I got confused by `common/hooks` , I think @SorinC6 might have meant to have ALL strings for all features in one place 🤔 That one I don't like that much, as it usually also gets deorganised and garbaged over time in such cases

packages/yoroi-extension/app/UI/features/nfts/common/hooks/useNfts.ts
zuzunker: Agreed to keep it out of the current context. We can create a ticket to extract it into a context provider and consider its priority then

packages/yoroi-extension/app/UI/features/nfts/common/components/NftsHeader.tsx
zuzunker: > hmm, still is weird to have all the strings spread along different pages in a feature but if is better we can do it like this as well

Strings definitely should be in one place for **one** feature. Totally agreed. It seems I have got the message wrong then from the start. Definitely better to extract all "nft gallery" strings into a single hook 👍

packages/yoroi-extension/app/UI/components/Input/SearchInput.tsx
SorinC6: yeah, I see that have this warning using it 🤔 let's keep it like as it seems is not barking any UI for the `OutlinedInput`

SorinC6 (dismissed):

zuzunker (approved):

PR #3972, opened 2025-05-05
dApp tests fixes after components update

Nebyt:


Nebyt (commented):
/dapp-general

Nebyt (commented):
/dapp-general

PR #4008, opened 2025-05-20
staking key signing fix

zuzunker:
https://emurgo.atlassian.net/browse/YOEXT-1981


packages/yoroi-extension/app/api/ada/transactions/shelley/transactions.js
```diff
@@ -1084,6 +1088,46 @@ async function newAdaUnsignedTxFromUtxoForConnector(
...
+    unsignedTx,
+    wallet.getParent().getPublicDeriverLevel(),
+    RustModule.WalletV4.Bip32PrivateKey.from_hex(normalizedKey.prvKeyHex),
+    stakingKey,
+    metadata,
+    otherRequiredSigners,
+  )
+}
+
 export function signTransaction(
```
yushih, line 1131: This function is no longer used by other modules now that there is `signTransactionFromWallet`  so maybe remove the `export` ?

yushih (approved):

packages/yoroi-extension/app/api/ada/transactions/shelley/transactions.js
zuzunker: @yushih , it's actually used directly in one place for signing transfer from Byron addresses txs, so the same kind of the wallet is not available for those inputs and calling `signFromWallet` would not work there. The generic signing is compatible and it's just passing null for staking key so that part is ignored.

PR #4054, opened 2025-06-09
Set USD as default currency

jarstorm:
Task: https://emurgo.atlassian.net/browse/YOEXT-715


jarstorm:
I tried to use DEFAULT_FIAT_PAIR variable (there are two defined) instead of **USD** but was not possible because there was a problem with the import.

zuzunker (dismissed):

Nebyt (commented):
/check

Nebyt (approved):

PR #4063, opened 2025-06-10
Avoid address whitespaces send assets

jarstorm:
task: https://emurgo.atlassian.net/browse/YOEXT-1582


packages/yoroi-extension/app/components/wallet/send/WalletSendFormRevamp.js
```diff
@@ -552,6 +552,12 @@ export default class WalletSendFormRevamp extends Component<Props, State> {
     return this.context.formatMessage(errMsg, values);
   }
 
+  avoidWhitespacesAddress(e: KeyboardEvent) {
+    if (e.key === ' ') {
+      e.preventDefault();
+    }
+  }
```
zuzunker: It is still possible to paste-in a whitespace with this change. According to the task the main issue is not really entering the spaces but the fact that an address is incorrect when there is a space at the beginning or the end. I'm not sure how often users do actual manual typing in this field, pasting a copied address is definitely 99.9% of use case. Running a trim of the value string on any change might be a good approach, which will also cover manually added whitespaces as well, as they will get removed right as they are added.

zuzunker (changes requested):

packages/yoroi-extension/app/components/wallet/send/WalletSendFormRevamp.js
jarstorm: There is an autoTrim flag to be able to do that but it's not working for me (maybe it's because of package version): https://github.com/foxhound87/mobx-react-form-demo/blob/25c2093c200dddbfae28f7bf9877d103567c2233/src/forms/setup/registerMaterial.ts#L20

So, my solution could be to trim the value in the validator (it's not he ideal but at least it works)

zuzunker (approved):

zuzunker (commented):
/check

PR #4076, opened 2025-06-16
fix: notification trigger does not have updated location

banklesss:
https://emurgo.atlassian.net/browse/YOEXT-2050


packages/yoroi-extension/app/UI/features/notifications/module/NotificationsProvider.tsx
```diff
@@ -115,26 +115,32 @@ export default function NotificationsProvider({ children, appLoadedSlots = {}, w
...
     const theme = await lsApi.getUserThemeMode();
     const notifyWallet = await isActiveSettingsForWallet();
     // Early returns:
     // return if settings are off
+    const currentLocation = locationRef.current;
+
+
     if (!notifyWallet) return;
-    // return if we're on the same route as the event redirection
+// return if we're on the same route as the event redirection
```
banklesss, line 133:
```suggestion
    // return if we're on the same route as the event redirection

```

packages/yoroi-extension/app/UI/features/notifications/module/NotificationsProvider.tsx
```diff
@@ -146,7 +152,7 @@ export default function NotificationsProvider({ children, appLoadedSlots = {}, w
       type,
       id: id || String(Date.now()),
     });
-  };
+  }, []);
```
jarstorm: Is **[]** working?

In order to make it work, shouldn't be **[id]**?

banklesss: I have reviewed the function and I have seen that useCallback is unnecessary because that function is held in a memo!

jarstorm (approved):

Nebyt (approved):

PR #4081, opened 2025-06-19
Notification duration setting

yushih:
Task: https://emurgo.atlassian.net/browse/YOEXT-1756


jarstorm (approved):

packages/yoroi-extension/app/UI/features/notifications/useCases/NotificationsSettings/NotificationsSettings.tsx
```diff
@@ -72,6 +74,58 @@ const NotificationsSettings = ({ selectedWalletId }: Props) => {
...
+            '::placeholder': { color: 'grayscale.600' },
+          }}
+          component="input"
+          type="text"
+          variant="body1"
+          color="ds.text_gray_medium"
+          placeholder="0"
+          bgcolor="ds.bg_color_max"
+          value={strings.durationDescription(duration)}
+          onClick={openDurationDialog}
```
SorinC6: As impovement on this component we can move the styles (sx props) outside of the render and use `styled` prop from MUI. When have a lots of styles inside it is kind of hard to read the component itself

packages/yoroi-extension/app/stores/toplevel/PushNotificationStore.js
```diff
@@ -47,6 +64,26 @@ export default class PushNotificationStore<
...
+        throw new Error('push notification metadata not loaded');
+      }
+
+      this.metadata.duration = duration;
+    });
+    if (!this.metadata) {
+      throw new Error('push notification metadata not loaded');
+    }
+    (new LocalStorageApi()).savePushNotificationMetadata(this.metadata);
+  }
```
SorinC6, line 86: In the future maybe we can add the logic for any feature inside the context api and not in the MobX store (if possible). The idea is for each feature logic to be inside the context and not the mobX store, as this later we can remove the store as well if possible

packages/yoroi-extension/app/containers/settings/categories/WalletSettingsPage.js
```diff
@@ -71,7 +73,15 @@ export default class WalletSettingsPage extends Component <StoresProps> {
...
-          <NotificationsSettings selectedWalletId={selectedWalletId} />
+          <NotificationsSettings
+            selectedWalletId={selectedWalletId}
+            duration={stores.pushNotificationStore.duration}
+            openDurationDialog={() => {
+              stores.uiDialogs.open({
+                dialog: NotificationDurationDialog,
+              });
+            }}
+          />
```
SorinC6, line 84: For the next modals we should use the hook in the new UI if possible 
```
  const { closeModal, openModal } = useModal();
```
Maybe was already done like this before adding the hook - but adding this as a sugestion only for later

SorinC6 (approved):
Looks good to me! I just added some suggestion for future implementation

Nebyt (approved):

PR #4091, opened 2025-06-24
Delegate on preprod if cannot get pool info

jarstorm:
task: https://emurgo.atlassian.net/browse/YOEXT-2085


yushih (dismissed):

SorinC6 (dismissed):

jarstorm:
The problem is here:
image

So, if this endpoint it's not available you cannot delegate to any pool.
But should I add a try/catch there? We are loading pool info because of a reason, not sure if we can skip this call or we should fail in this case.
@yushih @vsubhuman could you please help me here?

jarstorm (commented):
/check

Nebyt (approved):

pkEMURGO (approved):

jarstorm (commented):
/check

jarstorm (commented):
/check

PR #4105, opened 2025-07-01
Tokens displayed previous selected wallet

jarstorm:
Task: https://emurgo.atlassian.net/browse/YOEXT-2103


yushih (dismissed):

yushih (commented):
I think there's another subtle bug: if there is an outstanding filter (`keyword !== ''`) while wallet changes, the filter will lose effect. The solution is to use `useMemo` to calculate the filtered token list (`tokenList`).

packages/yoroi-extension/app/UI/features/portfolio/useCases/Wallet/PortfolioWallet.tsx
```diff
@@ -1,80 +1,84 @@
...
+      clearTimeout(timeout);
+      timeout = setTimeout(() => {
+        ampli.portfolioTokensListSearchActivated({ search_term: lowercaseKeyword });
+      }, 500); // 0.5s requirement
+    };
+
+    if (lowercaseKeyword.length > 0) sendMetrics();
+
+    return () => clearTimeout(timeout);
+  }, [keyword]);
```
jarstorm, line 52: @yushih not sure what is the solution you want to do.
What if we add it here **ftAssetList** as a dependency and remove the code I introduced?

yushih: I mean, `tokenList` should be calculated (using the value of `keyword` and `ftAssetList`), and wrapped inside `useMemo`.

jarstorm: @yushih I updated the code

SorinC6 (approved):

Nebyt (approved):

Nebyt (commented):
/check

PR #4109, opened 2025-07-02
Buy and sell disclaimer always shown

jarstorm:
task: https://emurgo.atlassian.net/browse/YOEXT-1997


SorinC6 (dismissed):

yushih (dismissed):
OK. But for a cleaner solution, getting the disclaimer flag from storage should be in `BuySellDialog` instead of `DisclaimerDialog`. Whether the disclaimer dialog is shown should be a concern of its user instead of itself.

jarstorm:
> OK. But for a cleaner solution, getting the disclaimer flag from storage should be in `BuySellDialog` instead of `DisclaimerDialog`. Whether the disclaimer dialog is shown should be a concern of its user instead of itself.

Yes, you are right. I didn't want to change a lot of code, but in this case it's better to do it like you mentioned

SorinC6 (approved):

pkEMURGO (approved):

PR #4111, opened 2025-07-02
Enter recovery phrase error

jarstorm:
Task: https://emurgo.atlassian.net/browse/YOEXT-811


packages/yoroi-extension/app/components/wallet/restore/steps/phrase/RestoreRecoveryPhraseForm.js
```diff
@@ -125,23 +130,42 @@ export default class RestoreRecoveryPhraseForm extends Component<Props, State> {
...
+    this.checkIfInitialPhraseIsValid();
   }
 
+  checkIfInitialPhraseIsValid: () => void = () => {
+    const { isValidMnemonic, numberOfMnemonics } = this.props;
+    const phrase = this.generateRecoveryPhrase();
+    const allWordsEntered = this.checkAllWordsEntered(numberOfMnemonics, phrase);
+    const isValidPhrase = allWordsEntered && isValidMnemonic(phrase);
+    if (isValidPhrase) {
+      this.submit();
```
zuzunker, line 146: @jarstorm , I see the original issue is about the fact that:
1. User enters a phrase
2. User goes to next step
3. User returns back to this step and enters a **new different** phrase
4. The new different phrase is not accepted as valid

I don't see how this change would resolve this problem. What the code seems to be doing is checking if the `initialRecoveryPhrase` (which is only present when the user goes back from the next step) is valid, and if so it just submits the form automatically. If I understand correctly this will just block the user from ever being able to input a different phrase. We don't want to do that. User should be able to change the recovery phrase at any moment and the new phrase should be correctly validated and accepted.

packages/yoroi-extension/app/components/wallet/restore/steps/phrase/RestoreRecoveryPhraseForm.js
```diff
@@ -82,15 +83,19 @@ export default class RestoreRecoveryPhraseForm extends Component<Props, State> {
       : new Array(this.props.numberOfMnemonics).fill('');
   };
 
+  generateRecoveryPhrase: void => Array<{| value: string, label: string |}> = () => {
```
zuzunker: Please apply better function naming, the function `generateRecoveryPhrase` does not in fact generate any recovery phrases, but gets the `initialRecoveryPhrase` prop as object entries.

zuzunker (changes requested):

packages/yoroi-extension/app/components/wallet/restore/steps/phrase/RestoreRecoveryPhraseForm.js
jarstorm: The main problem here was that the form can be only submitted once due to this line:
image

So, if the phrase is valid you cannot change this phrase anymore, because here it's not calling submit again.

What I try with my solution is, trigger **submit** method everytime there is a change in one input and there are **numberOfMnemonics** words enetered. 
And also validate the phrase once you enter into the screen (when user press go back button)

Nebyt (approved):

packages/yoroi-extension/app/components/wallet/restore/steps/phrase/RestoreRecoveryPhraseForm.js
zuzunker: Understood now. Sounds good and thank you for clarifying

zuzunker (approved):

jarstorm:
@Nebyt some of the tests are failing

Nebyt (commented):
/check

Nebyt (commented):
/check

PR #4138, opened 2025-07-15
Added build-nightly-test.yml

pkEMURGO:
Jira Ticket - https://emurgo.atlassian.net/browse/YOEXT-2131

### DIFF
--- a/.github/workflows/build-nightly-test.yml
+++ b/.github/workflows/build-nightly-test.yml
@@ -24,15 +24,6 @@ jobs:
         run: |
           . install-all.sh
 
-      - name: Run Lint
-        run: npm run eslint
-
-      - name: Run TypeScript
-        run: npm run tsc
-
-      - name: Run Tests
-        run: npm run test
-
       - name: Build Nightly app
         working-directory: ./packages/yoroi-extension
         run: |
--- a/.github/workflows/build-nightly-test.yml
+++ b/.github/workflows/build-nightly-test.yml
@@ -1,7 +1,7 @@
 name: Build Nightly for Testing
 
 on:
-  workflow_dispatch:  # You can trigger manually from the GitHub Actions UI
+  workflow_dispatch:
 
 jobs:
   build-nightly:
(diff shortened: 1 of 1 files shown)


jarstorm (dismissed):

jarstorm:
@Nebyt could you please check this?

.github/workflows/build-nightly-test.yml
```diff
@@ -0,0 +1,50 @@
...
+          . install-all.sh
+
+      - name: Run Lint
+        run: npm run eslint
+
+      - name: Run TypeScript
+        run: npm run tsc
+
+      - name: Run Tests
+        run: npm run test
```
Nebyt: Actually, you do need these checks

.github/workflows/build-nightly-test.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Build Nightly App
+        working-directory: ./packages/yoroi-extension
+        run: |
+          npm run prod:nightly
+
```
Nebyt, line 31: Please add a check that the build is correct, like we have here https://github.com/Emurgo/yoroi-frontend/blob/develop/.github/workflows/e2e-tests.yml#L57-L63

.github/workflows/build-nightly-test.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Zip Build Folder
+        working-directory: ./packages/yoroi-extension
+        run: |
+          cd dist && zip -r ../yoroi-nightly.zip . && cd ..
```
Nebyt: You don't need this step, because there is already the extension zip archive in case on prod nightlies build.

Archive location: packages/yoroi-extension/Yoroi Nightly.zip

.github/workflows/build-nightly-test.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Upload Build Artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: yoroi-nightly-build
+          path: ./packages/yoroi-extension/yoroi-nightly.zip
```
Nebyt: Use the zip archive as pointed in the comment above.

Nebyt (changes requested):

Nebyt (approved):

PR #4154, opened 2025-07-17
Show swap button only in mainnet

jarstorm:
task: https://emurgo.atlassian.net/browse/YOEXT-1743


jarstorm:
Not sure about testnet check as it seems there are multiple ways to do it

SorinC6 (approved):

Nebyt (approved):

PR #4169, opened 2025-07-22
Fix launch environment

jarstorm:
Task: https://emurgo.atlassian.net/browse/YOEXT-1707

The storage variables does not depend on the selected network, it depends on the env file you launched from.

So, do we want/need to change this? I don't think it worth it to be honest.


zuzunker:
I'd rather not

jarstorm:
We should not implement this solution

PR #4198, opened 2025-08-07
show incoming/outgoing assets info in notification toast

yushih:


packages/yoroi-extension/app/UI/features/notifications/module/NotificationsProvider.tsx
```diff
@@ -150,11 +134,79 @@ export default function NotificationsProvider({ children, appLoadedSlots = {}, w
         break;
     }
 
+    let title;
```
jarstorm, line 166: What if we extract this title calculation insied a method called getTitle or something like that? Maybe this could be easier to read.

jarstorm (approved):

pkEMURGO (approved):

PR #4221, opened 2025-08-19
Read backend env vars from OS

jarstorm:
task: https://emurgo.atlassian.net/browse/YOEXT-2185


yushih:
@jarstorm seems the flow check is broken

PR #4241, opened 2025-08-25
Create single address toggle

jarstorm:


packages/yoroi-extension/app/api/localStorage/index.js
```diff
@@ -173,6 +174,12 @@ export default class LocalStorageApi {
 
   unsetCardanoCardModalClosed: void => Promise<void> = () => removeLocalItem(storageKeys.CARDANO_CARD_MODAL_CLOSED);
 
+  // ========== SINGLE_ADDRESS Mode ========== //
+
+  getSingleAddressMode: void => Promise<?string> = () => getLocalItem(storageKeys.SINGLE_ADDRESS_MODE);
```
jarstorm, line 179: If I'm not mistaken this is only a flag, but it should be per wallet.

SorinC6: yes but we are adding the key value pair as a string in the local storage `[walletId]:isSingleAddress`

PR #4261, opened 2025-08-29
Swap UI Settings 

SorinC6:
- Fixed and update settings implementation 
- https://emurgo.atlassian.net/browse/YOEXT-2137


jarstorm (dismissed):

jarstorm (dismissed):

SorinC6 (commented):
/check

SorinC6 (commented):
/check

packages/yoroi-extension/app/UI/features/swap-new/common/components/AssetInput.tsx
```diff
@@ -80,7 +80,7 @@ export const AssetInput: React.FC<AssetInputProps> = ({ direction, onAssetSelect
       if (!touched) {
         return 'Select token';
       }
-      return tokenInputInfo?.ticker ?? tokenInputInfo?.name;
+      return tokenInput.tokenId === '.' ? primaryTokenInfo.name : (tokenInputInfo?.ticker ?? tokenInputInfo?.name);
```
jarstorm, line 83: id **'.'** means there is no token?
We should try to create constants for this

packages/yoroi-extension/app/UI/features/swap-new/common/components/Modals/SettingsModalContent.tsx
```diff
@@ -81,41 +100,27 @@ const SlipageOptions = ({ setIsManualSlippage, setSelectedSlippage, isManualSlip
...
-    const value = event.target.checked;
-    if (!value && !muesliswap) {
-      setMuesliswap(true);
-    }
-    setDexHunter(value);
-  };
-
-  const handleMuesliswapToggle = (event: React.ChangeEvent<HTMLInputElement>) => {
-    const value = event.target.checked;
+  const autoSelected = routingPreferance === 'auto';
```
jarstorm: all these strings could be constants?

packages/yoroi-extension/app/UI/features/swap-new/common/components/Modals/SettingsModalContent.tsx
```diff
@@ -81,41 +100,27 @@ const SlipageOptions = ({ setIsManualSlippage, setSelectedSlippage, isManualSlip
...
+  const dexHunter = routingPreferance === 'auto' || routingPreferance === 'dexhunter' || routingPreferance === 'both';
+  const muesliswap = routingPreferance === 'auto' || routingPreferance === 'muesliswap' || routingPreferance === 'both';
 
-    if (!value && !dexHunter) {
-      setDexHunter(true);
-    }
-    setMuesliswap(value);
+  const handleDexHunterToggle = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const checked = e.target.checked;
+    setRoutingPreferance(checked ? 'dexhunter' : 'muesliswap');
```
jarstorm: constants here too please

packages/yoroi-extension/app/UI/features/swap-new/common/components/Modals/SettingsModalContent.tsx
```diff
@@ -81,41 +100,27 @@ const SlipageOptions = ({ setIsManualSlippage, setSelectedSlippage, isManualSlip
...
-      setMuesliswap(true);
-    }
+  const handleMuesliswapToggle = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const checked = e.target.checked;
+    setRoutingPreferance(checked ? 'muesliswap' : 'dexhunter');
   };
 
+  const handleAutoChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const checked = e.target.checked;
+    setRoutingPreferance(checked ? 'auto' : 'both');
```
jarstorm: and here

packages/yoroi-extension/app/UI/features/swap-new/module/SwapContextProvider.tsx
```diff
@@ -69,13 +69,13 @@ export const SwapContextProvider = ({ children, currentWallet, stores }: any) =>
       storage,
       network: Chain.Network.Mainnet,
       stakingKey: String(stakingKey),
-      address: walletAddresses[0],
+      address: walletAddresses[1],
```
jarstorm, line 72: Why is this position 0 or 1 of the array?
I mean, can we create a constant describing what is this?
If the first value it's not valid because of whatever reason, maybe have a comment?

packages/yoroi-extension/app/UI/features/swap-new/module/SwapContextProvider.tsx
```diff
@@ -563,6 +564,7 @@ export type SwapContext = SwapState & {
   isCreateOrderLoading: boolean;
   isEstimateOrderLoading: boolean;
   explorer: { tokenInfo: { name: string; baseUrl: string } };
+  swapManager: any;
```
jarstorm, line 567: do we have a type for swapmanager?

packages/yoroi-extension/app/UI/features/swap-new/common/components/AssetInput.tsx
```diff
@@ -80,7 +80,7 @@ export const AssetInput: React.FC<AssetInputProps> = ({ direction, onAssetSelect
       if (!touched) {
         return 'Select token';
```
jarstorm, line 81: this needs to be a constant? Is this a label we show in UI (then needs to be internationalized) ir if we use this to a comparision we need a constant

packages/yoroi-extension/app/UI/features/swap-new/common/components/Modals/SettingsModalContent.tsx
```diff
@@ -1,17 +1,26 @@
...
+  const { closeModal } = useModal();
 
+  const [routingPreferance, setRoutingPreferance] = useState<any>(swapForm.selectedProtocol.value || 'auto');
+  const [selectedSlippage, setSelectedSlippage] = useState(String(swapForm.slippageInput.value || '1'));
+  const [isManualSlippage, setIsManualSlippage] = useState(!defaultSlippages.includes(selectedSlippage));
   const inputRef = useRef<HTMLInputElement | null>(null);
   const strings = useStrings();
 
+  useEffect(() => {
+    if (swapForm.slippageInput.value === 0) setSelectedSlippage(String(swapForm.slippageInput.value));
```
jarstorm: this is updated only in case it's 0?

packages/yoroi-extension/app/UI/features/swap-new/common/components/Modals/SettingsModalContent.tsx
```diff
@@ -1,17 +1,26 @@
...
 const defaultSlippages = ['0', '0.1', '0.5', '1', '2', '3', '5', '10'];
 
 export const SettingsModalContent = () => {
-  const [selectedSlippage, setSelectedSlippage] = useState('1');
-  const [isManualSlippage, setIsManualSlippage] = useState(!defaultSlippages.includes(selectedSlippage));
+  const { swapManager, swapForm } = useSwapRevamp();
+  const { closeModal } = useModal();
 
+  const [routingPreferance, setRoutingPreferance] = useState<any>(swapForm.selectedProtocol.value || 'auto');
+  const [selectedSlippage, setSelectedSlippage] = useState(String(swapForm.slippageInput.value || '1'));
```
jarstorm: here slippage it's a string but bellow we are comparing this to 0 (as number)

jarstorm (changes requested):

packages/yoroi-extension/app/UI/features/swap-new/common/components/Modals/SettingsModalContent.tsx
SorinC6: is case of 0 is reseting the state, as a quick solution I was just checking this case

packages/yoroi-extension/app/UI/features/swap-new/module/SwapContextProvider.tsx
SorinC6: first in array is enterprise address and second is the format we need, I have a fix next to handle this before when mapping the wallet address

packages/yoroi-extension/app/UI/features/swap-new/common/components/Modals/SettingsModalContent.tsx
SorinC6: yes, will add constants for all of this

packages/yoroi-extension/app/UI/features/swap-new/common/components/AssetInput.tsx
SorinC6: '.' Is for checking that id is for primary token - This will change in the next PR when we are adding the inital pair as USDA

packages/yoroi-extension/app/UI/features/swap-new/common/components/AssetInput.tsx
SorinC6: Yes, this string is hardcoded for now, but will be replaced next with the Initial Pair as `USDA`

packages/yoroi-extension/app/UI/features/swap-new/common/components/Modals/SettingsModalContent.tsx
SorinC6: The other cases are updating the state when using `useState` - its right above in the same component

packages/yoroi-extension/app/UI/features/swap-new/common/components/Modals/SettingsModalContent.tsx
SorinC6: Yes, it is a bit confusing as we have different formats in swapForm context (as a number) and in the component state as String. Below I'm checking the value from swapForm context data - butin component is a String as I save it in state. Let me check if I can convert it to only one format

packages/yoroi-extension/app/UI/features/swap-new/module/SwapContextProvider.tsx
SorinC6: will check, but as this was not final, I left it like this until we have the final form for it

SorinC6 (commented):
/check

jarstorm (approved):

Nebyt (approved):

PR #4289, opened 2025-09-09
Fix double call to enable error

jarstorm:
Task: https://emurgo.atlassian.net/browse/YOEXT-1925


jarstorm (commented):
/check

SorinC6 (approved):

jarstorm:
This PR is not a good solution for this problem

PR #4292, opened 2025-09-09
Backend calls should depend only on selected network

jarstorm:
Task: https://emurgo.atlassian.net/browse/YOEXT-2214


jarstorm (commented):
/check

SorinC6 (dismissed):
The change looks ok and doesn't seems to affect the overall beaviour but not sure how to actually check it as seems the calls are the same 🤔

jarstorm:
> The change looks ok and doesn't seems to affect the overall beaviour but not sure how to actually check it as seems the calls are the same 🤔

I think there were some BE calls depending on if  **isDev** instead of selected network.

Nebyt (approved):

PR #4296, opened 2025-09-10
Remove ADA fiat pair

jarstorm:
Task: https://emurgo.atlassian.net/browse/YOEXT-1636


SorinC6 (dismissed):

yushih (changes requested):
This change is not enough. It does prevent ADA from appearing in the dropdown menu but for users who have selected ADA or whose unit of account option is defaulted to ADA, this changes nothing.

yushih:
Please note that `localStorageApi.getUnitOfAccount()` may return `{ enabled: false, currency: null}` , when the user has explicitly set the unit of account to ADA. We must handle this properly. Please make sure this scenario is properly handled.

jarstorm:
@yushih do you think this is enough to fixe the problem?
image

So, everytime we want to get the value from localstorage try to override it if it's set to ADA

yushih (dismissed):

Nebyt (approved):

PR #4321, opened 2025-09-23
fix crash due to over-sized IPC message

yushih:
Should fix https://emurgo.atlassian.net/browse/YOEXT-2191


packages/yoroi-extension/chrome/extension/background/constants.js
```diff
@@ -0,0 +1,2 @@
+export const MAX_MESSAGE_LENGTH = 5 * 1024 * 1024;
+export const OVERSIZE_RESPONSE_MESSAGE_PLACEHOLDER = 'OVERSIZE_RESPONSE_MESSAGE_PLACEHOLDER';
```
jarstorm: This constant it's declared twice. Is that on purpose or it was a mistake?

packages/yoroi-extension/chrome/extension/background/handlers/yoroi/index.js
```diff
@@ -104,11 +108,18 @@ export function getHandler(typeTag: string): ?Handler {
...
         const result = await handler(request.request);
         //console.debug(`BACKGROUND [${typeTag}] sending result: `, JSON.stringify(sanitizeForLog(result)));
-        sendResponse(result);
+        const serializedResult = JSON.stringify(result);
+        if (serializedResult.length <= MAX_MESSAGE_LENGTH) {
+          sendResponse(result);
+        } else {
+          const messageId = String(incrementalId++);
+          sendResponse({ type: OVERSIZE_RESPONSE_MESSAGE_PLACEHOLDER, messageId });
+          sendLongMessage(sender.tab.id, serializedResult, 'long-response', messageId, MAX_MESSAGE_LENGTH);
```
jarstorm: 'long-response' should be a constant

jarstorm (dismissed):

packages/yoroi-extension/chrome/extension/background/constants.js
```diff
@@ -0,0 +1,2 @@
+export const MAX_MESSAGE_LENGTH = 5 * 1024 * 1024;
+export const OVERSIZE_RESPONSE_MESSAGE_PLACEHOLDER = 'afb5759a-eaef-4598-92a7-db414f716322';
```
jarstorm, line 2: This value it's ok? Should it be a random value?

yushih: Yes. We just need a unique value. As long as it's uni, any value is OK here.

Nebyt (approved):

